Issue #84, opened 2017-12-20
Correctly incorporate committer data everywhere


bockthom:
As we now have committer data available (#35), we need to incorporate the committer data everywhere:

- [ ] Rename several functions and configuration options (e.g., `author.only.committers` is not correctly named any more)
- [ ] Provide access to aggregated committer data (e.g., `committer2artifact` (respect different kinds of artifacts), `committer2commit`, etc.)
- [ ] Enable constructing committer networks (similar to all the other author networks)
- [ ] Construct networks with authors and committers (possible for author networks, bipartite networks, multi-networks)

Anything to add here?

PR #169, opened 2019-07-11
Add possibility of editor definition to add.vertex.attribute.artifact.editor.count

klaraschlueter:
### Prerequisites

### Description

Add a parameter 'editor.definition' to the function 'add.vertex.attribute.artifact.editor.count' which can be used to define, if author or committer or both count as editors when computing the attribute values.

### Changelog

Add a parameter 'editor.definition' to the function 'add.vertex.attribute.artifact.editor.count' which can be used to define, if author or committer or both count as editors when computing the attribute values. (ff1e147ba563b2d71f8228afd49492a315a5ad48)

### DIFF
--- a/util-networks-covariates.R
+++ b/util-networks-covariates.R
@@ -666,7 +666,8 @@ add.vertex.attribute.author.role = function(list.of.networks, classification.res
 #'                          \code{"project.cumulative"}, \code{"project.all.ranges"}, and
 #'                          \code{"complete"}. See \code{split.data.by.networks} for
 #'                          more details. [default: "range"]
-#' @param editor.definition Determines, who is counted as editor of an artifact. [default: c("author", "committer")]
+#' @param editor.definition Determines, who is counted as editor of an artifact (one ore more of
+#'                          \code{c("author", "committer")}). [default: "author"]
 #' @param default.value The default value to add if a vertex has no matching value [default: 0]
 #'
 #' @return A list of networks with the added attribute
@@ -679,15 +680,19 @@ add.vertex.attribute.artifact.editor.count = function(list.of.networks, project.
     aggregation.level = match.arg.or.default(aggregation.level, default = "range")
 
     ## match editor definitions to column name in commit dataframe
-    editor.definition = match.arg.or.default(editor.definition, several.ok = TRUE)
-    editor.definition = lapply(editor.definition, function(editor) {paste0(editor, ".name")})
+    if (missing(editor.definition)) {
+        editor.definition = "author"
+    } else {
+        editor.definition = match.arg.or.default(editor.definition, choices = c("author", "committer"), several.ok = TRUE)
+    }
+    editor.definition = paste0(editor.definition, ".name")
 
     nets.with.attr = split.and.add.vertex.attribute(
         list.of.networks, project.data, name, aggregation.level, default.value,
         function(range, range.data, net) {
             vertex.attributes = lapply(range.data$group.authors.by.data.column("commits", "artifact"),
                    function(artifact.commits) {
-                       editor.count = length(unique(unlist(lapply(editor.definition, function(editor.type) {artifact.commits[[editor.type]]}))))
+                       editor.count = length(unique(unlist(artifact.commits[editor.definition])))
                        return(editor.count)
                    }
             )
--- a/util-networks-covariates.R
+++ b/util-networks-covariates.R
@@ -666,6 +666,7 @@ add.vertex.attribute.author.role = function(list.of.networks, classification.res
 #'                          \code{"project.cumulative"}, \code{"project.all.ranges"}, and
 #'                          \code{"complete"}. See \code{split.data.by.networks} for
 #'                          more details. [default: "range"]
+#' @param editor.definition Determines, who is counted as editor of an artifact. [default: c("author", "committer")]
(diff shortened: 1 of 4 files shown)


util-networks-covariates.R
```diff
@@ -666,22 +666,28 @@ add.vertex.attribute.author.role = function(list.of.networks, classification.res
...
                                                       aggregation.level = c("range", "cumulative", "all.ranges",
                                                                             "project.cumulative", "project.all.ranges",
                                                                             "complete"),
+                                                      editor.definition = c("author", "committer"),
                                                       default.value = 0) {
     aggregation.level = match.arg.or.default(aggregation.level, default = "range")
 
+    ## match editor definitions to column name in commit dataframe
+    editor.definition = match.arg.or.default(editor.definition, several.ok = TRUE)
+    editor.definition = lapply(editor.definition, function(editor) {paste0(editor, ".name")})
```
clhunsen: The function `paste0` iterates automatically over all items of a vector. Your solution works but seems to be a little heavy, IMHO. How about the following: ` editor.definition = paste0(editor.definition, ".name")`?

util-networks-covariates.R
```diff
@@ -666,22 +666,28 @@ add.vertex.attribute.author.role = function(list.of.networks, classification.res
...
+
     nets.with.attr = split.and.add.vertex.attribute(
         list.of.networks, project.data, name, aggregation.level, default.value,
         function(range, range.data, net) {
             vertex.attributes = lapply(range.data$group.authors.by.data.column("commits", "artifact"),
-                   function(x) {
-                       editor.count = length(unique(c(x[["author.name"]], x[["committer.name"]])))
+                   function(artifact.commits) {
+                       editor.count = length(unique(unlist(lapply(editor.definition, function(editor.type) {artifact.commits[[editor.type]]}))))
```
clhunsen: Similar to my comment above: I suggest using `length(unique(unlist(artifact.commits[editor.definition])))`, as the function `unlist` also works on data.frames. Thus, would not need to embed  an `lapply` call.

clhunsen:
Ah, another comment because I forgot to look at the changelog :man_facepalming:: Can you please add the corresponding issue to the changelog entry? Also, please use backticks (<code>`</code>) for code there.

util-networks-covariates.R
```diff
@@ -666,22 +666,28 @@ add.vertex.attribute.author.role = function(list.of.networks, classification.res
 #'                          \code{"project.cumulative"}, \code{"project.all.ranges"}, and
 #'                          \code{"complete"}. See \code{split.data.by.networks} for
 #'                          more details. [default: "range"]
+#' @param editor.definition Determines, who is counted as editor of an artifact. [default: c("author", "committer")]
```
bockthom: I am not sure if `c("author", "committer")` is the default value we want to have here. I would go for just `"author"` as default value, as in most cases we are interested in the author of a commit and currently we only rely on the author in many other functions.

@clhunsen @klaraschlueter What do you think about that?

bockthom:
Thanks for this PR, @klaraschlueter.

As already mentioned in a personal meeting, could you please cherry-pick the following commit from my fork and add it to this pull request: 3a203cbe034045b035e55a31643112eb37355588

Thanks a lot in advance.

util-networks-covariates.R
clhunsen: As the integration of committer data is not completely rolled out yet (see also #84), using `"author"` as default value would be reasonable... :thinking:

I looked at the function `match.arg.or.default`, and it does not work properly for the current use case (due to restrictions on the parameter `default`). Consequently, we need some implementation work for this to run. For now, I would like to propose the following snippet to solve the current situation:
```R
if (missing(editor.definition)) {
    editor.definition = "author"
} else {
    editor.definition = match.arg.or.default(editor.definition, default = "author", several.ok = TRUE)
}
```

klaraschlueter: Using `author` as default sounds good to me @bockthom , another reason is backwards compatibility.

For what kind of call does `match.arg.or.default` return unexpected results, @clhunsen ?

bockthom: `match.arg.or.default` does not return unexpected results (as they are properly documented), but ignores the default parameter if `several.ok = TRUE` and returns the complete vector of allowed values.

klaraschlueter: Ah, I see. I ran some tests with your snippet, @clhunsen , (with one small change: I added the parameter `choices = c("author", "committer")` to the call to `match.arg.or.default`, since we decided for `"author"` as default).
I think, we do not need your if-case: if no editor.definition is given, the default value is inserted during the call to `add.vertex.attribute.artifact.editor.count` and `match.arg.or.default` returns the expected result `"author"`. A problem seems to arise if I call the function with a nonsense editor definition, for example: `editor.definition = nonsense`: The call to `add.vertex.attribute.artifact.editor.count` is satisfied and does not insert the default value, and `match.arg.or.default` throws an error: `'arg' should be one of “author”, “committer”`.

klaraschlueter: Okay, an update for you, @clhunsen : @bockthom and I agreed that a call with a nonsense editor definition should lead to an error, so everything works as expected if I call `match.arg.or.default` and specify the parameter `choices = c("author", "committer").

klaraschlueter: @clhunsen , @bockthom : done, committed, pushed!

PR #171, opened 2019-09-16
Evaluation plots for commit edit types

klaraschlueter:
### Prerequisites

### Description

A new file util-evaluation-plot contains functions to plot the number of commits per editor type (author and/or committer) and editor and the number of commits of the different edit types (author = committer, author /= committer) for the project.

### Changelog

- Add a new file util-evaluation-plot.R containing functions to plot commit edit types per author and project.


util-evaluation-plot.R
```diff
@@ -0,0 +1,97 @@
...
+## Plot functions ----------------------------------------------------------
+
+#' Produces a barplot showing for every editor the number of commits for which he is only author, only committer, and
+#' both author and committer.
+#'
+#' @param data The project data.
+#' @param percentage.per.author If true, the barplot shows the relative number of differently edited commits per author: each
+#'                              bar in the barplot (representing the commits of one editor) is scaled to 100%. Otherwise, the
+#'                              absolute number of commits per author is shown in the plot. [default: FALSE]
+plot.commit.editor.types.by.author = function(data, percentage.per.author = FALSE) {
```
bockthom: Could you please add a description for `@return`? Similarly to the `plot.network` function in the `util-plot` module, it would be helpful to know what the outcome is and to stay consistent with the other plotting module. 
What about `@return the barplot (invisibly)` ?
@clhunsen Do you agree with that?

This also holds for the second plotting function in this module.

NEWS.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 - Add a parameter `editor.definition` to the function `add.vertex.attribute.artifact.editor.count` which can be used to define, if author or committer or both count as editors when computing the attribute values. (#92, ff1e147ba563b2d71f8228afd49492a315a5ad48)
+- Add a new file util-evaluation-plot.R containing functions to plot commit edit types per author and project.
```
bockthom: In this description, the commit hashes are missing.

util-evaluation-plot.R
klaraschlueter: Looking at the `util-plot` file, I think the most consistent option is to state `@return a ggplot2/ggraph plot object` and actually return the plot object. The visual plot can then be produced by calling the returned value. This is the way it is done in `util-plot:plot.get.plot.for.network`, which also produces only a plot. What do you think about that? I  am going to commit this solution in 2 minutes, so you can have a look at it.

bockthom:
Thanks @hechtlC, I did not check for the missing space after `if`, my fault. I am not sure whether this is part of the contribution guide yet, but we have enforced this all the time and therefore should enforce this here. Please check if it is part of the contribution guide and if not, please open an issue @hechtlC.

Regarding complete sentences: I am currently not sure about that - and way too busy to check that. Please check that by yourselves and apply the rules stated there to the function descriptions.

Regarding line breaks: We break lines **after 120 characters**. And we stick to that rule, the preferred length will stay 120 characters at max. 80 characters is too short in our cases because we have lots of complex statements. The reason why we use 120 characters as limit is that if you view code on GitHub, GitHub displays up to 120 characters without placing a horizontal scrollbar.

_Sorry for being kind of rude in this comment. I just wanted to let you know about my opinion without having enough time to phrase this in polite enough way._

NEWS.md
```diff
@@ -4,7 +4,7 @@
 
 ### Added
 - Add a parameter `editor.definition` to the function `add.vertex.attribute.artifact.editor.count` which can be used to define, if author or committer or both count as editors when computing the attribute values. (#92, ff1e147ba563b2d71f8228afd49492a315a5ad48)
-- Add a new file util-evaluation-plot.R containing functions to plot commit edit types per author and project.
+- Add a new file util-evaluation-plot.R containing functions to plot commit edit types per author and project. (60c6abaed87af421f8d6b42465638b7359e205f4, 193bdf9349dc752674056233b0f5ded5871597f4)
```
bockthom: Thanks for adding the commit hashes here @klaraschlueter. The two commit hashes added represent the commits which contain the new plotting functions and introduce them to the showcase file. Unfortunately, the commit which actually creates the file `util-evaluation-plot.R` is missing here. So, could you please add a third commit hash here: d4465f325d2286e98a217eccdc4452721ed5db88
Thanks!

klaraschlueter:
> Another issue for which I am not sure if it is indeed an issue are the parameter descriptions in your roxygen comments. To my knowledge these should not be complete sentences and please correct me if I am wrong @bockthom @clhunsen @klaraschlueter.

I looked in several other files and it seems to me that the parameter descriptions aren't full sentences but only the subject part of a sentence, missing a verb, like: `The network list.` in `util.networks-covariates.R: add.vertex.attribute.author.email()`. If a parameter is more complex, for example a boolean flag, there are some whole sentences, too. So I think, the current documentation is okay, @hechtlC , are you okay with me leaving it as it is?

util-evaluation-plot.R
bockthom: Yes, that's ok. Did you try whether plotting still works (i.e., printing the plot after being returned from this function)? Maybe it would be nice to add an example statement to the showcase file - but if you are sure it works correctly, that is not really necessary.
I am not sure which naming convention to use here: As for `plot.get.plot.for.network`, it could be `plot.get.plot.for.commit.editor.types.by.author`, but that's way too long and not easy to understand any more. Due to the lack of a better idea, I am fine with the current name ;)

bockthom:
> > Another issue for which I am not sure if it is indeed an issue are the parameter descriptions in your roxygen comments. To my knowledge these should not be complete sentences and please correct me if I am wrong @bockthom @clhunsen @klaraschlueter.
> 
> I looked in several other files and it seems to me that the parameter descriptions aren't full sentences but only the subject part of a sentence, missing a verb, like: `The network list.` in `util.networks-covariates.R: add.vertex.attribute.author.email()`. If a parameter is more complex, for example a boolean flag, there are some whole sentences, too. So I think, the current documentation is okay, @hechtlC , are you okay with me leaving it as it is?

I just had a quick look at our coding conventions section in our contribution guide. It states:

_For each parameter of a function or method, a parameter description (@param parameter.name description) must be added. The same holds for descriptions of return values (@return description) – if something is returned. For both items, the descriptions should always be a phrase (i.e., no verbs should be used), should start with a lower-case letter, and should not end with a period – unless there are further phrases or sentences following._

That means: Your roxygen2 documentation is correct, unless the capital letters `T` in `The project data`

As we have coding conventions written down, we should just stick to them (which we already had agreed on some time ago...). However, we should stop being more picky than it is really worth to be. Functional changes should precede cosmetic changes.

util-evaluation-plot.R
klaraschlueter: Yes, of course i've tried it ;-)  It still works. I added example statements to the showcase, but I kept the names, since I also don't have any better idea.

bockthom: Ok, thanks @klaraschlueter.

bockthom:
I'm fine with this PR. If there are no further objections, I will merge.

klaraschlueter:
Thanks for your review, @clhunsen and @hechtlC , and sorry for taking so long to apply it! All requested changes are now applied and the corresponding conversations resolved, except the renaming, since I wasn't sure if @bockthom is okay with that.
I read the review message about my time contingent too late - yes, I am "out of time". Could you please take care of this pull request from now?

bockthom:
> Thanks for your review, @clhunsen and @hechtlC , and sorry for taking so long to apply it! All requested changes are now applied and the corresponding conversations resolved, except the renaming, since I wasn't sure if @bockthom is okay with that.

I am fine with both names. So, renaming would be ok for me.

> I read the review message about my time contingent too late - yes, I am "out of time". Could you please take care of this pull request from now?

No worries, we already assumed that you could already be "out of time". However, we all were very busy during the last months. So, nobody had time to finish this pull-request as we had to work on more important stuff. Hence, we are really glad that you have found some minutes to incorporate the requested changes. Thanks a lot!

Thanks for working on this pull-request again after so much time - we did not even expect that any more. All the best for your future @klaraschlueter!

clhunsen:
Also from my side, a very big thank you for your efforts, @klaraschlueter! It was a pleasure to work with you! :grinning::+1: All the best for the future and good luck. :blush: 

--------------------
As for the PR, we need a rebase, also to remove the conflict. We can update this PR or open another one with @klaraschlueter's commits.

clhunsen:
I would like to work on the changes in this PR tomorrow after PR #174 is merged. I will open a new PR then.

clhunsen:
Closing in favor of PR #175, as discussed before.

PR #172, opened 2019-09-24
Add patchstack mail filtering

SCPhantom:

### Description

This pull request enables patchstack mails to be filtered from the mails of the ProjectData.

### Changelog

Add the possibility to filter out patchstack mails from the mails of the ProjectData. The option can be toggled using
the newly added configuration option 'mails.filter.patchstack.mails'. (1608e28ca36610c58d2a5447d12ee2052c6eb976, c44c82b525e557a6cfe1f498478d2a7815ee9ae8)

### DIFF
--- a/tests/test-data.R
+++ b/tests/test-data.R
@@ -125,3 +125,52 @@ test_that("Compare two RangeData objects", {
     expect_false(proj.data.base$equals(range.data.four))
 
 })
+
+test_that("Filter patchstack mails", {
+
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("mails.filter.patchstack.mails", TRUE)
+
+    ## create the project data
+    proj.data = ProjectData$new(proj.conf)
+
+    ## retrieve the mails while filtering patchstack mails
+    mails.filtered = proj.data$get.mails()
+
+    ## create new project with filtering disabled
+    proj.conf$update.value("mails.filter.patchstack.mails", FALSE)
+    proj.data = ProjectData$new(proj.conf)
+
+    ## retrieve the mails without filtering patchstack mails
+    mails.unfiltered = proj.data$get.mails()
+
+    ## get message ids
+    mails.filtered.mids = mails.filtered[["message.id"]]
+    mails.unfiltered.mids = mails.unfiltered[["message.id"]]
+
+    expect_equal(setdiff(mails.unfiltered.mids, mails.filtered.mids), c("<hans2@mail.gmail.com>",
+                                                                        "<hans3@mail.gmail.com>",
+                                                                        "<hans4@mail.gmail.com>",
+                                                                        "<hans5@mail.gmail.com>",
+                                                                        "<hans6@mail.gmail.com>"))
+})
+
+test_that("Filter patchstack mails with PaStA enabled", {
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("mails.filter.patchstack.mails", TRUE)
+    proj.conf$update.value("pasta", TRUE)
+
+    proj.data = ProjectData$new(proj.conf)
+
+    ## retrieve filtered PaStA data by calling 'get.pasta' which calls the filtering functionality internally
+    filtered.pasta = proj.data$get.pasta()
+
+    ## ensure that PaStA data relating to Hans' mail 2 and 3 do not exist anymore since they have also been filtered
+    ## during patchstack mail filtering
+    expect_false("<hans2@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_false("<hans3@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+
+    ## ensure that all three PaStA entries that existed previously do still exist but have been associated to the
+    ## very first mail of the patchstack
+    expect_equal(3, sum(filtered.pasta[["message.id"]] == "<hans1@mail.gmail.com>"))
+})
--- a/tests/test-data.R
+++ b/tests/test-data.R
@@ -12,7 +12,8 @@
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
 ## Copyright 2018 by Christian Hechtl <hechtl@fim.uni-passau.de>
-## Copyright 2018 by Claus Hunsen <hunsen@fim.uni-passau.de>
+## Copyright 2018-2019 by Claus Hunsen <hunsen@fim.uni-passau.de>
+## Copyright 2019 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
 ## All Rights Reserved.
 
 
--- a/tests/test-data.R
+++ b/tests/test-data.R
@@ -172,8 +172,8 @@ test_that("Filter patchstack mails with PaStA enabled", {
     expect_true("<jlkjsdgihwkfjnvbjwkrbnwe@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
     expect_equal(2, sum(filtered.pasta[["message.id"]] == "<saf54sd4gfasf46asf46@mail.gmail.com>"))
 
-    ## ensure that out of three PaStA entries that existed previously, all of which pointing to the same commit hash,
-    ## one new PaStA entry has been created with has assigned the message ID of the first patchstack mail
+    ## ensure that the three PaStA entries relating to the filtered patchstack mails have been merged to a single new
+    ## PaStA entry which has assigned the message ID of the first patchstack mail
     expect_true("<hans1@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
 
     ## ensure that there are no other entries than the ones that have been verified to exist above
--- a/tests/test-data.R
+++ b/tests/test-data.R
@@ -34,6 +34,7 @@ test_that("Compare two ProjectData objects", {
 
     ##initialize a ProjectData object with the ProjectConf and clone it into another one
     proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("pasta", TRUE)
     proj.data.one = ProjectData$new(project.conf = proj.conf)
     proj.data.two = proj.data.one$clone()
 
@@ -43,19 +44,20 @@ test_that("Compare two ProjectData objects", {
     ## second object, as well, and test for equality.
 
     ##change the second data object
-    proj.data.one$get.commits()
-
-    expect_false(proj.data.one$equals(proj.data.two), "Two not identical ProjectData objects.")
-
-    proj.data.two$get.commits()
-
-    expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
 
     proj.data.two$get.pasta()
 
     expect_false(proj.data.one$equals(proj.data.two), "Two not identical ProjectData objects.")
 
     proj.data.one$get.pasta()
+
+    expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
+
+    proj.data.one$get.commits()
+
+    expect_false(proj.data.one$equals(proj.data.two), "Two not identical ProjectData objects.")
+
+    proj.data.two$get.commits()
 
     expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
 
--- a/tests/test-data.R
+++ b/tests/test-data.R
@@ -166,12 +166,16 @@ test_that("Filter patchstack mails with PaStA enabled", {
     ## retrieve filtered PaStA data by calling 'get.pasta' which calls the filtering functionality internally
     filtered.pasta = proj.data$get.pasta()
 
-    ## ensure that PaStA data relating to Hans' mail 2 and 3 do not exist anymore since they have also been filtered
-    ## during patchstack mail filtering
-    expect_false("<hans2@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
-    expect_false("<hans3@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    ## ensure that the remaining mails have not been touched
+    expect_true("<adgkljsdfhkwafdkbhjasfcjn@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_true("<asddghdswqeasdasd@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_true("<jlkjsdgihwkfjnvbjwkrbnwe@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_equal(2, sum(filtered.pasta[["message.id"]] == "<saf54sd4gfasf46asf46@mail.gmail.com>"))
 
-    ## ensure that all three PaStA entries that existed previously do still exist but have been associated to the
-    ## very first mail of the patchstack
-    expect_equal(3, sum(filtered.pasta[["message.id"]] == "<hans1@mail.gmail.com>"))
+    ## ensure that out of three PaStA entries that existed previously, all of which pointing to the same commit hash,
+    ## one new PaStA entry has been created with has assigned the message ID of the first patchstack mail
+    expect_true("<hans1@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+
+    ## ensure that there are no other entries than the ones that have been verified to exist above
+    expect_equal(6, nrow(filtered.pasta))
 })
--- a/util-conf.R
+++ b/util-conf.R
@@ -355,6 +355,12 @@ ProjectConf = R6::R6Class("ProjectConf", inherit = Conf,
                 allowed = c(TRUE, FALSE),
                 allowed.number = 1
             ),
+            mails.filter.patchstack.mails = list(
+                default = FALSE,
+                type = "logical",
+                allowed = c(TRUE, FALSE),
+                allowed.number = 1
+            ),
             synchronicity = list(
                 default = FALSE,
                 type = "logical",
--- a/util-data.R
+++ b/util-data.R
@@ -119,9 +119,9 @@ ProjectData = R6::R6Class("ProjectData",
 
         ## * * commit filtering --------------------------------------------
 
-        #' Filter commits by potentially removing untracked files and the base artifact (see parameters).
+        #' Filter commits by potentially removing commits to untracked files or to the base artifact (see parameters).
         #'
-        #' @param commits the data.frame of commits to be filtered
+        #' @param commits the data.frame of commits on which filtering will be applied
         #' @param remove.untracked.files flag whether untracked files are kept or removed
         #' @param remove.base.artifact flag whether the base artifact is kept or removed
         #'
@@ -145,8 +145,8 @@ ProjectData = R6::R6Class("ProjectData",
 
         ## * * mail filtering ----------------------------------------------
 
-        #' Filters patchstack mails from the mails that are currently cached in the field \code{mails} and returns them.
-        #' Detected patchstacks are also stored in the field \code{patchstack.mails}. They are used later in the
+        #' Filter patchstack mails from the mails that are currently cached in the field \code{mails} and return them.
+        #' Store detected patchstacks in the field \code{patchstack.mails}. They are used later in the
         #' function \code{filter.pasta.data} to also accommodate for the deleted mails in the PaStA data.
         #'
         #' In a thread, a patchstack spans the first sequence of mails where each mail has been authored by the thread
@@ -163,6 +163,7 @@ ProjectData = R6::R6Class("ProjectData",
             ## retrieve mails grouped by thread IDs
             thread.data = self$group.authors.by.data.column("mails", "thread")
 
+            ## extract the patchstack mails and the filtered mails for each thread
             result = parallel::mclapply(thread.data, function(thread) {
 
                 ## ensure that all mails within the thread are ordered correctly
@@ -196,7 +197,7 @@ ProjectData = R6::R6Class("ProjectData",
             mails = plyr::rbind.fill(thread.data)
 
             ## Retrieve patchstacks from the result above which are used to manipulate the PaStA data. This needs to be
-            ## done because the PaSta data relates to some of the filtered mails and must be adjusted accordingly.
+            ## done because the PaStA data relates to some of the filtered mails and must be adjusted accordingly.
             patchstacks = lapply(result, function(x) x[["patchstack"]])
 
             ## only patchstacks that contain at least two mails are considered patchstacks
@@ -211,10 +212,12 @@ ProjectData = R6::R6Class("ProjectData",
 
         ## * * PaStA data --------------------------------------------------
 
-        #' Uses the information about the deleted patchstack mails that are stored in the field \code{patchstack.mails}
-        #' to also filter out PaStA information that relates to the deleted mails. The PaStA information is not
-        #' discarded completely however but instead is gathered for each patchstack and is assigned to the first mail
-        #' in each patchstack because this very first mail has not been filtered and represents the patchstack.
+        #' Use the information about the deleted patchstack mails that are stored in the field \code{patchstack.mails}
+        #' to also filter out PaStA information that relates to the deleted mails.
+        #'
+        #' The PaStA information is not discarded completely however but instead is gathered for each patchstack and is
+        #' assigned to the first mail in each patchstack because this very first mail has not been filtered and
+        #' represents the patchstack.
         #'
         #' @return the filtered PaStA data
         filter.pasta.data = function() {
@@ -222,7 +225,7 @@ ProjectData = R6::R6Class("ProjectData",
 
             new.pasta = parallel::mclapply(private$mails.patchstacks, function(patchstack) {
 
-                ## get all PaStA data that relates to the current mail (do not drop data.frame structure!)
+                ## get all PaStA data that relates to the current patchstack (do not drop data.frame structure!)
                 pasta.tmp = private$pasta[private$pasta[["message.id"]] %in% patchstack[["message.id"]], , drop = FALSE]
 
                 ## override all old message IDs with the message ID of the first mail in the patchstack since it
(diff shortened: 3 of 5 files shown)


util-data.R
```diff
@@ -63,6 +63,9 @@ DATASOURCE.TO.ARTIFACT.COLUMN = list(
     "issues"  = "issue.id"
 )
 
+## the maximum time difference between subsequent mails of a patchstack
+PATCHSTACK.MAIL.DECAY.THRESHOLD = "30 seconds"
```
hechtlC: Do you really mean decay here and not delay?

SCPhantom: I did. Both "decay" and "delay" make sense:
- delay: If subsequent are *delayed* with a maximum of 30 seconds in-between the are seen as patchstack mails
- decay: If the time difference between subsequent mails becomes higher than 30 seconds, the following mails are not seen as patchstack mails anymore and the patchstack *decays*

Both make sense in some way and both are not perfect as far as I see it, so I would just leave it for now. Any other suggestions?

util-data.R
```diff
@@ -445,44 +640,38 @@ ProjectData = R6::R6Class("ProjectData",
...
                 }
-                ## merge PaStA data
-                commit.data = merge(commit.data, private$pasta.commits,
-                                    by = "hash", all.x = TRUE, sort = FALSE)
             }
 
             ## sort by date again (because 'merge' is doing bullshit!)
             commit.data = commit.data[order(commit.data[["date"]], decreasing = FALSE), ] # sort!
 
-            private$commits = commit.data
```
bockthom: Why is this line removed? There may be a justified reason, but I currently don't get it.
So, if this line would be removed correctly, what's the sense of the `commit.data = ...` statement above, which just orders the statement? So, in the case that field assignments in `R6` classes use call-by-value (which I just assume without checking whether this really holds), the assignment does not have any effect.

However, sorting the data is necessary here (even if  the comment `because 'merge' is doing bullshit` is not that obvious any more as the corresponding `merge` calls have been moved to several `update.*.data()` functions...)

Am I correct with my assumptions? Anyway, sorting of the data needs to be double-checked at several places...

util-data.R
```diff
@@ -113,37 +117,138 @@ ProjectData = R6::R6Class("ProjectData",
...
+            new.pasta = plyr::rbind.fill(new.pasta)
+
+            ## remove old items from PaStA data
+            ## 1) flatten the list of mail-dataframes (i.e. patchstacks) to a single mail-dataframe
+            patchstack.mails = plyr::rbind.fill(private$mails.patchstacks)
+            ## 2) delete any PaStA information that relate to message IDs of mails that will be discarded
+            pasta = private$pasta[!(private$pasta[["message.id"]] %in% patchstack.mails[["message.id"]]), ]
+
+            ## append the new pasta to the old pasta
+            pasta = plyr::rbind.fill(pasta, new.pasta)
```
bockthom: Have the pasta data been sorted before? If so, we need to sort them again, as appending `new.pasta` to `pasta` does break the sorted order. We need to check how and if and where the pasta have been sorted before and whether they are sorted in a later step - and if not, we may need to sort them here? We definitively should check for a reasonable order at several related places...

tests/test-data.R
```diff
@@ -125,3 +125,52 @@ test_that("Compare two RangeData objects", {
...
+    filtered.pasta = proj.data$get.pasta()
+
+    ## ensure that PaStA data relating to Hans' mail 2 and 3 do not exist anymore since they have also been filtered
+    ## during patchstack mail filtering
+    expect_false("<hans2@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_false("<hans3@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+
+    ## ensure that all three PaStA entries that existed previously do still exist but have been associated to the
+    ## very first mail of the patchstack
+    expect_equal(3, sum(filtered.pasta[["message.id"]] == "<hans1@mail.gmail.com>"))
```
bockthom: Maybe we should also check for the remaining entries of the PaStA data that they did not have gone or something else. I currently don't know how the underlying data looks like, but our test should make sure that the complete number of rows is correct and not only the amount of rows in which the message id has been replaced.
So, if there are only those three lines in this test, maybe it would be useful to also have a test where we have also some other rows which don't get changed. However, I am not sure whether it is easy to add such a test or not, depending on the underlying data...

util-conf.R
```diff
@@ -355,6 +355,12 @@ ProjectConf = R6::R6Class("ProjectConf", inherit = Conf,
                 allowed = c(TRUE, FALSE),
                 allowed.number = 1
             ),
+            mails.filter.patchstack.mails = list(
+                default = FALSE,
```
bockthom: Please add this configuration option also to the README file (see section https://github.com/se-passau/coronet#configurable-data-retrieval-related-parameters)

util-data.R
```diff
@@ -113,37 +117,138 @@ ProjectData = R6::R6Class("ProjectData",
...
         data.timestamps = NULL, #data.frame
 
-        ## * * filtering commits -------------------------------------------
+        ## * * commit filtering --------------------------------------------
 
-        #' Filter commits retrieved by the method \code{get.commits} after potentially removing untracked files and the
-        #' base artifact (see parameters).
+        #' Filter commits by potentially removing untracked files and the base artifact (see parameters).
         #'
+        #' @param commits the data.frame of commits to be filtered
```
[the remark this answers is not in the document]
SCPhantom: I don't really see your point. For me the meaning of the phrase "the data.frame of commits to be filtered" is clear.

What do you mean with "the data of this data.frame needs to be filtered"? Like before even passing the data.frame to the function?
What do you mean with "filtered from elsewhere"?

util-data.R
```diff
@@ -113,37 +117,138 @@ ProjectData = R6::R6Class("ProjectData",
...
+        #' discarded completely however but instead is gathered for each patchstack and is assigned to the first mail
+        #' in each patchstack because this very first mail has not been filtered and represents the patchstack.
+        #'
+        #' @return the filtered PaStA data
+        filter.pasta.data = function() {
+            logging::logdebug("filter.pasta.data: starting.")
+
+            new.pasta = parallel::mclapply(private$mails.patchstacks, function(patchstack) {
+
+                ## get all PaStA data that relates to the current mail (do not drop data.frame structure!)
```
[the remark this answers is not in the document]
SCPhantom: Yes, you're right! I changed this once but forgot to update the comment accordingly.

util-data.R
SCPhantom: That is definitely a mistake. The last statement does indeed not have any effect and the comment referencing a 'merge' that is not immediately visible is also not very helpful.

util-data.R
SCPhantom: Does PaStA data have to be sorted at all? Previous to this PR, this was not required as you can just set unordered PaStA data using `set.pasta`.

tests/test-data.R
SCPhantom: In this test, we do have more than the three lines that have changed. I'll add some statements that assert that the remaining lines do not have changed.

util-data.R
SCPhantom: > I wonder if the best way would be to move this functionality into the update.*.data() functions (but also keep it in the setters because the data that is passed to a setter could not be sorted)

I have now implemented it like this. This should work, however sometimes data will be sorted twice.

util-data.R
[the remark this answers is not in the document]
bockthom: There are two different kinds of interpretation:

Interpretation 1: `commits` are the data on which the filter should be applied (which is the correct interpretaion in this case)

Interpretation 2: `commits` are the data which need to be removed from the actual data (which is the wrong interpretation)

util-data.R
bockthom: Better sorting twice than never (at least, I hope, that does not really make a difference in computational time...)

util-data.R
bockthom: Could you please check if the data was previously sorted before reading it? If not, then just ignore my comment, if there is not really a need to sort the data...

tests/test-data.R
bockthom: Sounds good!

bockthom:
Please check also if you really adhere to the coding conventions regarding function documentation. For example, do you adhere to the following one?

_Use 2nd-person imperative ("Get the label.") rather than 3rd-person declarative ("Gets the label.")._

util-data.R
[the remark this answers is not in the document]
SCPhantom: Ah, now I see it. That is indeed an important ambiguity 😄 

What do you think of the detailed description "the data.frame of commits from which commits to untracked files and the base artifact will be filtered"?

Alternatively, I suggest "the data.frame of commits on which filtering will be applied".
But I also like your suggestion "the data.frame on which to apply the filtering".

util-data.R
SCPhantom: I just took a closer look. In the worst case when calling `set.commits` with both PaStA and synchronicity enabled, it will actually be sorted three times. Once directly in the setter (serving the purpose of sorting the incoming data), once within `private$update.synchronicity.data()` and once within `private$update.pasta.data()` (within this last function call there occurs even a fourth sorting operation which sorts the mails in addition to the commits).

I don't know to what extent computation time is affected here. I guess it is minor even though there are four sorting operations in the worst case. The alternative would be to add a parameter `sort=TRUE` to the methods `update.*.data` and whenever there is the possibility to save some operations, sorting can be disabled. Which way do you prefer?

Another alternative would be to just resestablish sorting manually by inserting the sort statement after every call to `update.*.data`. This, however, might cause problems when someone is reusing this function in future, unaware of the fact that `update.*.data` functions disturb order.

util-data.R
SCPhantom: I checked it using Busybox and calling `project.data$get.pasta()`. The dataframe is ordered by the column `revision.set.id`. The other columns `message.id` and `commit.hash` are unordered.

Should I reestablish this ordering?

util-data.R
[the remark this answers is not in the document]
bockthom: Agreed. Only one additional question: "... commits to untracked files and **to** the base artifact" Is it really an **and** here or an **or**? Please check and use the correct version of this sentence then.

util-data.R
bockthom: I would just sort it without any parameter. If PaStA and synchronicity are not enabled, it does not really matter, and if they are enabled, it should not make that much difference in computation time to sort them even more times. So, just sort it and then it should be fine.

util-data.R
bockthom: Yeah, that's what I had in mind. So, please order the data by `revision.set.id`.

tests/test-data.R
SCPhantom: I have one more question before I can push all the new commits.

The three new PaStA lines look like this:
```
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
```

Should this be the expected result? Or should this line be simplified to one? I think it should because this is a situation that can not occur when reading in PaStA from disk. At least I think so but I have already mailed @hechtlC. Since he has not replied yet maybe you, @bockthom, @clhunsen can answer this.

What should happen in the following case?
```
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-5>
```
Should it also be simplfied to one line? I also vote for yes. Unless, of course, the commit hashes differ.

Another scenario:
```
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | sdf34234lksfdlk2kl34nokfnsrl2k3j4l2k3j4l | <revision-set-5>
<hans1@mail.gmail.com> | 343jk52kldfgkln3ma0fjo3ej0j52jrkfldfj231 | <revision-set-6>
```
In this case, no duplicates should be removed, however the `revision.set.id` of all these three lines should be set to `revision-set-4` which is the ID of the first mail. I would do this because this scenario is also one that, I believe, can not occur when reading PaStA files.

bockthom: Good point @SCPhantom. However, this question is not that easy to answer because the interpretation behind the data has changed a little bit now. So I will try to refer to your three example scenarios in the reverse order, as it is easier to explain then.

> ```
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | sdf34234lksfdlk2kl34nokfnsrl2k3j4l2k3j4l | <revision-set-5>
> <hans1@mail.gmail.com> | 343jk52kldfgkln3ma0fjo3ej0j52jrkfldfj231 | <revision-set-6>
> ```
> 
> In this case, no duplicates should be removed, however the `revision.set.id` of all these three lines should be set to `revision-set-4` which is the ID of the first mail. I would do this because this scenario is also one that, I believe, can not occur when reading PaStA files.

I am not really sure about this. My first intuition was that we should **not** replace the revision set ids. One revision set contains all the patches which are identified as being similar by PaStA. As a commit can contain changes on different chunks, this one commit could be mapped to different revision sets. On the other hand, one e-mail could contain several patches being identified as different revision-sets. I am not sure about how PaStA would handle such cases. @hechtlC Do you know how PaStA handles such cases. Anyway, even if PaStA does it differently, I would keep the information we have determined and therefore treat them as different revision sets, as we know that they actually were part of different revision sets before filtering.

> What should happen in the following case (revision set id of the last line is different)?
> 
> ```
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-5>
> ```
> 
> Should it also be simplfied to one line? I also vote for yes. Unless, of course, the commit hashes differ.

Regarding the revision sets, I have the same opinion stated above: Keep the different revision sets as one commit can have different chunks being part of different revision sets.

> The three new PaStA lines look like this:
> 
> ```
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> ```

I am not sure about this case, that's why I have moved it to the end of this posting. As revision set and commit hash are identical, I would like to combine them to one line. However, keeping all three lines tells us that there have been three e-mails before. But I am not sure if this information is necessary. So, in that case, I currently would prefer to call `unique` on the complete data frame to avoid lines being completely identical.

Anyway, all those examples are very good examples, resulting in questions that are pretty hard to answer. The main idea should be: Don't mix revision sets, keep them separately. But remove redundant information. @hechtlC Do you agree with me?

SCPhantom: > I am not sure about this case, that's why I have moved it to the end of this posting. As revision set and commit hash are identical, I would like to combine them to one line. However, keeping all three lines tells us that there have been three e-mails before. But I am not sure if this information is necessary. So, in that case, I currently would prefer to call unique on the complete data frame to avoid lines being completely identical.

I agree with removing duplicate entries (duplicates that are equal in all three columns). If the `mails.filter.patchstack.mails` configuration option is active, all the mails of a patchstack should be treated as a single mail. Not eliminating those duplicates would be inconsistent and might confuse the user later.

> I am not really sure about this. My first intuition was that we should not replace the revision set ids. One revision set contains all the patches which are identified as being similar by PaStA. As a commit can contain changes on different chunks, this one commit could be mapped to different revision sets. On the other hand, one e-mail could contain several patches being identified as different revision-sets. I am not sure about how PaStA would handle such cases. @hechtlC Do you know how PaStA handles such cases. Anyway, even if PaStA does it differently, I would keep the information we have determined and therefore treat them as different revision sets, as we know that they actually were part of different revision sets before filtering.

It depends all on the question if it is allowed in PaStA to have patches that are part of mutliple revision sets (i.e., if the same message ID is allowed to appear in more than one row in the original PaStA file).
If is this is already possible in PaStA, I agree with you, @bockthom because it is simply extra information that we can keep. If is is not the case, I would once again replace all revision set IDs with the ID of the first mail. Otherwise, we would just introduce a special case to the PaStA information which is only relevant when patchstack mail filtering is active after all and thus, I assume that the benefit of having this extra bit of information is minor.

util-data.R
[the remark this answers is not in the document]
clhunsen: > Alternatively, I suggest "the data.frame of commits on which filtering will be applied".

@hechtlC and I vote for this.

util-data.R
bockthom: That's true, there is no need to order the data. However, I recommend to order the data for debugging reasons. It might be the case that -- due to some filtering and readding operations - the lines belong to one revision set are not any more right after each other. If someone looks at this data frame (for whatever reason), this is not a good idea. Therefore, I would like to order the data by revision set id to keep related lines just after each other. That does not involve a big effort, but may ease debugging in some point.

clhunsen: Sounds reasonable. :+1:

hechtlC: One more comment regarding the comment here: shouldn't we write "append the new pasta **data** to the old pasta **data**"? What do you think @clhunsen @bockthom ?

util-data.R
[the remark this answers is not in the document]
SCPhantom: >> Alternatively, I suggest "the data.frame of commits on which filtering will be applied".

>@hechtlC and I vote for this.

Then I'll adapt this.

tests/test-data.R
SCPhantom: I'll add this unique call and will wait for updates from @hechtlC.

SCPhantom:
Since I forgot to put in some "signed-off" tags, I unfortunately had to force push, thereby touching a few of the older commits. I tried to incorporate all the aspects discussed above. Also, I'll wait for the reply from the PaStA developer(s) 😄

tests/test-data.R
hechtlC: Ok so I heard from Ralf the PaStA maintainer. According to him there is no possibility that a message ID occurs multiple times in the PaStA data. If there happen to be two patches in one mail then PaStA still clusters it as one item. So the case that a message ID is within more than one revision set can not occur.

It can however happen that the case of the same message ID for different revision sets occurs after your filtering of the patchstack mails. One patchstack, as you know, consists of more than one mail (usually) and can contain more than one patch. So if the mails of the patchstack are simplified to the one first message ID of the stack then this can happen. We should however **not** lose the information on the different revision sets. So either we need a list of revision set ids in the one combined patchstack item or we need to keep one item for each revision set. 

@clhunsen  and I vote for the list of revision set ids as it makes the merging a lot easier and cleaner. 

I just realized another problem, which I just discussed with @clhunsen. Imagine the following scenario:
I send a patch via a patchstack to the mailing list. This then splits the patch into its single commits. These commits can now belong to different revision sets because PaStA views each commit as a single patch. If there happen to be two or more similar commits then it can happen that more than one of these commits is mapped to the same revision set. And currently there is not a possibility to get this information, right? This information is however important for example for my analyses about the number of revisions a developer needs to have the patch accepted. 

So @clhunsen and I vote for a possibility to get the unfiltered PaStA data in addition to the filtered ones (analogously to the commit data). What do you think @bockthom @SCPhantom ?

clhunsen: > It can however happen that the case of the same message ID for different revision sets occurs after your filtering of the patch-stack mails. One patch stack, as you know, consists of more than one mail (usually) and can contain more than one patch. So if the mails of the patch stack are simplified to the one first message ID of the stack then this can happen. We should however **not** lose the information on the different revision sets. So either we need a list of revision set ids in the one combined patchstack item or we need to keep one item for each revision set.
> 
> @clhunsen and I vote for the list of revision set ids as it makes the merging a lot easier and cleaner.

I had a look into the code again and realized that we (still) call [`ProjectData$aggregate.pasta.data`](https://github.com/se-passau/coronet/pull/172/files#diff-d11a0917903f656c45b6543d246e3ac3R363), which resolves the issue. Sorry for the false alarm.

SCPhantom: As a reply to the last paragraph of you, @bockthom:
PaStA filtering is indeed different than commit filtering and I also don't think it would be a good idea to implement the "unfiltered pasta" feature unless there is a relevant use case which needs to see both versions of PaStA. Unfortunately, I don't know your exact analyses, @hechtlC, so I can't tell whether this is the case for you. But even if it was, I would rather enhance the existing method of patchstack filtering by the features you need.

---
Citation of @hechtlC :
>I just realized another problem, which I just discussed with @clhunsen. Imagine the following scenario:
I send a patch via a patchstack to the mailing list. This then splits the patch into its single commits. These commits can now belong to different revision sets because PaStA views each commit as a single patch. If there happen to be two or more similar commits then it can happen that more than one of these commits is mapped to the same revision set. And currently there is not a possibility to get this information, right? This information is however important for example for my analyses about the number of revisions a developer needs to have the patch accepted.

As @bockthom pointed out in his example, this is not a problem related to patchstack mail filtering since message IDs of patchstack mails are the only thing being dropped (and redundant PaStA entries). Actually, I don't really see the problem in the first place. The information whether two or more comments do belong to the same revision set is appended to the dataframe `commits` in the column `revision.set.id`, isn't it?

SCPhantom: And now to come back to the actual issue here:
> Ok so I heard from Ralf the PaStA maintainer. According to him there is no possibility that a message ID occurs multiple times in the PaStA data. If there happen to be two patches in one mail then PaStA still clusters it as one item. So the case that a message ID is within more than one revision set can not occur.

 As @hechtlC already said, it is impossible for a message ID to be in more than one revision set. So, if a mail contains a patch, its message ID belongs to exactly one revision set.

To reference an old example:
```
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
<hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-5>
```
If the above case occurs after filtering patchstack mails and replacing all message IDs of the patchstack with the message ID of the first mail, we are required to set the revision set id of all three PaStA entries to the same revision set id, e.g. the one of the first mail `<revision-set-4>` in order to stay consistent with the notion of PaStA given above. After that `unique` can be called wrt all three columns.

Unfortunately, we now actually do lose information about revision sets. So, the problem that was mentioned by @hechtlC is justified 😃

bockthom: > And now to come back to the actual issue here:
> 
> > Ok so I heard from Ralf the PaStA maintainer. According to him there is no possibility that a message ID occurs multiple times in the PaStA data. If there happen to be two patches in one mail then PaStA still clusters it as one item. So the case that a message ID is within more than one revision set can not occur.
> 
> As @hechtlC already said, it is impossible for a message ID to be in more than one revision set. So, if a mail contains a patch, its message ID belongs to exactly one revision set.
> 
> To reference an old example:
> 
> ```
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-4>
> <hans1@mail.gmail.com> | 1143db502761379c2bfcecc2007fc34282e7ee61 | <revision-set-5>
> ```
> 
> If the above case occurs after filtering patchstack mails and replacing all message IDs of the patchstack with the message ID of the first mail, we are required to set the revision set id of all three PaStA entries to the same revision set id, e.g. the revision set id of the first mail, in order to stay consistent with the notion of PaStA given above. After that `unique` can be called wrt all three columns.

No, I would suggest to explicitly break consistency with the real notion of PaStA. The really important information of the PaStA data are the revision sets. Roughly summarized, we are only interested in which mails and commits belong to each other, which are basically revision sets. Unfortunately, PaStA is not able to differentiate between different patches within the same male as it assigns the same message id to both patches. So, in this case, PaStA is not able to provide more information.
However, when we filter the patch stack messages, we *are* able to differentiate between different patches assigned to the same message id (the id of the cover letter). So we *have* the information that those patches belong to different revision sets. Therefore we should *not* adjust the revision sets while filtering -- just keep the original revision set, that's all.

In sum, our goal should be to gain / keep as much information as possible. Even when PaStA is not able to that in such cases, we definitive are able to achieve that. That's why I would go for keeping the original revision sets even if they are assigned to the same message id, which cannot occur in the original PaStA data, but we then are more accurate.

SCPhantom: Alright, then we don't even have to change anything in the code regarding this! 😄

tests/test-data.R
```diff
@@ -166,12 +166,16 @@ test_that("Filter patchstack mails with PaStA enabled", {
...
-    ## very first mail of the patchstack
-    expect_equal(3, sum(filtered.pasta[["message.id"]] == "<hans1@mail.gmail.com>"))
+    ## ensure that the remaining mails have not been touched
+    expect_true("<adgkljsdfhkwafdkbhjasfcjn@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_true("<asddghdswqeasdasd@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_true("<jlkjsdgihwkfjnvbjwkrbnwe@mail.gmail.com>" %in% filtered.pasta[["message.id"]])
+    expect_equal(2, sum(filtered.pasta[["message.id"]] == "<saf54sd4gfasf46asf46@mail.gmail.com>"))
+
+    ## ensure that out of three PaStA entries that existed previously, all of which pointing to the same commit hash,
+    ## one new PaStA entry has been created with has assigned the message ID of the first patchstack mail
```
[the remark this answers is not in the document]
SCPhantom: True 😄 👍

tests/test-data.R
hechtlC: > Just a follow-up question @hechtlC : Is this related to your analyses? Don't you need every single e-mail in general at derive the number of patch revisions, so that patch stack filtering should not be applied to your analyses at all? So I am not sure if there is a benefit for your analyses to apply patch stack filtering? It may be useful for investigating review times. But as you cannot apply the filtering for investigating patch revisions, I would stay with the unfiltered data also for the intervals - maybe it is interesting to try both and compare the differences, but from my point of view we should just stick to the unfiltered data. What do you think @clhunsen?

You are right here @bockthom this does not apply to my analyses. My bad i misread the situation so I'm fime with keeping it the way it is if there are no further objections.

> I currently don't understand what the actual problem is. So, could you please give some more hints please @hechtlC? Let's have a small example:

As stated above you are right. For the cases we need the filtering there seems to be no need to keep the unfiltered data.
Again sorry for the confuision.

If @bockthom or @clhunsen have no further objections I feel good about merging.
Again thanks for the nice work @SCPhantom.

clhunsen:
Are there any objections to merging this PR now? Otherwise, I will do that tomorrow in the afternoon.

Anyway, thanks for participating, everybody! :grinning::+1:

PR #173, opened 2019-10-21
Add tensor representation and EDCPTD centrality calculation

fehnkera:
Functions that enable to represent OSS projects as forth-order tensors and to calculate EDCPTD centrality on them are added to a temporary file 'util-tensor.R'. The functions are not final. Feedback and discussions are welcome.

Signed-off-by: Anselm Fehnker <anselm@muenster.de>

### Prerequisites

- [x] I adhere to the coding conventions (described [here](https://github.com/se-passau/coronet/blob/master/CONTRIBUTING.md)) in my code.
- [x] I have updated the copyright headers of the files I have modified.
- [x] I have written appropriate commit messages, i.e., I have recorded the goal, the need, the needed changes, and the location of my code modifications for each commit. This includes also, e.g., referencing to relevant issues.
- [x] I have put signed-off tags in *all* commits.
- [x] I have updated the changelog file [NEWS.md](https://github.com/se-passau/coronet/blob/master/NEWS.md) appropriately.
- [x] The pull request is opened against the branch `dev`.

#### Added
- Add a new file `util-tensor.R` containing the class `FourthOrderTensor` to create (author x relation x author x relation) tensors from a list of networks (with each network having a different relation) and its corresponding utility function `get.author.networks.for.multiple.relations` (PR #173, c136b1f6127d73c25f08ae2f317246747aa9ea2b, e4ee0dc926b22ff75d5fd801c1f131bcff4c22eb, 051a5f0287022f97e2367ed0e9591b9df9dbdb3d)
- Add function `calculate.EDCPTD.centrality` for calculating the EDCPTD centrality for a fourth-order tensor in the above described form (c136b1f6127d73c25f08ae2f317246747aa9ea2b, e4ee0dc926b22ff75d5fd801c1f131bcff4c22eb, 051a5f0287022f97e2367ed0e9591b9df9dbdb3d)
- Add new file `util-networks-misc.R` which contains miscellaneous functions for processing network data and creating and converting various kinds of adjacency matrices: `get.author.names.from.networks`, `get.author.names.from.data`, `get.expanded.adjacency`, `get.expanded.adjacency.matrices`, `get.expanded.adjacency.matrices.cumulated`, `convert.adjacency.matrix.list.to.array` (051a5f0287022f97e2367ed0e9591b9df9dbdb3d)

#### Changed/Improved
[...]

PR #182, opened 2020-10-09
Fix function 'split.networks.time.based' regarding sliding windows

bockthom:
When calling the function `split.networks.time.based` (which splits several networks at the same time), the `sliding.window` parameter was useless: This function, in turn, calls `split.network.time.based` for each network, specifying fixed bins and passing the 'sliding.window' parameter to it. However, if bins are given to this function, the `sliding.window` parameter is ignored. To fix this, create the additional steps of the sliding window approach here manually in a similar way as it is done in
`split.network.time.based`. With this fix, it is possible to use sliding windows when splitting multiple networks simultaneously.

### Prerequisites

- [ ] I have updated the changelog file [NEWS.md](https://github.com/se-passau/coronet/blob/master/NEWS.md) appropriately.

### Description

See above.

### Changelog

This PR should be merged right after Anselm's PR #173 . I aim at rebasing this PR and adding a changelog as soon as Anselm's PR is finished. However, the fix provided in this PR can already be reviewed now.

### DIFF
--- a/util-split.R
+++ b/util-split.R
@@ -17,6 +17,7 @@
 ## Copyright 2017-2018 by Christian Hechtl <hechtl@fim.uni-passau.de>
 ## Copyright 2017 by Felix Prasse <prassefe@fim.uni-passau.de>
 ## Copyright 2017-2018 by Thomas Bock <bockthom@fim.uni-passau.de>
+## Copyright 2020 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -655,8 +656,51 @@ split.networks.time.based = function(networks, time.period = "3 months", bins = 
 
     ## split all networks to the extracted bins
     networks.split = lapply(networks, function(net) {
-        split.network.time.based(net, bins = bins.date, sliding.window = sliding.window,
-                                 remove.isolates = remove.isolates)
+        nets = split.network.time.based(net, bins = bins.date, sliding.window = sliding.window,
+                                        remove.isolates = remove.isolates)
+
+        ## perform additional steps for sliding-window approach
+        if (sliding.window && igraph::ecount(net) > 0) {
+            ## compute bins for sliding windows: pairwise middle between dates
+            bins.date.middle = mapply(
+                bins.date[1:(length(bins.date) - 1)],
+                bins.date[2:length(bins.date)],
+                FUN = function(d1, d2) d1 + ((d2 - d1) / 2)
+            )
+            bins.date.middle = get.date.from.unix.timestamp(bins.date.middle)
+
+            ## order edges by date
+            edges.all = igraph::E(net)
+            edges.dates = igraph::get.edge.attribute(net, "date")
+
+            ## identify edges to cut for sliding-window approach
+            edges.cut = sapply(edges.dates, function(date) {
+                date < bins.date.middle[1] || date > bins.date.middle[length(bins.date.middle)]
+            })
+
+            ## delete edges from the network and create a new network
+            net.cut = igraph::delete.edges(net, edges.all[edges.cut])
+
+            ## split network for sliding windows
+            nets.sliding = split.network.time.based(net.cut, bins = bins.date.middle, sliding.window = FALSE)
+
+            ## append data to normally-split data
+            nets = append(nets, nets.sliding)
+
+            ## sort data object properly by bin starts
+            bins.ranges.start = c(head(bins.date, -1), head(bins.date.middle, -1))
+            nets = nets[ order(bins.ranges.start) ]
+
+            ## construct proper bin vectors for configuration
+            bins.date = sort(c(bins.date, bins.date.middle))
+
+            attr(nets, "bins") = bins.date
+
+            ## set ranges as names
+            revs = get.date.string(bins.date)
+            names(nets) = construct.ranges(revs, sliding.window = sliding.window)
+        }
+        return(nets)
     })
 
     ## return the split networks


clhunsen:
Nice finding, @bockthom! I see the problem, but I had an idea for a solution of this problem that I would have liked to be discussed.  If this implementation idea works, we would not introduce code duplication.

So, the question is: Why don't you adjust the computation of the splitting bins directly?

### Disclaimer

I hope that I have not overlooked something that makes my proposal obsolete or garbage. :smile: In my head, everything seems to be perfect, but I do not trust this feeling right now. :grin::sweat_smile: 

### Implementation idea

Consider the following part of the buggy function.

https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-split.R#L636-L649

In my opinion, the buggy part is the block labeled `## 2) get bin information`: The user has not specified bins, but, when computing bins (for the splitting functionality to work in the first place), the sliding-window parameter is not considered appropriately.

The "easy" solution for the mentioned problem would be to introduce a better distinction of cases here to obtain the bins correctly:
- if `sliding.window` is set, call [the function `construct.overlapping.ranges`](https://github.com/se-passau/coronet/blob/dev/util-misc.R#L539), using `start = min(dates)`, `end = max(dates)`, and `overlap = 0.5`, and use the result to obtain the actual bins that are to be used in the splitting process,
- otherwise (i.e., if `number.windows` is set or both parameters are not set), call `split.get.bins.time.based(dates, time.period, number.windows)` (as it is called right now).

If we like to prefer `number.windows` over `sliding.window` if both are set, the if-else-block needs to reversed in its order.

### Request for feedback

What do you think about this idea? Does it work actually? Does it solve the concrete problem? What have I missed?

bockthom:
> The "easy" solution for the mentioned problem would be to introduce a better distinction of cases here to obtain the bins correctly:
> 
>  if `sliding.window` is set, call the function `construct.overlapping.ranges`, using `start = min(dates)`, `end = max(dates)`, and `overlap = 0.5`, and use the result to obtain the actual bins that are to be used in the splitting process,

> What do you think about this idea? Does it work actually? Does it solve the concrete problem? What have I missed?

First of all, thanks for your comment @clhunsen. I tried implementing your idea, but it seems to me that i does not work (or I have misunderstood something). The problem is, that we cannot pass sliding-window bins to the splitting function `split.network.time.based` as this function assumes that the bins are ordered and only can deal with normal (ordered, non-sliding-window) bins. As far as I have seen, there is no way to easily fix this.

However, I agree with you that we should avoid code duplication (as it is done in my commit). Maybe we can find a way to extract the duplicated code into a separate helper function. Then we would not need duplication any more :smile:

clhunsen:
Ah, you are right, @bockthom! :+1: Let me refine my idea. :smile: 

Can we use the function [the function `construct.overlapping.ranges`](https://github.com/se-passau/coronet/blob/dev/util-misc.R#L539) (as mentioned before) in conjunction with – and this is the new part – [the function `split.network.time.based.by.ranges`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-split.R#L792-L822)? 

Maybe, for the case `number.windows` is set, we need [the function `generate.date.sequence`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-misc.R#L357) with its parameter `length.out = number.windows` and [the function `construct.ranges`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-misc.R#L416) to construct proper ranges.

bockthom:
> Can we use the function [the function `construct.overlapping.ranges`](https://github.com/se-passau/coronet/blob/dev/util-misc.R#L539) (as mentioned before) in conjunction with – and this is the new part – [the function `split.network.time.based.by.ranges`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-split.R#L792-L822)?

Sounds promising. I just tried it and it seems to work, in general. However, the last two ranges constructed by the function `construct.overlapping.ranges` are different to the last two ranges constructed via the `bins.date.middle` approach in the split function. -- For the second last range,  `construct.overlapping.ranges` seems to be more precise, but the last range is just half a range and should be ignored. So, I need to further debug this what's going on for the last two ranges - maybe I need to adjust the split functions to use `construct.overlapping.ranges` instead of calculating the bins separately. In the end, both approaches have to result in exactly the same ranges, so either of them has to be fixed somehow.

Result: I will work on this and try to find out which of the two kinds of overlapping ranges is correct and which one needs to be adjusted. This will take some time, but as soon as I will have found a solution for that problem, I will try to come up with a new commit.
> 
> Maybe, for the case `number.windows` is set, we need [the function `generate.date.sequence`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-misc.R#L357) with its parameter `length.out = number.windows` and [the function `construct.ranges`](https://github.com/se-passau/coronet/blob/75ae4a505b36bd73f554a982645c82171c808618/util-misc.R#L416) to construct proper ranges.

Thanks also for your suggestions regarding`number.windows`. However, I did not deal with this problem yet. I'll postpone this until I have a fix for the original problem discussed above.

clhunsen:
> maybe I need to adjust the split functions to use `construct.overlapping.ranges` instead of calculating the bins separately. In the end, both approaches have to result in exactly the same ranges, so either of them has to be fixed somehow.

This is a good point. We should discuss this in our meeting tomorrow. :thinking:

bockthom:
While further debugging the above described and discussed problem, it turned out that there are few more problems regarding the sliding window implementation, especially at corner cases. To fix all these problems in all the various splitting functions, I have implemented the fixes in a new way and also added lots of tests for that (currently, we don't have tests for sliding-window construction at all, which makes it hard to debug.) For that reason, I close this PR now and open a new one containing the new fixes afterwards immediately. Stay tuned :smile:

PR #194, opened 2021-01-27
Mail count & Issue count

### LINKED ISSUE #185: When to do timestamp extraction for issue data? Before or after event filtering?
### Description

During various debugging activities in the last days, I stumbled over the following issue:

When performing data cutting (`get.data.cut.to.same.date`) on issue data, the outcome sometimes looks weird. Then I found out why: The timestamps are extracted *before* the filtering on `"issues.only.comments"` takes place -- see the function call `extract.timestamps` in the following code excerpt:
https://github.com/se-passau/coronet/blob/91fc448bad64dd17fa96a720322fe7d46337cc43/util-data.R#L900-L915

Hence, the data cutting function uses timestamps that may be outside the data-cutting range if they are related to an issue event which is not a comment. This might be confusing if a user is not aware of that. But then I thought about it: How can this happen? Can it happen that there are timestamps in the issue data which date back to before the first comment? -- Each issue / PR starts with a comment - so how can this happen? After doing some research, I found answer for these questions: If we have a pull request, all the commits added to this pull-request are an event themselves, the date of those events is equal to the author date of the commit. Actually adding this commit to the PR is a separate event having the timestamp of the time at which the commit was pushed. So, everything is ok, but we should think about how we want to handle this -- there are several options:

---

#### Option 1: Extract timestamps after the event-filtering on `"issues.only.comments"` has taken place.
*Advantage:* We only get the timestamps we actually use (but then it would be necessary to also write the filtered data to the data object, reducing the needs for filtering again and again whenever calling `get.issues`).

*Disadvantage:* The timestamps won't be updatable anymore. That is, as soon we have filtered the comment-events once, there is no way to return to the original data. However, we could change the implementation of the `"issues.only.comments"` configuration option in a way that it re-reads the data completely whenever changing this configuration option.

---

#### Option 2: Filter events that may pre-date the actual issue data (like `"commit_added"` events which rather belong to commit data than to issue-data)
Here we would also have two sub-options: Either remove them completely while reading the issue data, or just remove them from timestamp extraction.

*Advantage:* Comparably easy to implement, not many changes necessary.

*Disadvantage:* We may loose information on those events (depending on the sub-option) -- but this may be wanted in some kind.

---

#### Option 3: Don't change anything.

*Advantage:* Nothing to do, stop discussing it.

*Disadvantage:* Confusing behavior when there are very old commits referenced in a pull request and the pull request was initiated many months later than the commit was authored. This somehow circumvents the data cutting procedure as such old commits lead us to believe that the issue data started months ago, but actually it was just an old commit that was added to a pull request yesterday and the actual issue data started also yesterday.

---

What do you folks think about that? Which option would you prefer? I am also open for new suggestions.

### LINKED ISSUE #188: Add further vertex attributes for author vertices
### Description

Complementary to #92 and #170, we should also add further vertex attributes for author vertices that deal with other data sources than commits. At the moment, we (almost) only have commit-related vertex attributes for authors, such as
`add.vertex.attribute.commit.count.author`,
`add.vertex.attribute.commit.count.committer` (and several more of them combining author and committer information),
`add.vertex.attribute.artifact.count` (only for source-code artifacts (files, functions, etc.)

Along with them, we need to create new functions to determine 
- [x] `mail.count`, 
- [x] `mail.thread.count`, 
- [x] `issue.count` (possibly also PR count?) , 
- [x] `issue.comment.count`, etc... 

*[The idea of this issue is rather old, but it has not been documented properly yet. So I just opened this issue to not forget about it.]*

### Ideas for the Implementation
To get the mail count etc. as values, we need new functions like `get.author.mail.count`. To achieve that, see also the following comment: https://github.com/se-sic/coronet/issues/70#issuecomment-424668069

For the implementation of the vertex attributes, see the current implementations for the commit-count attribute [L156ff](https://github.com/se-sic/coronet/blob/91fc448bad64dd17fa96a720322fe7d46337cc43/util-networks-covariates.R#L156ff) and for the artifact-count attribute [L385ff](https://github.com/se-sic/coronet/blob/91fc448bad64dd17fa96a720322fe7d46337cc43/util-networks-covariates.R#L385ff).

JoJoDeveloping:
### Prerequisites

- [X] All tests pass

### Description

Adds methods that add relevant attributes to implement the changes noted in #188 
These are per-author and include:
* issue count
* issue comment count
* mail count
* mail thread count
* PR count
* PR comment count

The definition of when an author is related to an issue are tentative and should be looked over. In particular we have:
* An issue is counted for someone if they participated in that issue in any way (i.e. labeling, commenting, opening/closing, mentioning)
* An issue is counted for someone if they comment in it
* A PR is counted for the person opening it
* A PR is counted for a person when they comment in it (this includes the initial author, mostly, as they write the initial description)
* A PR is counted for a person when they participate it in any way (i.e. labeling, commenting, opening/closing, mentioning, _adding comments to it_)

As an added bonus, we also fix #185. We also rename the issue data getters in util-data.R so that they are named consistently with the commit data getters.

Merging this closes #185, closes #188.


JoJoDeveloping:
I don't know why CI fails. I can build it locally and run the tests just fine. The CI also only fails for version 3.3 with an error indicating sqldf is missing (?).

util-core-peripheral.R
```diff
@@ -705,58 +705,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
+        logging::logdebug(paste(name, ": starting.", sep=""))
+
+        df = data.extractor(proj.data)
+        if (distinctize) {
+            df = df[!duplicated(df[[distinctize.key]]), ]
+        }
+        df = df[grouping.keys]
+        grouping.keys.formatted = paste(grouping.keys, sep="`, `")
+        stmt = paste("SELECT `",grouping.keys.formatted,"`, COUNT(*) as `freq` FROM `df`
+                                 GROUP BY `",grouping.keys.formatted,"` ORDER BY `freq` DESC, `",grouping.keys.formatted,"`",sep="")
```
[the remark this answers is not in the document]
JoJoDeveloping: Well, yes, but actually no since `ASC` is the default ordering used.

util-core-peripheral.R
```diff
@@ -705,58 +705,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
+    logging::logdebug("get.author.mail.thread.count: starting.")
+
+    df = proj.data$get.mails()
+    df = df[!duplicated(df[["message.id"]]), ]
+    df = df[c("author.name", "message.id", "thread")]
+    stmt = "SELECT `author.name`, COUNT(DISTINCT thread) as `freq` FROM `df`
+                             GROUP BY `author.name` ORDER BY `freq` DESC, `author.name` ASC"
+    logging::logdebug(paste("get.author.mail.thread.count: running SQL ", stmt,sep=""))
+    res = sqldf::sqldf(stmt)
+    logging::logdebug("get.author.mail.thread.count: finished")
```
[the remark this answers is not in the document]
JoJoDeveloping: Yes, the reason was that I did not want to make the helper function too fancy. Adding this either requires specifying part of the SQL query as a string, or at least two args (does R have option types?). Making the function too general means one has to always look up what it does, which makes things harder.

hechtlC (changes requested):
Nice work @JoJoDeveloping . I found a couple of additional things which were not mentioned in @bockthom's review.

util-core-peripheral.R
```diff
@@ -705,58 +705,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
-    ## Get commit data
-    commits.df = proj.data$get.commits.filtered()
+#' Get the issues count for each author based on the issues data contained in the specified \code{ProjectData}.
+#' The issues count here is the number of issues the author participated in by commenting
+#'
+#' @param proj.data the \code{ProjectData} containing the mail data
+#'
+#' @return a dataframe consisting of two columns, the first of which holding the authors' names and the second holding
+#'         their respective mail counts
+get.author.issues.commented.in.count = function(proj.data) {
```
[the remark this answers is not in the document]
hechtlC: I was not involved in this as far as I can remember (I might be wrong though).

util-core-peripheral.R
```diff
@@ -705,58 +705,289 @@ get.committer.or.author.commit.count = function(range.data) {
     return(res)
 }
 
+
+#' Helper function that aggregates counts of things like commits, mails, ... on a per-author basis
+#'
+#' @param name the name the function will be bound to, for logging
+#' @param data.extractor a function to be called on the main argument that gets the dataframe from which to aggregate from
+#' @param grouping.keys the dataframe keys to group by
+#' @param distrinctize Whether to remove duplicates
```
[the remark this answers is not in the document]
clhunsen: Sounds like a typo for me. I guess this verb should come from `distinct` or `distinctive` (so, without an `r`!). In the end and for me, `remove.duplicates` or just `distinct` seem to be better choices.

util-core-peripheral.R
```diff
@@ -705,58 +705,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
+#' @param distrinctize.key if distinctize, then the key by which to remove duplicates
+#'
+#' @return A function that aggregates data according to the above specification contained in a given \code{ProjectData}.
+#'         This function itself returns a dataframe consisting of |grouping.keys|+1 columns, the last holding the count,
+#'         and the others the respective grouping
+#'
+group.data.by.key = function(name, data.extractor, grouping.keys, distinctize, distinctize.key) {
+    return (function(proj.data) {
+        logging::logdebug(paste(name, ": starting.", sep=""))
```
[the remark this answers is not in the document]
clhunsen: The functions `logging::log*` have `sprintf` built-in (see [here](https://github.com/mfrasca/r-logging/blob/master/pkg/R/utils.R#L165) and also look for `logging-entrypoints` in [the documentation](https://cran.r-project.org/web/packages/logging/logging.pdf), parameter `...` in particular). Thus, you could use the following rightaway:
```R
logging::logdebug("%s: starting.", name)
```

util-core-peripheral.R
```diff
@@ -705,58 +706,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
+#'         their respective mail thread counts
+get.author.mail.thread.count = function(proj.data) {
+    logging::logdebug("get.author.mail.thread.count: starting.")
+
+    df = proj.data$get.mails()
+    df = df[!duplicated(df[["message.id"]]), ]
+    df = df[c("author.name", "message.id", "thread")]
+    stmt = "SELECT `author.name`, COUNT(DISTINCT thread) as `freq` FROM `df`
+                             GROUP BY `author.name` ORDER BY `freq` DESC, `author.name` ASC"
+    logging::logdebug(paste("get.author.mail.thread.count: running SQL ", stmt,sep=""))
```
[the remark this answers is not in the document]
clhunsen: ... or use the built-in `sprintf` of `logging::log*` functions rightaway (see my comment above).

util-core-peripheral.R
```diff
@@ -705,58 +706,289 @@ get.committer.or.author.commit.count = function(range.data) {
...
+#'
+#' @return a dataframe consisting of two columns, the first of which holding the authors' names and the second holding
+#'         their respective mail counts
+get.author.pull.requests.count = function(proj.data) {
+    logging::logdebug("get.author.pull.requests.count: starting.")
+
+    df = proj.data$get.issues.unfiltered()
+    df = df[c("author.name", "issue.id", "issue.type", "event.name")]
+    df.new = data.frame(author.name=character(),issue.id=character())
+    for (rnum in 1:nrow(df)) {
```
clhunsen, line 953: Please use `seq_len(nrow(df))`. If the data.frame has no rows, `1:0` yields `c(1, 0)` and, thus, wrong computations. My code snippet yields `integer(0)` and no iteration in the loop at all.

This holds also for other instances of this pattern!

util-data.R
```diff
@@ -959,6 +959,22 @@ ProjectData = R6::R6Class("ProjectData",
             }
         },
 
+        #' Get the issue data, unfiltered.
+        #' Unfiltered means that we ignore the "issues.only.comments" config and always return the full data
+        #' If it does not already exist call the read method.
+        #'
+        #' @return the issue data
+        get.issues.unfiltered = function() {
```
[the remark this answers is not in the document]
clhunsen: Please consider [my comment](https://github.com/se-sic/coronet/issues/185#issuecomment-774655806) on Issue #185 in your discussions.

Also, shouldn't the function `ProjectData$get.issues` rely on `ProjectData$get.issues.unfiltered` to avoid code duplication and potential problems when extending the functionality of these functions?

clhunsen (commented):
Thanks for your PR, @JoJoDeveloping! :smiley: Keep up the good work!

This is _not_ a full review as I have not looked into the implementation in detail. I just want to share my thoughts on some things that I spotted while scanning the comments of @bockthom and @hechtlC. :wink:

tests/codeface-data/results/testing/test_feature/feature/issues-github.list
```diff
@@ -15,3 +15,15 @@
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"subscribed";"Björn";"bjoern@example.org";"2016-12-07 15:30:02";"udo";"""udo@example.org"""
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"labeled";"Olaf";"olaf@example.org";"2017-05-23 12:31:34";"decided";""""""
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"commented";"Björn";"bjoern@example.org";"2017-05-23 12:32:39";"open";"[]"
+"1";"Example pull request 1";"[""pull request""]";"closed";"[]";"2013-05-13 02:46:44";"2013-09-05 14:48:10";"[]";"created";"Thomas";"thomas@example.org";"2016-07-12 15:59:25";"open";"[]"
```
JoJoDeveloping, line 18: These pull requests are the reason tests are failing. Unfortunately, I do not understand what some of the tests are even testing. We might need to discuss this next meeting.

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+                                                          default.value = 0L, issue.type = "all") {
+    nets.with.attr = add.vertex.attribute.commit.count.helper(
+        list.of.networks, project.data, name, aggregation.level,
+        default.value, function(data) {return (get.author.issues.commented.in.count(data, type = issue.type))}, "author.name"
+    )
+
+    return(nets.with.attr)
+}
+
+#' Add issues-count attribute based on the number of issues created in where the person represented by the vertex is the author.
```
hechtlC, line 463: I guess the 'in' can be removed here.

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+#' @return A list of networks with the added attribute
+add.vertex.attribute.issue.creation.count = function(list.of.networks, project.data,
+                                                          name = "issue.creation.count",
+                                                          aggregation.level = c("range", "cumulative", "all.ranges",
+                                                                                "project.cumulative", "project.all.ranges",
+                                                                                "complete"),
+                                                          default.value = 0L, issue.type = "all") {
+    nets.with.attr = add.vertex.attribute.commit.count.helper(
+        list.of.networks, project.data, name, aggregation.level,
+        default.value, function(data) {return (get.author.issues.created.count(data, type = issue.type))}, "author.name"
```
hechtlC, line 485: No space between return and the (

util-split.R
```diff
@@ -20,6 +20,7 @@
 ## Copyright 2017-2018 by Thomas Bock <bockthom@fim.uni-passau.de>
 ## Copyright 2020 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
+## Copyright 2021 by Johannes Hostert <s8johost@stud.uni-saarland.de>
```
hechtlC, line 23: We already talked about this.

hechtlC (changes requested):
First of all thanks for the good work on this enhancement @JoJoDeveloping. 

There are some problems here (mostly documentation). These will need to be fixed. Please also check whether there are more than one occurrence of the documentation problems as I have not marked every one.

Please also don't forget to address the old comments.

tests/codeface-data/results/testing/test_feature/feature/issues-github.list
```diff
@@ -15,3 +15,16 @@
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"subscribed";"Björn";"bjoern@example.org";"2016-12-07 15:30:02";"udo";"""udo@example.org"""
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"labeled";"Olaf";"olaf@example.org";"2017-05-23 12:31:34";"decided";""""""
 6;"Distinguish directedness of networks and edge-construction algorithm";"[""issue"", ""bug"", ""enhancement""]";"open";"[]";"2016-07-12 14:30:13";"";"[]";"commented";"Björn";"bjoern@example.org";"2017-05-23 12:32:39";"open";"[]"
+"1";"Example pull request 1";"[""pull request""]";"closed";"[]";"2016-07-14 13:37:00";"2017-01-01 01:01:01";"[]";"created";"Thomas";"thomas@example.org";"2016-07-12 15:59:25";"open";"[]"
```
bockthom, line 18: Is there any reason for quoting the id in the first column?

tests/codeface-data/results/testing/test_feature/feature/issues-github.list
```diff
@@ -15,3 +15,16 @@
...
+"1";"Example pull request 1";"[""pull request""]";"closed";"[]";"2016-07-14 13:37:00";"2017-01-01 01:01:01";"[]";"commented";"Thomas";"thomas@example.org";"2016-07-12 15:59:25";"open";"[]"
+"1";"Example pull request 1";"[""pull request""]";"closed";"[]";"2016-07-14 13:37:00";"2017-01-01 01:01:01";"[]";"state_updated";"Thomas";"thomas@example.org";"2016-07-12 15:59:59";"closed";"""open"""
```
bockthom, line 20: Please try to use realistic scenarios. Updating the state from open to closed but having commented events afterwards in which the state is open?
As this are pull requests, it may additionally be helpful to also incorporate traditional PR events, such as "merged" etc. :wink:

tests/test-networks-covariates.R
```diff
@@ -486,6 +491,254 @@ test_that("Test add.vertex.attribute.commit.count.committer.or.author", {
...
+        actual.attributes = lapply(networks.with.attr, igraph::get.vertex.attribute, name = "issue.count")
+
+        expect_identical(expected.attributes.issues.only[[level]], actual.attributes)
+    })
+
+    # Test PRs only
+
+    lapply(AGGREGATION.LEVELS, function(level) {
+        networks.with.attr = add.vertex.attribute.issue.count(
+            networks.and.data[["networks"]], networks.and.data[["project.data"]], aggregation.level = level, issue.type = "pull.requests", name = "pull.request.count"
```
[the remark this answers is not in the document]
bockthom: In such function calls, we should break lines after 120 characters. We only make exceptions when setting test data or expected data, for comprehensibility reasons.

util-core-peripheral.R
```diff
@@ -705,55 +706,241 @@ get.committer.or.author.commit.count = function(range.data) {
     return(res)
 }
 
+
+#' Helper function that aggregates counts of things like commits, mails, ... on a per-author basis.
+#'
+#' For example, called with name="commit.count", data.extractor=function(proj.data) {return(proj.data$get.commits.filtered())},
```
bockthom, line 712: Please also respect our coding conventions within comments, that is, add spaces around `=` etc.
As this actually is code, please use \code{...}. (We do this to achieve automated rendering in typewriter font when we will generate documentation out of this comments in far far future :smile:)

util-core-peripheral.R
```diff
@@ -705,55 +706,241 @@ get.committer.or.author.commit.count = function(range.data) {
...
+#' grouping.keys=c("committer.name"), distinctize=true and distinctize.key=c("hash"), the returned function will
+#'
+#' 1. get the data frame using the extractor (in this case, the commit df)
+#' 2. remove duplicate entries so that there is only one entry per commit hash
+#' 3. project away unneeded columns, leaving only committer.name
+#' 4. count the commits grouped by the commiter name
+#' 5. return a dataframe with cols commiter.name and freq, which contains the number of commits authored by each author
```
bockthom, line 719: Please add quotation marks around colnames

util-core-peripheral.R
```diff
@@ -705,55 +706,241 @@ get.committer.or.author.commit.count = function(range.data) {
...
+        #if necessary, make sure that there is only one entry for each remove-duplicate key (combination)
+        if (remove.duplicates) {
+            df = df[!duplicated(df[[remove.duplicates.by]]), ]
+        }
+        #throw away unnecessary columns
+        df = df[grouping.keys]
+        grouping.keys.formatted = paste(grouping.keys, sep="`, `")
+        #execute a query that counts the number of occurrences of the grouping.keys
+        stmt = paste0("SELECT `", grouping.keys.formatted, "`, COUNT(*) as `freq` FROM `df`
+                                 GROUP BY `", grouping.keys.formatted, "` ORDER BY `freq` DESC, `", grouping.keys.formatted, "`")
```
bockthom, line 745: Here the indentation seems to be broken.

util-core-peripheral.R
```diff
@@ -705,55 +706,241 @@ get.committer.or.author.commit.count = function(range.data) {
...
+        logging::logdebug(paste0(name, ": running SQL ", stmt))
```
bockthom, line 746: Please omit `paste0` here and use the syntax you also have used two lines below :wink:

util-core-peripheral.R
```diff
@@ -705,55 +706,241 @@ get.committer.or.author.commit.count = function(range.data) {
...
+    # so we abort beforehand
+    if (nrow(df) == 0) {
+        return (df[c("author.name", "issue.id", "event.name")]);
+    }
+    switch (type,
+            all = {
+                df = df[c("author.name", "issue.id", "event.name")]
+            },
+            issues = {
+                df = df[sapply(df[["issue.type"]], function (k) {return ("issue" %in% k)}), c("author.name", "issue.id", "event.name")]
```
bockthom, line 838: Please don't copy-paste the vector `c("author.name", "issue.id", "event.name")` that often in this function. Just assigning it to variable and reusing the variable will do it :wink: 

In addition, there should be no space between `function` and its parameters.

util-read.R
```diff
@@ -342,6 +342,15 @@ read.issues = function(data.path, issues.sources = c("jira", "github")) {
     issue.data[["date"]] = get.date.from.string(issue.data[["date"]])
     issue.data[["creation.date"]] = get.date.from.string(issue.data[["creation.date"]])
     issue.data[["closing.date"]] = get.date.from.string(issue.data[["closing.date"]])
+
+    # fix all dates to be after the creation date.
+    # violations can happen for commit_added events if the commit was made before the PR was opened
+    # the original date for commit_added events is stored in event.info.2 in any case
```
bockthom, line 348: Please use quotation marks around "commit_added" and also for column names

bockthom (changes requested):
Thanks @JoJoDeveloping for working on this PR. I just had a quick view at it and spotted only minor issues. I will do a full pass when all the remaining issues (especially to the test suite) are fixed.

I have already spotted a few issues:

1. This PR contains files which should not be part of it, e.g., `tests/testthat-problems.rds`. Please remove them via rebase, as we don't want such files to show up in the commit history. Please also check if there are other files that should not appear here.
2. There still are some violations of the coding conventions (spacing, indentation, etc.). The individual comments below just indicate some of them. Please also check for similar violations not explicitly mentioned in our reviews.
3. The test data for pull-requests contains unrealistic data. Please adjust events in such a way that they can occur in a real setting, see also my comments below.

Please see also the individual comments below.

util-core-peripheral.R
bockthom: Just to not forget about it: When adding a new file `util-data-misc.R` for the issue-data related and generally usable helper functions, please don't forget to add the new file to the [File/Module overview](https://github.com/se-sic/coronet#filemodule-overview) in the `README.md`.
I would suggest to add the new file directly below `util-data.R` and phrase the description of the new file in a similar way than for `util-networks-misc.R`.

And also please remember that the new file needs to be added to `util-init.R`.

bockthom (changes requested):
I had a closer look at your pull request and have identified a few more issues. In the NEWS.md, there are still the issues which we have already identified in our past reviews, please have a look at them again.

In addition, please also have a look at the new comments below.

clhunsen (commented):
Thanks for your work on this PR, @JoJoDeveloping! :+1:  Please have a look on my comments, I just wanted to add my two cents on your changes: mostly some things for a good discussion with @bockthom and @hechtlC. :wink:

util-core-peripheral.R
```diff
@@ -705,55 +706,205 @@ get.committer.or.author.commit.count = function(range.data) {
...
+    df = proj.data$get.mails()
+    df = df[!duplicated(df[["message.id"]]), ]
+    df = df[c("author.name", "message.id", "thread")]
+    stmt = "SELECT `author.name`, COUNT(DISTINCT thread) as `freq` FROM `df`
+                             GROUP BY `author.name` ORDER BY `freq` DESC, `author.name` ASC"
+    logging::logdebug("get.author.mail.thread.count: running SQL %s", stmt)
+    res = sqldf::sqldf(stmt)
+    logging::logdebug("get.author.mail.thread.count: finished")
+    return(res)
+}
```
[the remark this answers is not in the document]
bockthom: Sounds good! I'd support this idea :+1:

util-init.R
```diff
@@ -59,6 +59,7 @@ source("util-split.R")
 source("util-motifs.R")
 source("util-bulk.R")
 source("util-plot.R")
+source("util-data-misc.R")
```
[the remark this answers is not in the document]
JoJoDeveloping: Which line, directly below what line?

hechtlC (changes requested):
Thanks again for the nice additions and addressed issues.

I just found a couple more small issues with the PR. Once these are addressed (and also the outstanding ones), I will be fine with merging.

tests/test-networks-covariates.R
```diff
@@ -486,6 +492,258 @@ test_that("Test add.vertex.attribute.commit.count.committer.or.author", {
...
+    networks.and.data = get.network.covariates.test.networks()
+
+    expected.attributes = list(
+        range = network.covariates.test.build.expected(c(1L), c(0L), c(1L, 0L, 0L)),
+        cumulative = network.covariates.test.build.expected(c(1L), c(1L), c(2L, 0L, 1L)),
+        all.ranges = network.covariates.test.build.expected(c(1L), c(2L), c(2L, 0L, 1L)),
+        project.cumulative = network.covariates.test.build.expected(c(3L), c(1L), c(2L, 0L, 1L)),
+        project.all.ranges = network.covariates.test.build.expected(c(3L), c(2L), c(2L, 0L, 1L)),
+        complete = network.covariates.test.build.expected(c(3L), c(2L), c(2L, 0L, 1L))
+    )
```
bockthom, line 534: Unfortunately, in this test, we only consider mail threads which exactly consist of one single e-mail. So, the test result is exactly the same as for testing mail count. So, I am not convinced that the test really tests for the thread count. Let's think about whether there is an easy solution to test also for e-mail threads having more than one e-mail – without adding new data which would force us to adjust all the data/network tests... Let's discuss about the test data in our next meeting.

tests/test-networks-covariates.R
```diff
@@ -486,6 +492,258 @@ test_that("Test add.vertex.attribute.commit.count.committer.or.author", {
...
+
+    expected.attributes.prs.only = list(
+        range = network.covariates.test.build.expected(c(0L), c(3L), c(1L, 0L, 0L)),
+        cumulative = network.covariates.test.build.expected(c(0L), c(3L), c(3L, 0L, 1L)),
+        all.ranges = network.covariates.test.build.expected(c(2L), c(3L), c(3L, 0L, 1L)),
+        project.cumulative = network.covariates.test.build.expected(c(1L), c(3L), c(3L, 0L, 1L)),
+        project.all.ranges = network.covariates.test.build.expected(c(3L), c(3L), c(3L, 0L, 1L)),
+        complete = network.covariates.test.build.expected(c(3L), c(3L), c(3L, 0L,  1L))
+    )
```
bockthom, line 570: We should also test for the "all" variant here and also for all the other issue-related test functions here. Should be easy to achieve, as we just need to sum up expected attributes for "issues.only" and "prs.only".

util-data-misc.R
```diff
@@ -0,0 +1,421 @@
...
+#' The type argument specifies whether we count PRs alone, issues alone, or both (\code{"all"}).
+#'
+#' @param proj.data the \code{ProjectData} containing the issue data
+#' @param type which issue type to consider. see \code{preprocess.issue.data}
+#'             One of \code{"issues"}, \code{"pull.requests"} or \{"all"}
+#'             [default: "all"]
+#'
+#' @return a dataframe consisting of two columns, the first of which holding the authors' names and the second holding
+#'         their respective issue counts
+get.author.issue.count = function(proj.data, type = "all") {
```
bockthom, line 341: Let's talk about the `type` parameter in our next meeting. I just have a few questions here – but nothing to worry about.

util-data.R
```diff
@@ -1066,7 +1069,25 @@ ProjectData = R6::R6Class("ProjectData",
             private$authors = data
         },
 
-        #' Get the issue data.
+        #' Get the issue data, filtered according to options in the project configuration.
+        #' If it does not already exist call the read method.
+        #'
+        #' @return the issue data
+        get.issues.filtered = function() {
```
bockthom, line 1076: Just a reminder for the discussion in our next meeting:
For getting filtered commit data, there is an `uncached` function variant.
@hechtlC Should we also add such a function for filtering issues `uncached`, to make both filtering procedures consistent? I am not really sure about that. This is open for discussion.

bockthom (commented):
Fortunately, I eventually found the time to have a closer look at the count-related tests of this PR. While comprehending all these tests in detail, I reflected also about some other implementation details. In sum, I really like this PR and all the implementations. Really nice work!

I just spotted a few highly relevant issues which we need to discuss in our next meeting, and also some minor documentation-related issues (as always when performing reviews :wink:).
Please find my detailed comments below. I am looking forward to discussing some of them in our next meeting in detail.

util-data-misc.R
```diff
@@ -0,0 +1,421 @@
...
+#' @param retained.rows the rows to be retained. [default: "c("author.name", "issue.id", "event.name")"]
+#' @param type which issue type to consider.
+#'             One of \code{"issues"}, \code{"pull.requests"} or \code{"all"}
+#'             [default: "all"]
+#'
+#'
+preprocess.issue.data = function(proj.data, retained.rows = c("author.name", "issue.id", "event.name"),
+                                 type = c("all", "pull.requests", "issues")) {
+  type = match.arg(type)
+  df = proj.data$get.issues()
```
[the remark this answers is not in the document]
JoJoDeveloping: 1. One really wants to use the unfiltered data, since the usual filters that `.filtered` uses are ones like "only retain `"comment"` events", which would mean that methods like `get.author.issue.creation.count` and related would assing `0` for each author since all issue creation events got filtered.

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+    return(nets.with.attr)
+}
+
+## * Issue / PR count --------------------------------------------------------------
+
+#' Add issues-count attribute based on the number of issues participated in, where the person represented by the vertex is the author.
+#'
+#' @param list.of.networks The network list
+#' @param project.data The project data
+#' @param name The attribute name to add. You might want to change this [default: "issues.count"]
```
bockthom, line 411: Wrong default value. Should be "issue.count"

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+    )
+
+    return(nets.with.attr)
+}
+
+#' Add issues-count attribute based on the number of issues participated in by commenting, where the person represented by the vertex is the author.
+#'
+#' @param list.of.networks The network list
+#' @param project.data The project data
+#' @param name The attribute name to add [default: "issue.count.by.commenting"]
```
bockthom, line 439: Wrong default value. Should be "issue.commented.count"

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+                                                          default.value = 0L, issue.type = "all") {
+    nets.with.attr = add.vertex.attribute.commit.count.helper(
+        list.of.networks, project.data, name, aggregation.level,
+        default.value, function(data) {return(get.author.issues.commented.in.count(data, type = issue.type))}, "author.name"
+    )
+
+    return(nets.with.attr)
+}
+
+#' Add issues-count attribute based on the number of issues created, where the person represented by the vertex is the author.
```
bockthom, line 463: Line length and "issue-count"

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+#'
+#' @param list.of.networks The network list
+#' @param project.data The project data
+#' @param name The attribute name to add [default: "issues.count.by.commenting"]
```
bockthom, line 467: Wrong default value. Should be "issue.creation.count"

util-networks-covariates.R
```diff
@@ -345,6 +346,176 @@ add.vertex.attribute.commit.count.helper = function(list.of.networks, project.da
...
+                                                          default.value = 0L, issue.type = "all") {
+    nets.with.attr = add.vertex.attribute.commit.count.helper(
+        list.of.networks, project.data, name, aggregation.level,
+        default.value, function(data) {return(get.author.issues.created.count(data, type = issue.type))}, "author.name"
+    )
+
+    return(nets.with.attr)
+}
+
+#' Add issue-comments-count attribute based on the number of issues participated in by commenting, where the person represented by the vertex is the author.
```
bockthom, line 491: Line length and "issue-comment-count"

util-read.R
```diff
@@ -342,6 +343,15 @@ read.issues = function(data.path, issues.sources = c("jira", "github")) {
...
     issue.data[["creation.date"]] = get.date.from.string(issue.data[["creation.date"]])
     issue.data[["closing.date"]] = get.date.from.string(issue.data[["closing.date"]])
+
+    ## fix all dates to be after the creation date.
+    ## violations can happen for "commit_added events" if the commit was made before the PR was opened
+    ## the original date for "commit_added" events is stored in "event.info.2" in any case
+    commit.added.events = issue.data[["event.name"]] == "commit_added"
+    issue.data[commit.added.events, "event.info.2"] = get.date.string(issue.data[commit.added.events, "date"])
+    commit.added.events.before.creation = commit.added.events & (issue.data["date"] < issue.data["creation.date"])
+    issue.data[commit.added.events.before.creation, "date"] = issue.data[commit.added.events.before.creation, "creation.date"]
```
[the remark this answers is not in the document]
bockthom: This one is still missing: If the "creation.date" is missing (e.g., ""), the we agreed to not perform the date fixes (neither update "event.info.2", nor "date").

tests/test-networks-covariates.R
```diff
@@ -46,8 +47,13 @@ myranges = construct.ranges(mybins, sliding.window = FALSE)
...
 #' Load test data and generate test networks
 #'
+#' @param network.type which network to get (\code{"author"} or \code{"artifact"}). [default: "author"]
+#' @param issues whether to retain issue data. If \code{FALSE}, issue data is deleted. [default: FALSE]
+#' @param author.relation passed to the network config. [default: "cochange"]
+#'
 #' @return Tuple containing project data and list of networks
-get.network.covariates.test.networks = function(network.type = c("author", "artifact")) {
+get.network.covariates.test.networks = function(network.type = c("author", "artifact"), issues = FALSE,
+                                                author.relation = "cochange") {
```
bockthom, line 56: Please provide the possible options here: `author.relation = c("cochange", "issue", "mail")` and add `author.relation = match.arg(author.relation)`.

tests/test-networks-covariates.R
bockthom: Unfortunately, the mail-count related tests are not changed to use mail data :disappointed: 
(Instead, you adjusted all the issue-count related tests to use issue data, which has not been necessary but time consuming..... :angry:).

NEWS.md
```diff
@@ -16,12 +16,17 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, d1e4413a49ab83a115a7f26719592876371ab264)
 - Add tests for the `get.data.sources.from.relations` function (PR #195, add0c746dde8279da41d180deecf52b91a46095c)
 - Add logo directory containing several logo variants (PR #196, 82f99719b8a43b0a44914b67b26bf1a34bb076c6, dc4659ea354e97159f8ee6167811db544f0b7087, fdc5e677325225f92d1f99948cb9441bfe1d150d, 752a9b376ffeffd5d6b380b4fdba838a890e3ef7)
+- Add per-author vertex attributes regarding issue/pull request/mail count and derived quantities (like mail thread count, issue creation count) (PR #194, 9f9150a97ffbb64607df0ddcbce299e16c2580da, 7260d62cf6f1470584753f76970d19664638eeed, 139f70b67903909bcd4c57e26afa458681a869f2, eb4f649c04155e22195627072e0f08bb8fe65dc4, 627873c641410182ca8fee0e78b95d7bda1e8e6b, 1e1406f4a0898cac3e61a7bc9a5aa665dceef79f, 98e11abc651b5fe0ec994eddea078635b0d6f0b2, cc64b9b3e3f3e33798a6d138225f745bd2f550a0)
```
bockthom, line 19: Please use some hyphens here for clarification:
"issue/pull-request/mail count"
"mail-thread count"
"issue-creation count"
"issue-comment count"
and also reference not only PR 194, but also the related issue number.

Maybe it is also worth to mention the function `preprocess.issue.data` (and what it can be used for) explicitly in a separate item (before the vertex attributes).

In addition, please add an additional item stating that we have added a new function `get.issues.filtered.uncached`.

NEWS.md
```diff
@@ -16,12 +16,17 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
 - Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (PR #195, cf795f26652b00de5d717c703c688af55a972943, 70c05ecd1e3c0f10810acc2b2ae06a3eb8856317, 5a46ff4d428af7f301fe57d6e9e10421f553a9cc, fd767bb37ca608c28d9ff4a449415cc0e863d7ee)
+- Rename `get.issues` to `get.issues.filtered`, and write a new `get.issues` to get the unfiltered issues so that these methods follow the naming scheme known from the respective methods for commits (b9dd94c8575b8cab40d0d1185368854f84299d87)
+
+### Fixed
+- Fix fencing issue timing data so that all issue actions "happen" after the issue was created (issue #185, 627873c641410182ca8fee0e78b95d7bda1e8e6b, 6ff585d9da1da3432668605f0c09f8e182ad0d2f)
```
bockthom, line 29: Please talk about events instead of actions.
Unfortunately the description here does not correspond to the implementation: In the implementation we only adjust all the "commit_added" events but not "all" events in general (even if these was the purpose, this is not what we have implemented). And also "after" is not correct–we just make sure that it is "not before", so same time is still possible. Please also add the information that the original dates are not completely removed but now stored as event.info.2.

(And don't forget to adjust hashes in case they are outdated.)

bockthom (changes requested):
Yet another review on my part :smiley: 

In the end, I have added way more comments than I wanted to add... but I think that we are close to merging, though, if the comments are adressed.  However, the longer the time span between two subsequent reviews, the higher the probability that I will spot new issues that I have not spotted yet :wink:
Maybe you would like make use of this information when addressing my comments :smile: 
Anyway, please find my detailed comments below. If anything is unclear, don't hesitate to ask.

util-data-misc.R
```diff
@@ -0,0 +1,407 @@
...
+#'
+#' @param name the name the function will be bound to, for logging
+#' @param data.source one of \code{"commits"}, \code{"mails"}, \code{"issues"}
+#' @param grouping.keys the dataframe keys to group by
+#' @param remove.duplicates whether to remove duplicates
+#' @param remove.duplicates.by if \code{remove.duplicates}, then the key by which to remove duplicates
+#'
+#' @return a function that aggregates data according to the above specification contained in a given \code{ProjectData}.
+#'         This function itself returns a dataframe consisting of |grouping.keys|+1 columns, the last holding the count,
+#'         and the others the respective grouping
```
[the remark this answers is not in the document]
JoJoDeveloping: Their length is exactly 120 :tada:

util-data-misc.R
```diff
@@ -0,0 +1,407 @@
...
+    ## count distinct since an author may appear in the same issues at the same time
+    stmt = "SELECT `author.name`, COUNT( DISTINCT `issue.id`) as `freq` FROM `df`
+                             WHERE `event.name` = 'commented'
+                             GROUP BY `author.name` ORDER BY `freq` DESC, `author.name` ASC"
+    res = sqldf::sqldf(stmt)
+    logging::logdebug("get.author.issues.commented.in.count: finished")
+    return(res)
+}
+
+#' Get the issue/pr comment count for each author based on the issues data contained in the specified \code{ProjectData}.
```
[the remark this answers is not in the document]
JoJoDeveloping: It has length 112

util-data-misc.R
```diff
@@ -11,24 +11,32 @@
...
 
+## Copyright 2017 by Mitchell Joblin <mitchell.joblin@uni-passau.de>
+## Copyright 2017 by Ferdinand Frank <frankfer@fim.uni-passau.de>
+## Copyright 2017 by Sofie Kemper <kemperso@fim.uni-passau.de>
+## Copyright 2017-2020 by Claus Hunsen <hunsen@fim.uni-passau.de>
+## Copyright 2017 by Felix Prasse <prassefe@fim.uni-passau.de>
+## Copyright 2018-2019 by Christian Hechtl <hechtl@fim.uni-passau.de>
+## Copyright 2018 by Klara Schlüter <schluete@fim.uni-passau.de>
+## Copyright 2019 by Thomas Bock <bockthom@fim.uni-passau.de>
+## Copyright 2019 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
```
bockthom, line 22: Adjusting the copyright header does not mean to just copy the copyright header of the other file. Only take the copy-right information for the functions you have moved from the previous file. Therefore, you need to have a look at the git blame **and** at the commit history of the file `util-core-peripheral.R` to see which persons have been involved (and when they have been involved) in editing the moved functions.
To make a long story short, I had a quick look at the history of `util-core-peripheral.R` right now and figured out that the following people have **not** been involved in editing the moved functions, so please *remove* them from the copy right header of the file `util-data-misc.R` (ideally using a rebase so that these people will never show up in the copyright header in the history of the `util-data-misc.R` file):
Mitchell, Christian, Thomas.

util-data-misc.R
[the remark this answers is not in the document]
bockthom: Lucky you.

util-data-misc.R
JoJoDeveloping: Thanks for looking at that history!

util-data-misc.R
```diff
@@ -348,22 +362,22 @@ get.author.issues.created.count = function(proj.data, type = c("all", "issues",
...
     stmt = "SELECT `author.name`, COUNT( DISTINCT `issue.id`) as `freq` FROM `df`
                              WHERE `event.name` = 'created'
                              GROUP BY `author.name` ORDER BY `freq` DESC, `author.name` ASC"
     res = sqldf::sqldf(stmt)
     logging::logdebug("get.author.issues.created.count: finished")
     return(res)
 }
-
-#' Get the issue/pr count for each author based on the issues data contained in the specified \code{ProjectData}.
+0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901
```
bockthom, line 373: What's that?

util-data.R
```diff
@@ -148,6 +148,24 @@ ProjectData = R6::R6Class("ProjectData",
...
         },
+        ## * * issue filtering --------------------------------------------
+
+        #' Filter issue by potentially removing all issue events that are not comments.
+        #'
+        #' @param issues the data.frame of issues on which filtering will be applied
+        #' @param issues.only.comments flag whether non-comment issue events are removed
+        #'
+        #' @return the commits after all filters have been applied
+        filter.issues = function(issues, issues.only.comments) {
```
bockthom, line 159: You return issues, not commits :wink: 
And if you already touch this again, please also add a blank line before the `## * * issue filtering` line. Thanks!

util-read.R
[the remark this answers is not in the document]
bockthom: This one still seems to be not completely fixed, if I am correct. Missing "create.date" values indicated by an *empty string* are still not catched...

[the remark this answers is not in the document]
JoJoDeveloping: In the current code, we do
```R
    commit.added.events.before.creation = commit.added.events &
        !is.na(issue.data["creation.date"]) & (issue.data["date"] < issue.data["creation.date"])
```

Which works, because `issue.data["creation.date"]` has already been passed through `get.date.from.string` (see line 344), which returns `NA` when given an empty string. Thus catching `NA` suffices.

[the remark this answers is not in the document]
bockthom: Ok, I take back what I said. I just overlooked that the `get.date.from.string` is called before. (Sorry for my previous comment.)
I guess it is time for sleeping now instead of posting wrong comments...

showcase.R
```diff
@@ -186,7 +186,7 @@ y = NetworkBuilder$new(project.data = y.data, network.conf = net.conf)
 # sample.cumulative = add.vertex.attribute.commit.count.author(my.networks, x.data, aggregation.level = "cumulative")
 # ## add email-address vertex attribute
 # sample.mail = add.vertex.attribute.author.email(my.networks, x.data, "author.email")
-
+###### TODO HERE
```
bockthom, line 189: Last comment for tonight: Is this intended (e.g., to ask for further clarifications), or did you just forget about updating the showcase file?

JoJoDeveloping: It is still to do.

NEWS.md
bockthom: > Unfortunately the description here does not correspond to the implementation: In the implementation we only adjust all the "commit_added" events but not "all" events in general (even if these was the purpose, this is not what we have implemented). And also "after" is not correct–we just make sure that it is "not before", so same time is still possible. Please also add the information that the original dates are not completely removed but now stored as event.info.2.


This part of the NEWS.md still is imprecise and waits for an adjustment...

showcase.R
```diff
@@ -186,7 +186,9 @@ y = NetworkBuilder$new(project.data = y.data, network.conf = net.conf)
 # sample.cumulative = add.vertex.attribute.commit.count.author(my.networks, x.data, aggregation.level = "cumulative")
 # ## add email-address vertex attribute
 # sample.mail = add.vertex.attribute.author.email(my.networks, x.data, "author.email")
-
+# sample.mail.thread = add.vertex.attribute.mail.thread.count(my.networks, x.data)
+# sample.issues.created = add.vertex.attribute.issue.creation.count(my.networks, x.data)
+# sample.pull.requests = add.vertex.attribute.issue.count(my.networks, x.data, issue.type = "pull.requests")
```
bockthom, line 191: Maybe we should also add a custom name (e.g., `"pr.count"`) when calling `add.vertex.attribute.issue.count` with `issue.type = "pull.requests"`, as the default name is `"issue.count"`.

This way, I just spotted another issue: All the `add.vertex.attribute.issue.*.count` have a parameter `issue.type`. But in the function declaration, only the default value `"all"` is mentioned. Could you please add a vector `issue.type = c("all", "issues", "pull.requests")` to all these function declarations in `util-data-misc.R` ?

(With that, I finished reviewing your changes. However, I will have a final check at the recently added tests soon.)

tests/test-networks-covariates.R
bockthom: Thanks for updating the test. Unfortunately, the new test reveals the same issue as the previous test: Mail count and mail thread count still are identical, and so the test for the thread count does not really test for the thread count.
The reason for this is that Hans (which is the person who participates in the same thread multiple times and, thus, makes a difference in these tests) only participated in 2010, whereas the ranges cover the ranges between 2016-07-12 15:00:00 and 2016-08-31 18:00:00.

However, I have a suggestion for an easy solution of this problem: We need to have a range which contains Hans's e-mails, but only for these tests. Hence, I suggest to add a new global variable `mybins.starting.from.2010` and also `myranges.starting.from.2010` (see lines 41-42). In the new bins, we can have an additional range at the beginning to cover Hans's activity: `c("2010-07-12 12:00:00", "2016-07-12 15:00:00", "2016-07-12 16:00:00", "2016-07-12 16:05:00", "2016-08-31 18:00:00")` 

Then we could add an additional parameter `ranges` to `get.network.covariates.test.networks` which defaults to `myranges`. In the two tests where we need Hans's data, we then can pass `myranges.starting.from.2010` to the function. Then we have four ranges there, where the first one covers Hans's mailing mail activity to the same thread. This should fix the problem, if I didn't overlook anything.

NEWS.md
```diff
@@ -16,17 +16,19 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add tests for the `get.data.sources.from.relations` function (PR #195, add0c746dde8279da41d180deecf52b91a46095c)
 - Add logo directory containing several logo variants (PR #196, 82f99719b8a43b0a44914b67b26bf1a34bb076c6, dc4659ea354e97159f8ee6167811db544f0b7087, fdc5e677325225f92d1f99948cb9441bfe1d150d, 752a9b376ffeffd5d6b380b4fdba838a890e3ef7)
-- Add per-author vertex attributes regarding issue/pull request/mail count and derived quantities (like mail thread count, issue creation count) (PR #194, 9f9150a97ffbb64607df0ddcbce299e16c2580da, 7260d62cf6f1470584753f76970d19664638eeed, 139f70b67903909bcd4c57e26afa458681a869f2, eb4f649c04155e22195627072e0f08bb8fe65dc4, 627873c641410182ca8fee0e78b95d7bda1e8e6b, 1e1406f4a0898cac3e61a7bc9a5aa665dceef79f, 98e11abc651b5fe0ec994eddea078635b0d6f0b2, cc64b9b3e3f3e33798a6d138225f745bd2f550a0)
+- Add function `preprocess.issue.data`, which implements common issue data filtering operations.
+- Add function `get.issues.filtered.uncached`, which gets the issues filtered without poisoning or using the cache.
```
bockthom, line 20: Commit hashes are missing.

NEWS.md
```diff
@@ -16,17 +16,19 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
 - Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (PR #195, cf795f26652b00de5d717c703c688af55a972943, 70c05ecd1e3c0f10810acc2b2ae06a3eb8856317, 5a46ff4d428af7f301fe57d6e9e10421f553a9cc, fd767bb37ca608c28d9ff4a449415cc0e863d7ee)
-- Rename `get.issues` to `get.issues.filtered`, and write a new `get.issues` to get the unfiltered issues so that these methods follow the naming scheme known from the respective methods for commits (b9dd94c8575b8cab40d0d1185368854f84299d87)
+- Rename `get.issues` to `get.issues.filtered`, write a new `get.issues` to get the unfiltered issues so that these methods follow the naming scheme known from the respective methods for commits, and add `get.issues.filtered.uncached` (b9dd94c8575b8cab40d0d1185368854f84299d87)
```
bockthom, line 28: `get.issues.filtered.uncached` is not added in this commit. However, as we mention this function in the Added section, it is not necessary to mention it here.

bockthom:
I've done with my final review! As soon as the stated issues regarding test, news, and showcase/util-networks-metrics are fixed, this PR can be merged.

bockthom:
Thanks @JoJoDeveloping for your endurance. This PR contains a lot of substantial and valuable changes and will definitively improve coronet in various directions.
All the implementation-related issues (at least, those which we have spotted) are fixed now, looks really good :smile: :+1: 

The only point of criticism left is that the last commit does not contain a signed-off tag :wink:

bockthom (approved):

hechtlC (approved):

Issue #181, opened 2020-07-17
Minor issues and enhancements in the metrics module


bockthom:
### Description

In the metrics module `util-metrics.R`, there are some minor inconsistencies which should be fixed:

**(1) `metrics.avg.pathlength`:**
This function takes three parameters: `network`, `directed`, `unconnected`. However, there are no default values for `directed` and `unconnected`, which lets the `igraph::average.path.length` function crash if these parameters are not passed to the function.
Suggestion: Add default values for those parameters.

In addition, the description of the `unconnected` parameter seems to be wrong as it states the opposite:
```R
#' @param unconnected whether all vertices of the network are connected
```
It is also not clear what happens when this parameter is `TRUE` or `FALSE`. Maybe we should add more information from igraph's description of this parameter.

**(2) Default values:**
Most of the `metrics.*` functions have parameters which take default values. However, in many cases, the default values are missing in the parameter documentation, which violates our [contribution guidelines](https://github.com/se-passau/coronet/blob/master/CONTRIBUTING.md#r-coding-conventions).

**(3) `is.smallworld`:**
We do have a function `metrics.scale.freeness` returning several numerical values and its boolean version `metrics.is.scale.free`.
However, for small worldness, we only have the function `metrics.smallworldness`, which returns a numerical value, but no boolean version. There is just a comment in the documentation saying `is.smallworld = s.delta > 1`. Why not moving that into a boolean function `metrics.is.smallworld` like we have it for scale freeness?

**(4) `metrics.smallworldness`:**
This function crashes with an error 
```R
Error in .Call("R_igraph_erdos_renyi_game", as.numeric(n), as.numeric(type1), : At games.c:681 : Invalid number (too large) of edges, Invalid value
```
or something similar if you pass a non-simplified network to this function. It is stated in the documentation that the network needs to be simplified, but it would be a better option to also add a check to this function and stop with a reasonable error message if the network is not simplified, just to make debugging easier...

bockthom:
In addition to the above mentioned problems, there are additional problems regarding some corner cases, which should be fixed also:

 **(5) `metrics.hub.degree`:**
When passing an empty network (i.e., a network without vertices) to this function, the function leads to an error. However, instead, we handle this case within this function and return an appropriate value instead (e.g., print a warning and return `NA`).

**(6) Integrate centrality functions into `util-metrics`:**
[...]

PR #195, opened 2021-02-09
Fixes in the metrics module

nlschn:
### Prerequisites

### Description
Fix some bugs discussed in #181 in `util-network-metrics`:  
- [x] (1) Clarify `metrics.avg.pathlength` (f929248182594613bd203e100268e3e3dce87f34)
- [x] (2) Add default values to several methods (de9988cc171cafdd084701d5a2693a74176a802a)
- [x] (3) Clarify `is.smallworld` (ce1f8124298d73830f3fd96aa84e82ba89181aa6)
- [x] (4) Handle error in `metrics.smallworldness` (4b164bebea1e8258cb93febf51271a4b6f486779)
- [x] (5) Handle empty networks in `metrics.hub.degree` (ec984f46b6f2fc1613e5861e2f33c0fb99ec5c08)
- [x] (6) Integrate centrality functions into `util-network-metrics`
  - [x] (6.1) Add centrality function to simplify getting a centrality data frame (d3cd528609480f87658601ef13326e950a74cce7)
  - [x] (6.2) Automate finding the suitable data source for restricting authors (d1e4413a49ab83a115a7f26719592876371ab264)
  - [x] (6.3) Test new `get.data.sources.from.relations` function (add0c746dde8279da41d180deecf52b91a46095c)
### Changelog

- Add more explanation to `metrics.avg.path.length` and as well as default values for that function.
- Add default parameters to documentation of several other methods missing them.
- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness).
- Handle error in igraph function by printing user warning instead. When calling  `metrics.smallworldness` with a network that is too large, i.e. that has too many edges a warning is being printed and `NA` is returned.
- Add check for empty network in `metrics.hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned.
- Add 'metrics.centrality' function to metrics module in order to simplify getting a data frame containing author names
and their respective centrality values.
- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,9 +5,19 @@
 ### Added
 - Add functionality to read and process commit messages in order to merge them to the commit data (see issue #180). Three values are available for the new attribute `commit.messages` in `ProjectConf`: `none`, `title` and `messages` (PR #193, 85b1d0572c0fb9f4c062bceb1363b0398f98b85f, fdc414ade1a640f533e809a25cfe012e42b3cffa, 43e1894998e18faff3a65114fa65ee54e1d2f66e)
 - Add functions `cleanup.commit.message.data` and `cleanup.synchronicity.data` to remove commit hashes that are not any more present in the commit data from the commit message data or synchronicity data (PR #193, 98e83b037ecc88d9a29e8e4ca93598a9978e85a2)
+- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, 597843db3dfd18e49deb28dde9ee99057003087d)
+- Add 'metrics.centrality' function to metrics module in order to simplify getting a data frame containing author names
+and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8)
+- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, 450a74d56a065f8e4a69ae28edb0b49ffd112652)
+- Add tests for the `get.data.sources.from.relations` function (PR #195, afbba88f85e7777828534611445e8463be4d55dc)
 
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
+- Add more explanation to `metrics.avg.path.length` and as well as default values for that function (PR #195, 6bf6e4056692e5265a065c91da779355b3fbf66a)
+- Add default parameters to documentation of several other methods missing them (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae)
+- Handle error in igraph function by printing user warning instead. When calling  `metrics.smallworldness` with a network that is too large, i.e. that has too many edges a warning is being printed and `NA` is returned (PR #195, 8489c055557aba9a21bec0c7704fc2e1176a3a2c)
+- Add check for empty network in `metrics.hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
+
 
 ## 3.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,8 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
 - Update documentation in `util-network-metrics.R` (PR #195, f929248, f929248182594613bd203e100268e3e3dce87f34, de9988cc171cafdd084701d5a2693a74176a802a)
 - Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
+- Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted.
+
 
 ## 3.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,17 +5,16 @@
 ### Added
 - Add functionality to read and process commit messages in order to merge them to the commit data (see issue #180). Three values are available for the new attribute `commit.messages` in `ProjectConf`: `none`, `title` and `messages` (PR #193, 85b1d0572c0fb9f4c062bceb1363b0398f98b85f, fdc414ade1a640f533e809a25cfe012e42b3cffa, 43e1894998e18faff3a65114fa65ee54e1d2f66e)
 - Add functions `cleanup.commit.message.data` and `cleanup.synchronicity.data` to remove commit hashes that are not any more present in the commit data from the commit message data or synchronicity data (PR #193, 98e83b037ecc88d9a29e8e4ca93598a9978e85a2)
-- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, 597843db3dfd18e49deb28dde9ee99057003087d)
+- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, ce1f8124298d73830f3fd96aa84e82ba89181aa6)
 - Add `metrics.vertex.centralities` function to metrics module in order to simplify getting a data frame containing author names
-and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8)
-- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, 450a74d56a065f8e4a69ae28edb0b49ffd112652)
-- Add tests for the `get.data.sources.from.relations` function (PR #195, afbba88f85e7777828534611445e8463be4d55dc)
+and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7)
+- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, d1e4413a49ab83a115a7f26719592876371ab264)
+- Add tests for the `get.data.sources.from.relations` function (PR #195, add0c746dde8279da41d180deecf52b91a46095c)
 
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
-- Update documentation in `util-network-metrics.R` (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae, 6bf6e4056692e5265a065c91da779355b3fbf66a)
-- Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
-
+- Update documentation in `util-network-metrics.R` (PR #195, f929248, f929248182594613bd203e100268e3e3dce87f34, de9988cc171cafdd084701d5a2693a74176a802a)
+- Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
 
 ## 3.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 - Add functionality to read and process commit messages in order to merge them to the commit data (see issue #180). Three values are available for the new attribute `commit.messages` in `ProjectConf`: `none`, `title` and `messages` (PR #193, 85b1d0572c0fb9f4c062bceb1363b0398f98b85f, fdc414ade1a640f533e809a25cfe012e42b3cffa, 43e1894998e18faff3a65114fa65ee54e1d2f66e)
 - Add functions `cleanup.commit.message.data` and `cleanup.synchronicity.data` to remove commit hashes that are not any more present in the commit data from the commit message data or synchronicity data (PR #193, 98e83b037ecc88d9a29e8e4ca93598a9978e85a2)
 - Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, 597843db3dfd18e49deb28dde9ee99057003087d)
-- Add 'metrics.centrality' function to metrics module in order to simplify getting a data frame containing author names
+- Add `metrics.vertex.centralities` function to metrics module in order to simplify getting a data frame containing author names
 and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8)
 - Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, 450a74d56a065f8e4a69ae28edb0b49ffd112652)
 - Add tests for the `get.data.sources.from.relations` function (PR #195, afbba88f85e7777828534611445e8463be4d55dc)
@@ -14,7 +14,7 @@ and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
 - Update documentation in `util-network-metrics.R` (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae, 6bf6e4056692e5265a065c91da779355b3fbf66a)
-- Add check for empty network in `metrics-hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
+- Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
 
 
 ## 3.7
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,10 +13,8 @@ and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8
 
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
-- Add more explanation to `metrics.avg.path.length` and as well as default values for that function (PR #195, 6bf6e4056692e5265a065c91da779355b3fbf66a)
-- Add default parameters to documentation of several other methods missing them (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae)
-- Handle error in igraph function by printing user warning instead. When calling  `metrics.smallworldness` with a network that is too large, i.e. that has too many edges a warning is being printed and `NA` is returned (PR #195, 8489c055557aba9a21bec0c7704fc2e1176a3a2c)
-- Add check for empty network in `metrics.hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
+- Update documentation in `util-network-metrics.R` (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae, 6bf6e4056692e5265a065c91da779355b3fbf66a)
+- Add check for empty network in `metrics-hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
 
 
 ## 3.7
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -12,6 +12,7 @@
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
 ## Copyright 2018-2019 by Claus Hunsen <hunsen@fim.uni-passau.de>
+## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -695,4 +696,33 @@ test_that("Addition of edge attributes with data", {
         info = "multi network – issue/cochange"
     )
 
+})
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Extract data sources ----------------------------------------------------
+test_that("Get the data sources from a network with two relations", {
+    expected.relations = c("mails", "commits")
+    network = get.sample.network()
+
+    expect_identical(expected.relations, get.data.sources.from.relations(network), info = "data sources: mails, commits")
+})
+
+test_that("Get the data sources from a network with one relation", {
+    expected.relations = c("mails")
+
+    ## configurations
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("commits.filter.base.artifact", FALSE)
+    ## construct data object
+    proj.data = ProjectData$new(project.conf = proj.conf)
+
+    ## construct network builder
+    net.conf = NetworkConf$new()
+    network.builder = NetworkBuilder$new(project.data = proj.data, network.conf = net.conf)
+    network.builder$update.network.conf(updated.values = list(author.relation = "mail"))
+
+    ## build network
+    network = network.builder$get.author.network()
+
+    expect_identical(expected.relations, get.data.sources.from.relations(network), info = "data sources: mails")
 })
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -701,14 +701,14 @@ test_that("Addition of edge attributes with data", {
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Extract data sources ----------------------------------------------------
 test_that("Get the data sources from a network with two relations", {
-    expected.relations = c("mails", "commits")
+    expected.data.sources = c("mails", "commits")
     network = get.sample.network()
 
-    expect_identical(expected.relations, get.data.sources.from.relations(network), info = "data sources: mails, commits")
+    expect_identical(expected.data.sources, get.data.sources.from.relations(network), info = "data sources: mails, commits")
 })
 
 test_that("Get the data sources from a network with one relation", {
-    expected.relations = c("mails")
+    expected.data.sources = c("mails")
 
     ## configurations
     proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
@@ -724,5 +724,5 @@ test_that("Get the data sources from a network with one relation", {
     ## build network
     network = network.builder$get.author.network()
 
-    expect_identical(expected.relations, get.data.sources.from.relations(network), info = "data sources: mails")
+    expect_identical(expected.data.sources, get.data.sources.from.relations(network), info = "data sources: mails")
 })
--- a/util-data.R
+++ b/util-data.R
@@ -1102,11 +1102,11 @@ ProjectData = R6::R6Class("ProjectData",
 
         #' Get the list of artifacts from the given \code{data.source} of the project.
         #'
-        #' @param data.source The specified data source. One of \code{"commits"},
-        #'                    \code{"mails"}, and \code{"issues"} or multiple of them. [default: "commits"]
+        #' @param data.sources The specified data source. One of \code{"commits"},
+        #'                     \code{"mails"}, and \code{"issues"} or multiple of them. [default: "commits"]
         #'
         #' @return the character vector of unique artifacts (can be empty)
-        get.artifacts = function(data.source = c("commits", "mails", "issues")) {
+        get.artifacts = function(data.sources = c("commits", "mails", "issues")) {
             logging::loginfo("Getting artifact data.")
 
             ## get the right value when nothing is passed for 'data.source'
@@ -1115,24 +1115,24 @@ ProjectData = R6::R6Class("ProjectData",
             ## behavior: if 'data.source' is missing, 'several.ok' will be 'FALSE' making the function return
             ## the first element of the choice vector. If an argument is present, 'several.ok' will be 'TRUE' allowing
             ## multiple values for the argument
-            data.source = match.arg(data.source, several.ok = !missing(data.source))
+            data.sources = match.arg(data.sources, several.ok = !missing(data.sources))
 
-            data.source.func = sapply(data.source, function(d) {
+            data.source.funcs = sapply(data.sources, function(d) {
                 return(DATASOURCE.TO.ARTIFACT.FUNCTION[[d]])
             }, USE.NAMES = FALSE)
 
-            data.source.col = sapply(data.source, function(d) {
+            data.source.cols = sapply(data.sources, function(d) {
                 return(DATASOURCE.TO.ARTIFACT.COLUMN[[d]])
             }, USE.NAMES = FALSE)
 
             ## get actual artifact data
-            data = lapply(data.source.func, function(d) {
+            data = lapply(data.source.funcs, function(d) {
                 return(self[[d]]())
             })
 
             artifacts = c()
             for (df in data) {
-                unique.artifacts = unique(df[[data.source.col]])
+                unique.artifacts = unique(df[[data.source.cols]])
                 artifacts = append(artifacts, unique.artifacts)
             }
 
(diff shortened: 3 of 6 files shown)


util-networks-metrics.R
```diff
@@ -85,11 +86,14 @@ metrics.density = function(network) {
 #' Calculate the average path length for the given network.
 #'
 #' @param network the network to be examined
-#' @param directed whether to consider directed paths in directed networks
-#' @param unconnected whether all vertices of the network are connected
+#' @param directed whether to consider directed paths in directed networks [default: TRUE]
+#' @param unconnected whether there are vertices in the network that are not connected.
+#'                    If TRUE only the lengths of the existing paths are considered and averaged;
```
bockthom, line 91: Please follow our [coding conventions](https://github.com/se-sic/coronet/blob/master/CONTRIBUTING.md#r-coding-conventions):
> Variable names, class names, or other code snippets which are part of the function documentation must be placed within \code{...}

So, please use `\code{TRUE}` and `\code{FALSE}` within `roxygen2` function documentation. (Note that this doesn't hold for the default value [default: ...] as the default value is assumed to always be code (this also part of our conventions).

util-networks-metrics.R
```diff
@@ -32,7 +32,7 @@ requireNamespace("igraph")
 #' Determine the maximum degree for the given network.
 #'
 #' @param network the network to be examined
-#' @param mode the mode to be used for determining the degrees
+#' @param mode the mode to be used for determining the degrees [default: c("total", "in", "out")]
```
bockthom, line 35: This is actually wrong, unfortunately. It is a bit tricky here: Usually, you would be right and the default value is the complete vector assigned to the parameter in the default setting. However, there is one exception: If there is a `match.arg(...)` call at the beginning of the function's implementation, then this does not hold. `match.arg` makes sure that the given parameter is **one** of the allowed ones specified in the default setting. Otherwise, an error is thrown. If no parameter is explicitly passed to the function, only the **first** element of the vector is taken. Only if `match.arg(..., several.ok = TRUE)` is used, then multiple values are allowed and the complete vector is used as default.
However, all this only holds if the parameter variable is reassigned when calling `match.arg`, as it is the case in line 38: `mode = match.arg(mode)`. Without the reassignment, the complete vector would still be the default value. I hope I didn't confuse you with all the details, that's a bit tricky :mag: 
<sub>(To be honest, the functionality of `match.arg` even was not sufficient for our needs in some cases, as sometimes we would like to have another default value than `match.arg` provides us with the first element. Therefore, we also have written our own extension to `match.arg`, namely `match.arg.or.default` which behaves a little bit different (see `util-misc.R`), but to the best of my knowledge, our extension is not used in the metrics module, so you don't need to check for that one here.)</sub>

bockthom (commented):
Thanks for working on these little fixes @nlschn. I know that this PR is just a draft PR and not yet ready for review. However, I just spotted two issues which I would already highlight while you are working on the PR, as you could potentially take care of that in similar situations in the remaining part of this PR.

(In addition, I would like to get rid of the empty merge commit. But this is not yet relevant as long as this PR is just a draft :wink: )

I will do a full review once the PR is ready for review. (Sorry for sliding in already.)

nlschn:
The rebase in order to erase the empty merge commit created by GitHub lead to the commits getting new hashes and being dated to today. However the old commits still seem to appear in this conversation, whereas the history is correct in the commits tab above.
I now removed the "draft" tag so the PR can be reviewed. I hope, that I don't have forgotten anything 😇

util-networks.R
```diff
@@ -1567,6 +1567,35 @@ delete.authors.without.specific.edges = function(network, specific.edge.types =
...
 
 
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Helper functions---------------------------------------------------------
+
+#' Calculate the data sources of a network based on its edge relations
+#'
+#' @param network the network with the relations to be extracted
+#' @return a vector with all data.sources from the network; an element is set to \code{NA} if the network contains an
+#' empty relation, i.e. \code{character(0)}
```
bockthom, line 1577: The first sentence (just describing the function in general) should end with a period.
Between parameter description and return-value description, there should be an empty line.

In addition, please don't forget to add your copyright header to this file :wink:

util-networks.R
```diff
@@ -1567,6 +1567,35 @@ delete.authors.without.specific.edges = function(network, specific.edge.types =
...
+    ## get all relations in the network
+    data.sources = unique(igraph::E(network)$relation)
+
+    ## map them to data sources respectively using the defined translation constant
+    data.sources = sapply(data.sources, function(relation) {
+            ## check for a \code{character(0)} relation and abort if there is one
+            if (length(relation) == 0) {
+                logging::logwarn("There seems to be an empty relation in the network. Cannot proceed.")
+                return (NA)
+            }
```
bockthom, line 1588: Did you really encounter such a case where the relation actually is `character(0)`? If so, I would be happy if you could tell me where/when this actually happens :smile:

util-networks.R
```diff
@@ -1567,6 +1567,35 @@ delete.authors.without.specific.edges = function(network, specific.edge.types =
...
+
+            ## use the translation constant to get the appropriate data source
+            return(RELATION.TO.DATASOURCE[[relation]])
+        ## avoid element names as we only want the data source's name
+        }, USE.NAMES = FALSE)
```
bockthom, line 1593: The indentation looks weird. Let's make the indentation of the line containing `}` the same as for the line containing the corresponding `}`

NEWS.md
```diff
@@ -5,9 +5,19 @@
 ### Added
 - Add functionality to read and process commit messages in order to merge them to the commit data (see issue #180). Three values are available for the new attribute `commit.messages` in `ProjectConf`: `none`, `title` and `messages` (PR #193, 85b1d0572c0fb9f4c062bceb1363b0398f98b85f, fdc414ade1a640f533e809a25cfe012e42b3cffa, 43e1894998e18faff3a65114fa65ee54e1d2f66e)
 - Add functions `cleanup.commit.message.data` and `cleanup.synchronicity.data` to remove commit hashes that are not any more present in the commit data from the commit message data or synchronicity data (PR #193, 98e83b037ecc88d9a29e8e4ca93598a9978e85a2)
+- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, 597843db3dfd18e49deb28dde9ee99057003087d)
+- Add 'metrics.centrality' function to metrics module in order to simplify getting a data frame containing author names
```
bockthom, line 9: Wrong quotation marks use here :wink: However, the name of this function may change anyway.

NEWS.md
```diff
@@ -5,9 +5,19 @@
...
+and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8)
+- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, 450a74d56a065f8e4a69ae28edb0b49ffd112652)
+- Add tests for the `get.data.sources.from.relations` function (PR #195, afbba88f85e7777828534611445e8463be4d55dc)
```
bockthom, line 12: It's not worth to mention that (as tests should usually come with newly implemented functionality). It's only worth to mention if there are a bunch of new tests for already existing functionality.
However, it doesn't hurt to mention your tests. So you can keep it here :wink:

tests/test-networks.R
```diff
@@ -1,18 +1,18 @@
...
+# ## but WITHOUT ANY WARRANTY; without even the implied warranty of
+# ## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# ## GNU General Public License for more details.
+# ##
+# ## You should have received a copy of the GNU General Public License along
+# ## with this program; if not, write to the Free Software Foundation, Inc.,
+# ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+# ##
+# ## Copyright 2018-2019 by Claus Hunsen <hunsen@fim.uni-passau.de>
+# ## All Rights Reserved.
```
bockthom, line 15: Was this change intended?

Also for the rest of this test file: You have removed almost all tests of this file. Please undo that :exclamation: 
It is quite hard to figure out where the tests are which you have newly added to this file, as the git diff is bloated. As a consequence, I will review your tests as soon you have fixed the changes regarding the already existing tests.

NEWS.md
```diff
@@ -5,9 +5,19 @@
...
 
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
+- Add more explanation to `metrics.avg.path.length` and as well as default values for that function (PR #195, 6bf6e4056692e5265a065c91da779355b3fbf66a)
+- Add default parameters to documentation of several other methods missing them (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae)
```
bockthom, line 17: I would suggest to combine those two bullet points. In the news file, we don't need to be that detailed, but just summarize the changes on a higher level. So just mentioning that the documentation in the metrics module `util-networks-metrics.R` is updated should be enough.

NEWS.md
```diff
@@ -5,9 +5,19 @@
...
+- Handle error in igraph function by printing user warning instead. When calling  `metrics.smallworldness` with a network that is too large, i.e. that has too many edges a warning is being printed and `NA` is returned (PR #195, 8489c055557aba9a21bec0c7704fc2e1176a3a2c)
```
bockthom, line 18: This is not worth to mention, as it belongs to the "documentation update" above.

NEWS.md
```diff
@@ -5,9 +5,19 @@
...
+- Add check for empty network in `metrics.hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
```
bockthom, line 19: This is a change in the behavior which needs to be mentioned here. However, the function is called `metrics.hub.degree`, isn't it?

bockthom (changes requested):
Thank you very much for working on this PR @nlschn, good work!

One comment on the commits mentioned in this PR: Outdated commits are appearing in this PR as you have referenced this PR in your commit message. (Btw, mentioning the PR there is not necessary, but you can do that if you are happy to do so). For reviewing, we should stick to the commits tab anyway, and on this tab only the current commits are displayed. So, no worries.

I just found a few documentation-related issues which are worth to be fixed, mostly small stuff.
However, there is one issue regarding error handling in the smallworldness function which I would like you to adjust for better comprehensibility.

In addition, there is a bigger issue with the `metrics.centrality` function which I have not yet thought about (my fault)––see the corresponding individual comment. Let's discuss this one in our weekly meeting, maybe we find with a pragmatic solution that does not cost that much effort :crossed_fingers: 

Tests will be reviewed later (see also the corresponding individual comment).

tests/test-networks.R
nlschn: Ohhh, sorry, that was absolutely not on purpose.

util-networks.R
nlschn: No, I didn't. When I tried to create a network with an empty relation by hand it failed when trying to build. 
But I recall from the meeting that I was supposed to implement a check nevertheless, just in case.

tests/test-networks.R
nlschn: So it should be repaired now. I am again very sorry, that should not happen.

util-networks-metrics.R
```diff
@@ -16,13 +16,15 @@
...
 ## Copyright 2018 by Barbara Eckl <ecklbarb@fim.uni-passau.de>
+## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Libraries ---------------------------------------------------------------
 
 requireNamespace("igraph")
+#requireNamespace("util-core-peripheral")
```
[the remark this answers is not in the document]
nlschn: This is not meant to be there (anymore). I thought I'd have to include the namespace manually, because a function was not found. Turned out it was a spelling error. So I commented the line out. As I don't need it, I will remove it.

tests/test-networks.R
bockthom: No worries, something of the like has happened to everyone of us. Even if that should not happen, sometimes it happens :wink:

bockthom: The tests look good to me, I have no further comments on them. :smiley:

util-networks-metrics.R
```diff
@@ -250,3 +292,67 @@ metrics.hierarchy = function(network) {
...
+                              restrict.classification.to.authors = NULL) {
+    type = match.arg(type)
+
+    ## check whether the restrict parameter is set to default (\code{NULL})
+    if (is.null(restrict.classification.to.authors)) {
+        ## now check whether both data and network are present
+        if (!is.null(network) && !is.null(proj.data)) {
+            ## in this case calculate the restrict parameter based on the edge relation along with the authors from
+            ## these data.sources
+            restrict.classification.to.authors = get.authors.by.data.source(get.data.sources.from.relations(network),
```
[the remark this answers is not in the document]
nlschn: From what I can see using the busybox and openssl examples, the only relation returned by `unique(igraph::E(bipartite.network))$relation` seems to be the artifact relation; the author relation seems to be ignored.

util-networks-metrics.R
```diff
@@ -285,64 +285,67 @@ metrics.hierarchy = function(network) {
...
 
 
-## The column headers for a centrality data frame calculated by the function \code{metrics.centrality}
-CENTRALITY_COLUMN_NAMES = c("name", "centrality")
-
-#' Calculate the centrality value for authors from a network and project data.
-#' Only considers authors from the network that are also present in the project data and the
-#' \code{restrict.classification.to.authors} vector.
+## The column headers for a centrality data frame calculated by the function \code{metrics.vertex.centralities}
+VERTEX_CENTRALITIES_COLUMN_NAMES = c("vertex.name", "centrality")
```
bockthom, line 289: Missed that already in my previous review: Use "." in variable names instead of "_". This also holds for constants.

util-networks-metrics.R
```diff
@@ -285,64 +285,67 @@ metrics.hierarchy = function(network) {
...
+
+#' Calculate the centrality value for vertices from a network and project data.
+#' Only considers vertices from the network that are also present in the project data and the
+#' \code{restrict.classification.to.vertices} vector.
```
bockthom, line 293: We don't need this sentence any more as we have added a similar one directly below :wink:

NEWS.md
```diff
@@ -13,10 +13,8 @@ and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8
...
-- Add more explanation to `metrics.avg.path.length` and as well as default values for that function (PR #195, 6bf6e4056692e5265a065c91da779355b3fbf66a)
-- Add default parameters to documentation of several other methods missing them (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae)
-- Handle error in igraph function by printing user warning instead. When calling  `metrics.smallworldness` with a network that is too large, i.e. that has too many edges a warning is being printed and `NA` is returned (PR #195, 8489c055557aba9a21bec0c7704fc2e1176a3a2c)
-- Add check for empty network in `metrics.hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
+- Update documentation in `util-network-metrics.R` (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae, 6bf6e4056692e5265a065c91da779355b3fbf66a)
+- Add check for empty network in `metrics-hub-degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
```
bockthom, line 17: Interesting change in the function-name spelling. Unfortunately, you have changed the wrong separator, as the function is called `metrics.hub.degree`

bockthom (commented):
Thanks for your fast adjustments @nlschn, most of the identified problems are fixed now. I just spotted a few additional or newly introduced minor issues.

In addition, unfortunately, many of your commit messages lack the signed-off-tag.

NEWS.md
nlschn: I hope I don't need a third time to finally get it right...

NEWS.md
```diff
@@ -5,9 +5,17 @@
...
+- Add function `metrics.is.smallworld` to the metrics module in order to unify checks for smallworldness (similar to scalefreeness) (PR #195, 597843db3dfd18e49deb28dde9ee99057003087d)
+- Add `metrics.vertex.centralities` function to metrics module in order to simplify getting a data frame containing author names
+and their respective centrality values (3f41aa7af1c35947035ce7cf0e6c6edcec7a1cf8)
+- Add `get.data.sources.from.relations` to `util-networks.R` which extracts the data sources of a network that were used when building it (PR #195, 450a74d56a065f8e4a69ae28edb0b49ffd112652)
+- Add tests for the `get.data.sources.from.relations` function (PR #195, afbba88f85e7777828534611445e8463be4d55dc)
 
 ### Changed/Improved
 - Add `.drone.yml` to enable running our CI pipelines on drone.io (PR #191, 1c5804b59c582cf34af6970b435add51452fbd11)
+- Update documentation in `util-network-metrics.R` (PR #195, 2085024661fcdc287a009cf6aa7f438cad7459ae, 6bf6e4056692e5265a065c91da779355b3fbf66a)
+- Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 231affc5d1c201b0cc3ea093477afde6acd6b9a5)
```
bockthom, line 17: Almost done :wink: Thanks for adding the missing signed-off tags.
As a follow-up task, as you have rebased the commits, several commit hashes now wait for being updated :wink:

nlschn: ~third~ second time's the charm 🌟

tests/test-networks.R
```diff
@@ -696,3 +697,32 @@ test_that("Addition of edge attributes with data", {
...
+## Extract data sources ----------------------------------------------------
+test_that("Get the data sources from a network with two relations", {
+    expected.relations = c("mails", "commits")
+    network = get.sample.network()
+
+    expect_identical(expected.relations, get.data.sources.from.relations(network), info = "data sources: mails, commits")
+})
+
+test_that("Get the data sources from a network with one relation", {
+    expected.relations = c("mails")
```
hechtlC, line 711: Maybe rename this to `expected.data.sources`, as you compare it to the data.sources you extract from the relations. Just to be consistent.

util-networks-metrics.R
```diff
@@ -85,11 +95,14 @@ metrics.density = function(network) {
 #' Calculate the average path length for the given network.
 #'
 #' @param network the network to be examined
-#' @param directed whether to consider directed paths in directed networks
-#' @param unconnected whether all vertices of the network are connected
+#' @param directed whether to consider directed paths in directed networks [default: TRUE]
+#' @param unconnected whether there are vertices in the network that are not connected.
```
hechtlC, line 99: This might be wrong but aren't we using 'isolated' for unconnected vertices in coronet @bockthom ? If I am right, then we should keep naming consistent.

util-networks-metrics.R
```diff
@@ -133,41 +146,63 @@ metrics.modularity = function(network, community.detection.algorithm = igraph::c
...
+#'
+#' @param network the network to be examined
+#'
+#' @return \code{TRUE}, if the network is smallworld,
+#'         \code{FALSE}, if it is not,
+#'         \code{NA}, if an error occured.
+metrics.is.smallworld = function(network) {
+    s.delta = metrics.smallworldness(network)
+
+    ## else return whether the network is smallworld
```
hechtlC, line 201: Why else? I think you can remove the else here.

util-networks-metrics.R
```diff
@@ -250,3 +285,68 @@ metrics.hierarchy = function(network) {
...
+## The column headers for a centrality data frame calculated by the function \code{metrics.vertex.centralities}
+VERTEX.CENTRALITIES.COLUMN.NAMES = c("vertex.name", "centrality")
+
+#' Calculate the centrality value for vertices from a network and project data.
+#' If a \code{ProjectData} is supplied, only vertices from the network that are also present in the project data are
+#' considered. Otherwise, if no custom vector \code{restrict.classification.to.vertices} is supplied, all vertices of
+#' the network are considered.
+#'
+#' @param network the network containing the vertices to classify
+#' @param proj.data the \code{ProjectData} containing the vertices to classify
```
hechtlC, line 297: I am not completely sold on that wording. The `ProjectData` object does not hold the vertices but the data for them. I am not sure if I am too sensitive here or not. What do you think @bockthom.

util-networks-metrics.R
```diff
@@ -250,3 +285,68 @@ metrics.hierarchy = function(network) {
...
+                              restrict.classification.to.vertices = NULL) {
+    type = match.arg(type)
+
+    ## check whether the restrict parameter is set to default 'NULL'
+    if (is.null(restrict.classification.to.vertices)) {
+        ## now check whether both data and network are present
+        if (!is.null(network) && !is.null(proj.data)) {
+            ## in this case calculate the restrict parameter based on the edge relation along with the vertices from
+            ## these data.sources
+            restrict.classification.to.vertices = get.authors.by.data.source(get.data.sources.from.relations(network),
```
hechtlC, line 329: Shouldn't we rename the function `get.authors.by.data.source` as the plan was to also support artifact networks? Or am I missing something? Maybe this is subject to further discussion in our next meeting.

util-networks-metrics.R
```diff
@@ -250,3 +285,68 @@ metrics.hierarchy = function(network) {
...
+                                                                             proj.Data)
+        }
+        ## else leave the parameter at 'NULL' which still serves as a default value for the
+        ## 'get.auther.class.by.type' function
```
hechtlC, line 333: 'auther' should be 'author'. Also in line 337.

hechtlC (changes requested):
First of all great work @nlschn. I have identified a couple of problems and potential problems, which should be fixed/addressed before merging. Let's talk about that in our next meeting.

util-networks-metrics.R
bockthom: No, you are wrong here @hechtlC. `unconnected` really means `unconnected` :wink: 
Let me just cite Wikipedia on the quick: ["A graph is said to be connected if every pair of vertices in the graph is connected. This means that there is a path between every pair of vertices."](https://en.wikipedia.org/wiki/Connectivity_(graph_theory))
Hence, a network is unconnected if there are, for example, two separate components (you may call them clusters, not isolates) of the network that are not connected by an edge.
And there is another reason why this parameter is called `unconnected`: igraph also calls it `unconnected` :smile:

util-networks-metrics.R
bockthom: Good point, missed that. This is a relic of the former implementation where there was an unnecessary check at the beginning.

util-networks-metrics.R
bockthom: Hm. On the one hand I agree that the data do not hold vertices. But on the other hand, explaining that in more detail would be very complex and even more confusing, I guess. Instead of vertices, we would need to differentiate between authors, artifacts, and what else could be passed to this function in the future. As there is no common superordinate, I would just stick to vertices, even if it is not 100 percent precise.

util-networks-metrics.R
bockthom: This is the only review comment from my review which is [not resolved yet ](https://github.com/se-sic/coronet/pull/195#discussion_r584730047):wink: We already have discussed that last week in our meeting and @nlschn is currently working on this.
Indeed, you are missing something here @hechtlC: We cannot rename the function, but need to call different functions dependent on whether the vertex types in the network are "author", "artifact", or both. There is already a function `get.artifacts` which gets the data.sources, but this one is not ready for handling mutliple data sources, so @nlschn is also working on that right now.

util-data.R
```diff
@@ -1103,20 +1103,32 @@ ProjectData = R6::R6Class("ProjectData",
...
         #'
         #' @return the character vector of unique artifacts (can be empty)
         get.artifacts = function(data.source = c("commits", "mails", "issues")) {
             logging::loginfo("Getting artifact data.")
 
-            ## check given data source
-            data.source = match.arg.or.default(data.source, several.ok = FALSE)
-            data.source.func = DATASOURCE.TO.ARTIFACT.FUNCTION[[data.source]]
-            data.source.col = DATASOURCE.TO.ARTIFACT.COLUMN[[data.source]]
+            data.source = match.arg.or.default(data.source, several.ok = TRUE)
```
bockthom, line 1112: Please check if the default value has changed. `several.ok = TRUE` might take the vector of all possible values as default, but I am not sure. I am also not sure if there are any usages within coronet which rely on the default value. Please also check that.

util-data.R
```diff
@@ -1103,20 +1103,32 @@ ProjectData = R6::R6Class("ProjectData",
...
+            data.source.func = sapply(data.source, function (d) {
+                return(DATASOURCE.TO.ARTIFACT.FUNCTION[[d]])
+            }, USE.NAMES = FALSE)
+
+            data.source.col = sapply(data.source, function (d) {
```
bockthom, line 1117: Please remove space between function name and its parameters: `function(d)` (two occurrences)

util-networks-metrics.R
hechtlC: I am ok with that.

util-networks-metrics.R
```diff
@@ -326,8 +326,21 @@ metrics.vertex.centralities = function(network,
...
+                ## in this case, use the 'get.authors.by.data.source' function to get the author list
+                restrict.classification.to.vertices = get.authors.by.data.source(
+                    get.data.sources.from.relations(network),
+                    proj.Data)
+            }
+            else if ("artifact" %in% vertex.types) {
+                ## in this case, always use artifact relation, as both unipartite edges connecting to artifact vertices
+                ## as well as bipartite have an artifact relation
+                restrict.classification.to.vertices = get.artifacts(get.data.sources.from.relations(network))
+            }
```
bockthom, line 343: The implementation looks good to me, but one case is missing:
If there are both artifacts and authors in the network we need to combine the lists from `get.authors.by.data.source` and `get.artifacts` into one vector containing all of them.
Only if there only artifacts, then we can simply collect only artifacts :smile:

util-data.R
nlschn: this function uses the R function `match.arg` internally, when `several.ok` is set to `TRUE`.  This however returns only the first element of the choice vector when provided with the argument `NULL`. So everything should be fine.

util-networks-metrics.R
hechtlC: @bockthom and I have just talked about this. My confusion came from the wording 'unconnected vertices' but we are here talking about unconnected subnetworks. A single unconnected vertex would just be an isolate. So please change the documentation accordingly.

util-data.R
bockthom: I am not yet convinced. Default values behave different when `several.ok = TRUE`. Just have a look at the following snippet:

```R
> x = function(a = c("1", "2", "3")) {a = match.arg(a, several.ok = TRUE); print(a);}
> x()
[1] "1" "2" "3"

```

util-data.R
```diff
@@ -1103,20 +1103,32 @@ ProjectData = R6::R6Class("ProjectData",
...
-            data = self[[data.source.func]]()
-            artifacts = unique(data[[data.source.col]])
+            data = lapply(data.source.func, function(d) {
+                return(self[[d]]())
+            })
+
+            artifacts = c()
+            for (df in data) {
+                unique.artifacts = unique(df[[data.source.col]])
+                print(unique.artifacts)
```
bockthom, line 1129: Did you forget to remove the `print` statement?

nlschn: oops, sorry, yes, seems like I did...

util-data.R
nlschn: What the function calls is `match.arg(arg, choices, several.ok)` with three parameters.
For example
```
> match.arg(c("1", "2"), c("1", "2", "3"), several.ok = TRUE)
[1] "1" "2"

> match.arg(NULL, c("1", "2", "3"), several.ok = TRUE)
[1] "1"
```
In the case of `get.artifacts` the function `match.arg.or.default` is called with the parameters 
```
match.arg.or.default(data.source, several.ok = TRUE)
```
Because the parameter `choice` is missing, it is calculated from the parent function's signature and therefore is `c("commits", "mails", "issues")`. So the function will return
```
match.arg(data.sources, c("commits", "mails", "issues"), several.ok)
```
And then we are at my first snippet, when only the first element of the choice vector is returned, if `data.sources` is `NULL`.
Do I miss something here?

bockthom: This behavior of `match.arg` is weird. A missing argument is handled differently than `NULL` :cry: 
All your examples are totally correct, but they assume that the function parameter `arg` is actually set. Without that, the behavior is different, as the following example shows:

```R
> x = function(a = c("1", "2", "3")) {print(a); a = match.arg(a, choices = c("1", "2", "3"), several.ok = TRUE); print(a);}
> x()
[1] "1" "2" "3"
[1] "1" "2" "3"
> x(NULL)
NULL
[1] "1"
```
In case of a missing argument, the vector from the function declaration is taken, and since several values are ok, all of them are kept instead of just the first one...

nlschn: okay, now it is clear, although I still find it a little weird, as you said. But I will address this and try to restore the old default behaviour.

nlschn: I have now tried to fix the problem, using the same rather "hacky" method as can be found in the implementation of `match.arg.or.default`. 
One could argue to introduce yet another 'match.arg. ...' function with this behavior, just in case it is wanted elsewhere, because I assume changing `match.arg.or.default` would cause issues with exisiting calls to that function.

bockthom: We should separate the handling of default values from the actual implementation of this function. At the moment, I see two options: Either we create a new default-handling function, or we adjust the existing one.
As creating a new function may lead to confusions (which function to use when?), I would prefer to integrate the behavior into the existing one. We just need to add an additional parameter `single.default.when.several.ok` (or any other name which is more suitable). This parameter needs to default to `FALSE`. If the parameter value is actively set to `TRUE`, the corresponding if-else cascade in `match.arg.or.default` needs to be adjusted. This should not cause any issues, as we have tests for this function––but then it is be necessary to add an additional test case for that.
@hechtlC Do you agree with that?

nlschn: After trying for a while now I face a problem when attempting to implement your suggestion.
When calling `single.default.when.several.ok(arg, choices, severl.ok = TRUE)` where `arg` was not given it becomes problematic. Because inside `single.default.when.several.ok` the `arg` equals the default vector of the calling function and it is not possible any more to tell whether it has been set so deliberately or it was not given and has been replaced by the default vector.

For example:
```
> f = function(x) {print(missing(x))}
> g = function(y = c("a", "b", "c") {return(f(y))}
> g()
[1] FALSE
> f = function(x) {print(x)}
> g()
[1] "a", "b", "c"
```

I have no idea where to get the information from that in `f` the parameter was missing when calling `g`. 
Maybe you have some ideas on that?

bockthom: To answer your question:
In your example, `f` always gets a parameter, it is never missing when calling `g`, as you actually call `f(y)` inside `g`. It does not depend on where `y` comes from, `y` is always passed to `f`, so `x` is never missing. If you would like to create an example, where `x` is missing, you need to adjust `g` so that it calls `f()` without argument.

But thanks for the example, it showed me that my idea from above does not work :see_no_evil: My bad! I did not have in mind that we need to figure out whether the argument is missing or not, just thought that `match.arg` does some magic for us, but it actually doesn't.

However, I have two pretty new ideas now (without guarantee that they work as intended):

1. This idea just adjusts your implementation of `get.artifacts` and uses a trick, resulting in one statement replacing the current if-else blocks there:
    ```R
    data.source = match.arg(data.source, several.ok = !missing(data.source))
    ```
    This should exactly do what we want––please correct me if I'm wrong.

2. If the idea above does not work, we could think about adding a `missing` parameter to `match.arg.or.default` and try to handle our situation inside `match.arg.or.default` then, but this sounds a little bit ugly to me...

nlschn: I get your idea and testing it with the console I could not find a case where it goes wrong. Thanks a lot!

bockthom: I am glad that the suggested solution works this time, and we have even achieved to address this problem using just one line of code :smile: 
I have spotted just one part in this functions which could be improved (see my comment below), then we are done with this PR––at least from my view.

util-data.R
```diff
@@ -1103,20 +1103,38 @@ ProjectData = R6::R6Class("ProjectData",
...
-            artifacts = unique(data[[data.source.col]])
+            data = lapply(data.source.func, function(d) {
+                return(self[[d]]())
+            })
+
+            artifacts = c()
+            for (df in data) {
+                unique.artifacts = unique(df[[data.source.col]])
+                artifacts = append(artifacts, unique.artifacts)
+            }
```
bockthom, line 1137: First of all a question: Why not combining the different `sapply` (resp. `lapply`) statements into one `lapply`, as it is done here: https://github.com/se-sic/coronet/blob/a7f4123dcb86a7a0f76adf791b66a7d9c24f7647/util-data.R#L1377-L1380

And then I would like also to get rid of the `for` loop here, which can be easily achieved, if I did not miss anything:
`unique(unlist(lapply(data, function(d) { ... })`

And I just noticed that we adjusted the parameter name `data.source` to `data.sources` when enabling multiple data sources in `get.authors.by.data.source`. I would like to stay consistent in these two functions and rename the parameter also here. As the parameter name seems to be used nowhere outside in this function, there should be no need for further adjustments on any function call.

And finally, please also add a sentence to the changelog file that we have adjusted this function, for example:

 Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (...)

This should be the very last change on this PR, then we should be done, except @hechtlC has further comments.

bockthom (approved):

hechtlC (approved):

PR #199, opened 2021-03-29
Fix wrong data path issue and emerging bugs

### LINKED ISSUE #177: Wrong data path for range data objects
### Description

I just encountered a situation in which the data path for range data objects is actually wrong.

(For time reasons, I will just explain how I think the problem occurs, but without checking whether the MWE is correct. This is just to document the issue, I will get back to this at a later point in time.)

### Steps to Reproduce or Minimal Working Example (MWE)

1. Build the `ProjectData` object and cut its data to same dates:
```R
project.configuration <- ProjectConf$new(DATA.PATH, ANALYSIS.RANGE.TYPE, CASESTUDY, ARTIFACT)
project.configuration$update.value("mails.filter.patchstack.mails", TRUE)
project.data <- ProjectData$new(project.configuration)

# cut the data from different data sources to overlapping ranges
project.data <- project.data$get.data.cut.to.same.date(data.sources = c("mails", "commits"))

commit.data <- project.data$get.commits()
```
As a consequence, `project.data` now is a `RangeData` object, as the cutting functionality internally uses the split functionality. The split functionality generates `RangeData` objects and updates the `project.configuration` regarding the split ranges etc. Up to now, everything is fine, the commit data is retrieved correctly.

2. Update project configuration, for example, add pasta data, and retrieve mail data afterwards.
```R
project.data$update.project.conf(updated.values = list("pasta" = TRUE))
pasta <- project.data$get.pasta()
mails <- project.data$get.mails()
```
This results in empty mail data, the mail data cannot be read, as the following warning states:
```R
Warning message:
In file(file, "rt") :
  cannot open file './testdata/results/threemonth/teststudy_feature/feature/001--2003-01/commits.list': No such file or directory
```
The reason for that is, that the `read` functions get the data path as parameter, which for `RangeData` objects is determined by:

https://github.com/se-passau/coronet/blob/91fc448bad64dd17fa96a720322fe7d46337cc43/util-data.R#L1408-L1411

And `ranges.paths` contains the following, due to data cutting in the beginning:

```R
 ranges.paths [final]
  - value
    - 2003-01-14 21:44:38-2016-02-18 21:50:25 = 001--2003-01
```
However, this actually wrong, as the directory `001--2003-01` does neither exist, nor should it be used at all. This kind of ranges and range directories is related to Codeface ranges (and also to module `util-bulk`), but not to split ranges created by coronet. Therefore, the function `get.data.path` of `RangeData` objects does not return the expected data path.

**Actual behavior:** Data is to be read from a non-existing directory.

**Expected behavior:** `get.data.path` of `RangeData` objects should return the same data path as the corresponding project data. However, this does not respect the previous data cutting (or splitting, in general). Maybe it would be an idea to read in the data and then split it to the ranges set in the project configuration? Would this be an appropriate solution?

We should make this more flexible to be compatible with Codeface ranges (maybe additional boolean flag necessary) and data cutting. Also automatic data cutting/splitting could cause further trouble, but it is necessary to stay consistent with the remaining range data.

### Versions

This affects several versions, including the most recent one (3.6). As far as I have seen, this problem exists, at least, since version 3.0.

I just figured this out by chance. In my script, I just rearranged the order of the get and update-config statements, to avoid this situation. However, in general, for the long term, we should come up with a solution for this problem, even for consistency reasons.

### LINKED ISSUE #197: Various bugs or inconsistencies regarding the data, conf, and network modules
While reflecting on #177, I spotted the following issues:

1. The function [`reset.environment()` in the ProjectData class](https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-data.R#L542) does not reset all the data, as `mails.patchstacks` is missing. (In addition, check if something else is missing there.)
2. The function [`reset.environment()` in the NetworkBuilder class](https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-networks.R#L557) does not reset all the data, as `artifacts.network.mail` and `artifacts.network.issue` are missing. (In addition, check if something else is missing there.)
3. The documentation of the function [`check.value` in the Conf class](https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-conf.R#L102) is incomplete, as the description of the return value is wrong, as, for instance, the value of `updatable` is missing, and not all values are returned always.
4. A comment in the function `get.pasta` of the data module is misleading:    "no storage" makes us believe that mails might not be stored, which is actually wrong. Instead, we should just state "no assignment", as the mails are not assigned to any variable here in this function.
    https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-data.R#L919-L920
5. When calling the function `update.pasta.data`, a warning is printed that one should think about calling the cleanup function:
    https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-data.R#L425-L426 However, this warning is also printed when calling the cleanup function itself, as the cleanup function in turn calls the update function. To prevent this silly behavior (which would end up in an endless recursion if the user would call the cleanup function again and again due to this warning), we could try to only print this warning of the function is **not** called by the cleanup function. I am not sure if this is possible, but it might be worth a try. @nlschn you already have some experience in finding out what the parent function call was (remember your work regarding on the `match.arg.or.default` function), you could check whether this would be possible here. If so, we should also check for the other cleanup functions whether we also end up in such silly warnings there.

nlschn:
### Prerequisites

### Description

Closes #197.
Closes #177.

#### Issues from #197  
- [x] The function `reset.environment()` in the ProjectData class does not reset all the data .
- [x] The function `reset.environment()` in the NetworkBuilder class does not reset all the data.
- [x] The documentation of the function `check.value` in the Conf class is incomplete.
- [x] A comment in the function get.pasta of the data module is misleading.
- [x] When calling the function `update.pasta.data`, a warning is printed that one should think about calling the cleanup function.

#### Issues from #177 
- [x] When the object is built from range data, we use the data path to the codeface-range folders and otherwise the "normal" data path. The re-reading of data sources from the associated data path will then be allowed and enabled.
- [x] Test the new functionality. 

### Changelog
- Update documentation in `util.conf.R` (PR #199, 059b286a3cebc586e88ce6446d98133ddd619260)
- Fix the function `reset.environment()` of both the `ProjectData` and `NetworkBuilder` class; they now reset all the data (PR #199, 87a243f5fea4d9c40aa5418965fff4e0c23c3358)
- Adjust the functions `update.commit.message.data()`, `update.pasta.data()`, and `update.synchronicity.data()`: no warning is being printed anymore when being called by the corresponding cleanup function (PR #199, e5c60a50f8fa0f5bf9d362fdf49845d58652dd75)
- Fix issue where the data path on `RangeData` objects was wrong in special cases. Introduce a (private) flag that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, cce95279692b8d29ae464e12cfe48e923f417ac7, 917bf64e7a439c6a33b922e04929030479722bd1). Also add tests for this new behaviour (PR #199, ef5bac63580d171dd7f68da6c4ec7279dcd400b5, 3aa8e7de4b847924afa350a1fc4a53a7f8fa86e1)


NEWS.md
```diff
@@ -19,11 +19,12 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
 - Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (PR #195, cf795f26652b00de5d717c703c688af55a972943, 70c05ecd1e3c0f10810acc2b2ae06a3eb8856317, 5a46ff4d428af7f301fe57d6e9e10421f553a9cc, fd767bb37ca608c28d9ff4a449415cc0e863d7ee)
-
-
+- Fix the function `reset.environment()` of both the `ProjectData` and `NetworkBuilder` class; they now reset all the data (PR #199, 87a243f5fea4d9c40aa5418965fff4e0c23c3358)
+- Adjust the function `update.pasta.data()`: no warning is being printed anymore when calling the cleanup function (PR #199, 432c66b839932660f144050c6399dc9112b023f3)
```
bockthom, line 26: If you mention pasta, then you should also mention the other two functions where this holds :wink: 
Moreover, instead of "when calling the cleanup function", it should rather be "when being called from the corresponding cleanup function"

NEWS.md
```diff
@@ -19,11 +19,12 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
+- Fix issue where the data path on `ProjectData` objects was wrong in special cases. Introduce a (private) flag that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, a5063ccdd7b4da75afd2c71655615b182b9f0f28, c2bf849f3b3ae902d411429298e853b748216b9c) Also add tests for this new behaviour (PR #199, 28cf3db267c68dc0892c2f63f741434ab165d079, 033567f3d1a4d43b188bf160edcd86779727965c)
```
bockthom, line 27: The data path on `RangeData` objects was wrong, not on `ProjectData` objects. If you start a new sentence, don't forget the period at the end of the sentence before.

NEWS.md
```diff
@@ -19,11 +19,12 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 ## 3.7
```
bockthom, line 28: All the three bullet points would rather fit to category "### Fixed" (but this again will cause merge conflicts later on).
And please don't remove the blank lines between your fixes and the last release number.

tests/test-split.R
```diff
@@ -1772,3 +1772,50 @@ test_that("Check and correct duplicate range names during network activity-based
...
+
+    # cut the data from different data sources to overlapping ranges
+    project.data = project.data$get.data.cut.to.same.date(data.sources = c("mails", "commits"))
+
+    commit.data = project.data$get.commits()
+
+    ## Update project configuration, for example, add pasta data, and retrieve mail data afterwards
+    project.data$update.project.conf(updated.values = list("pasta" = TRUE))
+    pasta <- project.data$get.pasta()
+    mails <- project.data$get.mails()
```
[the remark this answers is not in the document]
nlschn: I am a bit confused though. I have added a check which reflects my understanding of your post; but I am not 100% sure whether this understanding is right. I'd be happy to read your comments on this changes in the next review stage.

tests/test-split.R
```diff
@@ -1772,3 +1772,50 @@ test_that("Check and correct duplicate range names during network activity-based
...
+    result = project.data$get.data.path()
+
+    expect_identical(result, expected, info = "RangeData data path.")
+})
+
+test_that("Split data with bulk methods and check data path", {
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    data = construct.data(proj.conf, callgraphs = TRUE)
+
+    range.paths = run.lapply(data, "get.data.path")
```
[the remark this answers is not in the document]
nlschn: Always glad to help 😇

util-bulk.R
```diff
@@ -81,14 +82,14 @@ collect.multi.networks = function(project.conf, network.conf, step = 1) {
 #' @return the bipartite networks
 collect.bipartite.networks = function(project.conf, network.conf, step = 1) {
     ## we need to iterate over all ranges
-    ranges = project.conf$get.value("ranges")
+    ranges = project.conf$get.value("ranges", TRUE)
```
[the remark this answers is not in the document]
nlschn: I have probably pasted `, TRUE` to the wrong place. Thanks

util-data.R
```diff
@@ -1572,9 +1618,20 @@ RangeData = R6::R6Class("RangeData", inherit = ProjectData,
...
 
+        ## * * getter for private attributes -------------------------------
+
+        #' Getter for the private logical \code{built.from.range.data} attribute
+        #'
+        #' @return the value of \code{built.from.range.data} attribute
+        is.built.from.range.data = function() {
+            return(private$built.from.range.data)
+        },
+
```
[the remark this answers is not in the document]
nlschn: I can easily delete it if it is not necessary.

util-bulk.R
```diff
@@ -45,7 +46,7 @@ collect.multi.networks = function(project.conf, network.conf, step = 1) {
     ## collect the network objects for all the ranges
     networks = lapply(ranges, function(range) {
         ## construct range data
-        range.data = RangeData$new(project.conf, range)
+        range.data = RangeData$new(project.conf, range, built.from.range.data = TRUE)
```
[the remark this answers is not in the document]
nlschn: There is nothing to replace in `showcase.R` (at least nothing I am aware of

tests/test-split.R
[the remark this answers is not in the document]
bockthom: I had a quick look at your changes, your understanding is right. I would just suppose to additional changes:
1. Please move the check for the correct data path (`expected`, `result`, and `expect_identical`) up directly below the function call `get.data.cut.to.same.date`. Everything what is done below is not needed for this particular check.
2. Please adjust the comments for the second check at the end: At the moment, there are two comments stating that something is checked after resetting.

util-data.R
```diff
@@ -246,8 +246,22 @@ ProjectData = R6::R6Class("ProjectData",
...
+            caller = NULL
+
+            ## only assign the last but one element if the stack is long enough
+            if (length(stacktrace) >= 2) {
+                caller = tail(stacktrace, n=2)[[1]]
+            }
+
+            ## only print warning if this function has not been called by 'cleanup.commit.message.data' including the
+            ## case that it is called manually, i.e. the stack is too short.
+            if(caller != "cleanup.pasta.data()") {
```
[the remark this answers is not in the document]
bockthom: Btw, there is a space missing after `if`. Also holds for similar statements (even after the changes you already did here).

util-bulk.R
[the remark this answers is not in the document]
bockthom: I am not sure if this really is the case. 
Let's consider the following statement, which is part of the showcase file: `RangeData$new(project.conf = proj.conf, range = ranges[[22]])`
The `ranges` seem to come from the project configuration and not from splitting. Please print these ranges and check where they come from.
And there are similar statements in the ProjectConf.
It might be the case that we don't need to set the flag here, but I am not sure about this.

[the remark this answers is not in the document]
nlschn: When running the showcase with the test data it spit out the two codeface ranges.

tests/test-data.R
```diff
@@ -266,3 +266,48 @@ test_that("Merge commit message titles to commit data", {
...
+## Data paths of RangeData -------------------------------------------------
+
+##
+## Tests that check RangeData objects for the right data path
+##
+
+test_that("Cut data and check for right data path", {
+    ## Build the ProjectData object and cut its data to same dates
+    project.configuration = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    project.configuration$update.value("mails.filter.patchstack.mails", TRUE)
```
hechtlC, line 280: Why are you filtering patchstack mails here?

util-misc.R
```diff
@@ -202,6 +202,21 @@ create.empty.data.frame = function(columns, data.types = NULL) {
...
 
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Vector misc--------------------------------------------------------------
+
+#' Get the last element of a vector but one.
+#'
+#' @param v the vector of which to retrieve the last element but one
+#'
+#' @return the last element but one of \code{v}. If the vector has less than two elements, return \code{NA}.
+get.last.but.one.element = function(v) {
```
hechtlC, line 213: I think the name and description of this function can be a bit confusing. Am I understanding it correctly that you get the last element of a vector **if** the vector has a least 2 elements? I think you should clarify that. It is explained in the @return tag but I think it should be explained in the description itself.
What do you think @bockthom ?

hechtlC (changes requested):
Nice work @nlschn!
I only have two minor remarks which I added as comments.
One other thing is that you forgot the copyright header in util-misc.R so please add that.

tests/test-data.R
bockthom: I'd guess that this was part of my minimum working example. I may have added it to somehow read the e-mails implicitly at the beginning. It does not hurt to keep that one also in the test, but as it seems not to be necessary for this test, we could also remove it.

nlschn:
Since the empty merge conflict is _before_ the logo commits that I pulled, every time I try to remove the empty commit it changes the hashes of @bockthom's commits concerning the logo. Is there any possibility how to avoid that?

bockthom:
> Since the empty merge conflict is _before_ the logo commits that I pulled, every time I try to remove the empty commit it changes the hashes of @bockthom's commits concerning the logo. Is there any possibility how to avoid that?

Maybe the following steps will do it (without any guarantees –– so make sure to not get lost of anything):

1. Create a new local branch from your branch `mydev`: `checkout -b mydev2`
2. Create a new local branch from your branch `mydev`: `checkout -b mydev3`
3. On `mydev3`, make a hard reset to the merge commit from the logo PR: `git reset --hard bf8b331cd874f6fef787af0a6ad69c2a0496d3bb`
4. On `mydev2` (`git checkout mydev2`), make a hard reset to your last commit before the merge commit we want to remove: `git reset --hard 87a243f5fea4d9c40aa5418965fff4e0c23c3358`
5. Rebase this one on the previous dev after the logo PR: `git rebase mydev3`. After that, everything before the merge commit you want to remove should be correctly rebased.
6. On `mydev` (`git checkout mydev`), rebase on the stuff we have cleaned before: `git rebase mydev2`
7. Then everything should be fine (please check if all commits are still there and the unwanted merge commit is removed and commits up to bf8b331 have not changed their hashes).

I hope that this would solve the problem, if I did not overlook anything.

util-misc.R
```diff
@@ -202,6 +202,21 @@ create.empty.data.frame = function(columns, data.types = NULL) {
...
 
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Vector misc--------------------------------------------------------------
+
+#' Get the last element of a vector but one, if it has, at least, two elements.
+#'
+#' @param v the vector of which to retrieve the last element but one
+#'
+#' @return the last element but one of \code{v}. If the vector has less than two elements, return \code{NA}.
+get.last.but.one.element = function(v) {
```
clhunsen, line 213: Would you consider renaming this function to `get.second.last.element`? As @hechtlC, I myself was confused by reading "last but one" (and variants!) in the description text. Writing and reading about the "second-last element" is less confusing, I guess.

nlschn:
I have managed to get rid of the empty merge commit, pulled the commits (using `--rebase` this time...) from @JoJoDeveloping 's PR and tried to manually resolve the conflicts (as predicted there were a few). As soon as you have reviewed this, I am going to change the hashes in the changelog  to the new ones and then it should be ready to go.

tests/test-data.R
hechtlC: Ok so please remove this @nlschn .

hechtlC:
Please don't forget about your copyright header in `util-misc.R` @nlschn .

tests/test-data.R
nlschn: This should have happened with e0145e2b72aa9c6be706fd6f08abd8d7da811215

NEWS.md
```diff
@@ -22,14 +22,16 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (PR #195, cf795f26652b00de5d717c703c688af55a972943, 70c05ecd1e3c0f10810acc2b2ae06a3eb8856317, 5a46ff4d428af7f301fe57d6e9e10421f553a9cc, fd767bb37ca608c28d9ff4a449415cc0e863d7ee)
 - Rename `get.issues` to `get.issues.filtered`, and write a new `get.issues` to get the unfiltered issues so that these methods follow the naming scheme known from the respective methods for commits(b9dd94c8575b8cab40d0d1185368854f84299d87)
 
 ### Fixed
 - Fix fencing issue timing data so that issue events "happen" after the issue was created. Since only `commit_added` events are affected, that only happens for these. (issue #185, 627873c641410182ca8fee0e78b95d7bda1e8e6b, 6ff585d9da1da3432668605f0c09f8e182ad0d2f)
-
+- Fix the function `reset.environment()` of both the `ProjectData` and `NetworkBuilder` class; they now reset all the data (PR #199, 87a243f5fea4d9c40aa5418965fff4e0c23c3358)
```
bockthom, line 32: This commit hash is still outdated...

NEWS.md
```diff
@@ -22,14 +22,16 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
+- Adjust the functions `update.commit.message.data()`, `update.pasta.data()`, and `update.synchronicity.data()`: no warning is being printed anymore when being called by the corresponding cleanup function (PR #199, e5c60a50f8fa0f5bf9d362fdf49845d58652dd75)
+- Fix issue where the data path on `RangeData` objects was wrong in special cases. Introduce a (private) flag that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, cce95279692b8d29ae464e12cfe48e923f417ac7, 917bf64e7a439c6a33b922e04929030479722bd1). Also add tests for this new behaviour (PR #199, ef5bac63580d171dd7f68da6c4ec7279dcd400b5, 3aa8e7de4b847924afa350a1fc4a53a7f8fa86e1)
```
bockthom, line 34: Could you please add the name of the (private) flag  `...` here? Then please also reference commit 
169c03 as third commit here.
For the tests, please also add the commit hashes d454e5a and 66ad127 as in this two commits the test behavior is substantially updated.

bockthom (commented):
Almost done. Looking forward to merging this PR soon.

bockthom (approved):

hechtlC (approved):

PR #200, opened 2021-04-22
Fix wrong ProjectConf after various data splittings

### LINKED ISSUE #198: Wrong ProjectConf after various data splittings
While reflecting on #177, I also identified that when splitting data, all the resulting RangeData objects use the same ProjectConf (and yes, all of them have a reference to exactly the same ProjectConf object). This is ok for all the RangeData objects which result from one splitting, but it is wrong when performing different splittings, as the following example might show:

First a ProjectConf object is created which is used to create a ProjectData object named `data`.
Then we split `data` into several 6-month RangeData objects using `split.data.time.based`.
When we then split `data` again into several 12-month RangaData objects again using `split.data.time.based`, then all ProjectData and all RangeData objects contain a reference to exactly the same ProjectConf object.
However, this does not make sense, as data splitting adjusts the ProjectConf object, as it sets the revisions, see for example here: https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-split.R#L146 or here: https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-split.R#L188 (and maybe also at additional similar lines).

I already talked to @hechtlC and we agreed on the following solution:

Whenever we split data, all the resulting RangeData objects should get a new ProjectConf object. That is, all the RangeData objects which result from the same splitting get the same ProjectConf, but if we perform a new splitting, then we generate a new ProjectConf. Hence, when splitting data, we need to create a clone of the existing ProjectConf (from the data object to split) and let all the resulting RangeData objects point to this clone. In the cloned ProjectConf object which is used by all the RangeData objects, just the splitting information needs to be updated and the rest could be kept as is: https://github.com/se-sic/coronet/blob/bf8b331cd874f6fef787af0a6ad69c2a0496d3bb/util-split.R#L195-L203.

When cloning a ProjectConf object, just calling
```R
project.conf$clone()
```
should be enough. However, we need to find out whether we need a deep clone or whether a shallow clone of an `R6` class is already sufficient.

JoJoDeveloping:

### Description

Make splitting create new configuration objects for the split data results, as discussed in #198.

Closes #198

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,7 +33,7 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Fix the function `reset.environment()` of both the `ProjectData` and `NetworkBuilder` class; they now reset all the data (PR #199, de091a5e6c70fea5161276d6585bea916178e4de)
 - Adjust the functions `update.commit.message.data()`, `update.pasta.data()`, and `update.synchronicity.data()`: no warning is being printed anymore when being called by the corresponding cleanup function (PR #199, e5c60a50f8fa0f5bf9d362fdf49845d58652dd75)
 - Fix issue where the data path on `RangeData` objects was wrong in special cases. Introduce the (private) flag `built.from.range.data.read` that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, cce95279692b8d29ae464e12cfe48e923f417ac7, 917bf64e7a439c6a33b922e04929030479722bd1, 169c034c2364fc56507573e6f2316cc432631ba6). Also add tests for this new behaviour (PR #199, ef5bac63580d171dd7f68da6c4ec7279dcd400b5, 3aa8e7de4b847924afa350a1fc4a53a7f8fa86e1, d454e5a3468d765247ed4827e708b8576e54f87c, 66ad1272148e3d98f9508ff0e83180d967b6abaa)
-- Make splitting no longer modify the original ProjectConf, instead create a copy (c894ccefc0554c85281af74ca8a16e26e6072543)
+- Make splitting no longer modify the original ProjectConf, instead create a copy (e82d056b7a67288addb5989ded1e8c921d5ed0a4)
 
 ## 3.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,7 +34,6 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Adjust the functions `update.commit.message.data()`, `update.pasta.data()`, and `update.synchronicity.data()`: no warning is being printed anymore when being called by the corresponding cleanup function (PR #199, e5c60a50f8fa0f5bf9d362fdf49845d58652dd75)
 - Fix issue where the data path on `RangeData` objects was wrong in special cases. Introduce the (private) flag `built.from.range.data.read` that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, cce95279692b8d29ae464e12cfe48e923f417ac7, 917bf64e7a439c6a33b922e04929030479722bd1, 169c034c2364fc56507573e6f2316cc432631ba6). Also add tests for this new behaviour (PR #199, ef5bac63580d171dd7f68da6c4ec7279dcd400b5, 3aa8e7de4b847924afa350a1fc4a53a7f8fa86e1, d454e5a3468d765247ed4827e708b8576e54f87c, 66ad1272148e3d98f9508ff0e83180d967b6abaa)
 - Make splitting no longer modify the original ProjectConf, instead create a copy (c894ccefc0554c85281af74ca8a16e26e6072543)
->>>>>>> 0cb5372 (Adjust NEWS.md)
 
 ## 3.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,8 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Fix the function `reset.environment()` of both the `ProjectData` and `NetworkBuilder` class; they now reset all the data (PR #199, de091a5e6c70fea5161276d6585bea916178e4de)
 - Adjust the functions `update.commit.message.data()`, `update.pasta.data()`, and `update.synchronicity.data()`: no warning is being printed anymore when being called by the corresponding cleanup function (PR #199, e5c60a50f8fa0f5bf9d362fdf49845d58652dd75)
 - Fix issue where the data path on `RangeData` objects was wrong in special cases. Introduce the (private) flag `built.from.range.data.read` that is set according to how the object has been created (splitting manually or reading codeface ranges) and calculating the data path accordingly (PR #199, cce95279692b8d29ae464e12cfe48e923f417ac7, 917bf64e7a439c6a33b922e04929030479722bd1, 169c034c2364fc56507573e6f2316cc432631ba6). Also add tests for this new behaviour (PR #199, ef5bac63580d171dd7f68da6c4ec7279dcd400b5, 3aa8e7de4b847924afa350a1fc4a53a7f8fa86e1, d454e5a3468d765247ed4827e708b8576e54f87c, 66ad1272148e3d98f9508ff0e83180d967b6abaa)
+- Make splitting no longer modify the original ProjectConf, instead create a copy (c894ccefc0554c85281af74ca8a16e26e6072543)
+>>>>>>> 0cb5372 (Adjust NEWS.md)
 
 ## 3.7
 
--- a/tests/test-split-sliding-window.R
+++ b/tests/test-split-sliding-window.R
@@ -88,7 +88,9 @@ test_that("Split a data object time-based (split.basis = 'commits', sliding.wind
         "2016-07-12 16:04:59-2016-07-12 16:06:33"
     )
     lapply(results, function(res) {
-        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")})
+        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")
+    })
+    expect_equal(proj.conf$get.value("ranges"), c("v1-v2", "v2-v3"))
 
     ## check data for all ranges
     expected.data = list(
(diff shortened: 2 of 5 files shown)


tests/test-split-sliding-window.R
```diff
@@ -86,9 +87,8 @@ test_that("Split a data object time-based (split.basis = 'commits', sliding.wind
         "2016-07-12 16:03:29-2016-07-12 16:06:29",
         "2016-07-12 16:04:59-2016-07-12 16:06:33"
     )
-    result = proj.conf$get.value("ranges")
-
-    expect_equal(result, expected, info = "Time ranges.")
+    lapply(results, function(res) {
+        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")})
```
bockthom, line 91: Could you please move the `})` at the end to a new line and align the new one with the indentation of `lapply`?
(This also holds for similar statements below and also in the other touched test files.)

In addition, shouldn't we also test that the original `proj.conf` is not adjusted, to make sure that we indeed have different object instances here?

util-split.R
```diff
@@ -185,15 +195,15 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
             ## re-set project configuration due to object duplication
-            cf.conf = cf$set.project.conf(project.data$get.project.conf())
+            cf.conf = cf$set.project.conf(project.conf.new)
         }
     }
 
     ## add splitting information to project configuration
-    project.data$get.project.conf()$set.splitting.info(
+    project.conf.new$set.splitting.info(
         type = "time-based",
```
bockthom, line 207: Can we also add some checks for that in some of the existing splitting tests, to make sure that the splitting info is adjusted correctly everywhere (that is, on the cloned conf objects but not on the original one) ?

NEWS.md
```diff
@@ -22,6 +22,7 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
 - Update documentation in `util-network-metrics.R` (PR #195, f929248182594613bd203e100268e3e3dce87f34, de9988cc171cafdd084701d5a2693a74176a802a)
 - Add check for empty network in `metrics.hub.degree` function. In the case of an empty network, a warning is being printed and `NA` is returned (PR #195, 4b164bebea1e8258cb93febf51271a4b6f486779)
 - Adjust the function `ProjectData$get.artifacts`: Rename its single parameter to `data.sources` and change the function so that it can extract the artifacts for multiple data sources at once. The default is still that only artifacts from the commit data are extracted. (PR #195, cf795f26652b00de5d717c703c688af55a972943, 70c05ecd1e3c0f10810acc2b2ae06a3eb8856317, 5a46ff4d428af7f301fe57d6e9e10421f553a9cc, fd767bb37ca608c28d9ff4a449415cc0e863d7ee)
+- Make splitting no longer modify the original ProjectConf, instead create a copy (70e73ea16e09fbb2b4bd9be38a19e89c798d4d8e)
```
bockthom, line 25: I would rather move that to the `### Fixed` section, as the previous behavior was actually wrong.
@hechtlC What is your opinion on that?

bockthom (commented):
Nice work @JoJoDeveloping , I like it. :+1: 
I only have some minor comments on how we could further improve our tests.

tests/test-split-sliding-window.R
```diff
@@ -88,7 +88,9 @@ test_that("Split a data object time-based (split.basis = 'commits', sliding.wind
         "2016-07-12 16:04:59-2016-07-12 16:06:33"
     )
     lapply(results, function(res) {
-        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")})
+        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")
+    })
+    expect_equal(proj.conf$get.value("ranges"), c("v1-v2", "v2-v3"))
```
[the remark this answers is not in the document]
JoJoDeveloping: `info` means into the documentation of that "method"?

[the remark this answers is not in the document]
bockthom: I'm not sure what you are referring to. I refer to the `info` parameter of the `expect_equal` function (see also two lines above) that provides a very short notice on what is covered with the `expect_equal` statement.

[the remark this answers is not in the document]
JoJoDeveloping: Ah, I was not aware that parameter existed.
Will do.

util-split.R
JoJoDeveloping: Is there a way to get more than 1 value from the config? Something like `$get.values(c("foo", "bar"))` which then returns a dataframe, or a vector, or something like that??

bockthom: No, i don't think so. (There is a function `get.conf.as.string` which dumps the complete config with all its values, but this is not what you are looking for, I guess).

JoJoDeveloping: I would propose we do something like this
```
expected = list(
    split.type = "time-based",
    split.length = "3 min",
    split.basis="commits"
    split.sliding.window = FALSE,
    split.revisions = c("2016-07-12 15:58:59", "2016-07-12 16:01:59", "2016-07-12 16:04:59", "2016-07-12 16:06:33"),
    split.revision.dates=NULL
)
actual = lapply(names(expected), function(x) {the.split.config$get.entry(x)})
expect_equal(expected,actual)
```

JoJoDeveloping: These 6 entries are set by `set.splitting.info`. It also sets `split.ranges`, however that is constructed from these 6 values within `set.splitting.info`, so we don't need to test for it if we only want to assert that the parameters of that function are set correctly

tests/test-split-sliding-window.R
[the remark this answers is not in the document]
JoJoDeveloping: like this?
```
    lapply(results, function(res) {
        expect_equal(res$get.project.conf()$get.value("ranges"), expected, info = "Time ranges.")
    })
    ## `c("v1-v2", "v2-v3")` is the default, make sure that did not change
    expect_equal(proj.conf$get.value("ranges"), c("v1-v2", "v2-v3"), info = "splitting modified original ProjectConf!")
    ```

File: util-split.R (hunk 2 above)
Comment (bockthom) [reply util-split.R:207]: Yes, sounds like a good idea (if you also fix the spacing issues etc...)

File: tests/test-split-sliding-window.R (hunk 4 above)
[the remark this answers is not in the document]
Comment (bockthom) [reply tests/test-split-sliding-window.R:93]: I would like to start with "make sure that the original project configuration did not change" (and then you could explain the default in parentheses, if wanted).

Regarding the info parameter: Your suggestions specifies the opposite of what is expected, if I am correct.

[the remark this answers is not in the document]
Comment (JoJoDeveloping) [reply tests/test-split-sliding-window.R:93]: The info is what is printed out when the assert fails. Thus  I made it denote the error condition.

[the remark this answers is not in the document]
Comment (bockthom) [reply tests/test-split-sliding-window.R:93]: That might be actually true (but we never have used it in coronet in such a way to denote the error condition, if I am true, but rather phased it in a neutral way).
So, I don't care, phrase it like you would like to and let's @hechtlC decide on that :wink:

File: tests/test-split-sliding-window.R [hunk 5]
```diff
@@ -390,7 +443,23 @@ test_that("Split a data object time-based (bins = ... , sliding.window = TRUE)."
...
-    expect_equal(proj.conf$get.value("ranges"), c("v1-v2", "v2-v3"))
+    ## This value should not change, so we compare it with the default, which is `c("v1-v2", "v2-v3")`.
+    expect_equal(proj.conf$get.value("ranges"), c("v1-v2", "v2-v3"),
+                 info = "splitting must mot modify the original ProjectConf.")
+    ## test that the config contains the correct splitting information
+    expected.config = list(
+        split.type = "time-based",
+        split.length = c("2016-01-01 00:00:00", "2016-12-31 23:59:59", "2017-06-03 03:03:03"),
+        split.basis = NULL,
+        split.sliding.window = FALSE, ## TODO is this the supposed value
```
JoJoDeveloping, line 454: split.sliding.window is `FALSE` as per this:
https://github.com/se-sic/coronet/blob/9a4ccb156127fd3ec001fb5ec534c6c870f64590/util-split.R#L103

Is this intended?

JoJoDeveloping (commented):
Is this a bug?

tests/test-split-sliding-window.R
bockthom: > Is this intended?

Yes, it is. When bins are specified, we cannot have sliding windows. Thus, the `sliding.window` parameter is ignored in this case. See also the comment which describes this test:
https://github.com/JoJoDeveloping/coronet/blob/dc74235abb7b3eab5c4347b5e55b64d319333727/tests/test-split-sliding-window.R#L412

---

Btw, could you please add a blank line before each of your comments `## This value should...` and `## test thatthe config ...` as this would ease reading.

util-split.R
```diff
@@ -277,6 +277,10 @@ split.data.activity.based = function(project.data, activity.type = c("commits",
         }
         ## compute the number of time windows according to the activity amount
         number.windows = ceiling(activity / activity.amount)
+        if (activity < activity.amount) {
+            logging::logwarn("Can not form bins of %s %s for splitting data %s, as there are only %s data rows.",
+                             activity.amount, activity.type, project.data$get.class.name(), activity)
```
bockthom, line 282: This warning is a little bit imprecise in one specific case, namely when splitting issues:
If `activity.type` is `issues`, then you would print `Can not form bins of x issues ...`. However, we don't count issues, but activities, that is, issue events. Could you please create a variable that takes the `activity.type` when talking about commits or mails, but uses "issue events" in case of issues and use the new variable instead of `activity.type` when printing the warning. Please also use the variable again at the end instead of "data rows", to also be more precise here, since we don't count data rows but activities.

bockthom (changes requested):
Thanks @JoJoDeveloping, your PR looks really good now. I only have one minor comment regarding the warning you have added, as it is a little bit imprecise in certain cases - please see my comment below. (And when you fix that, please reword the commit message as we don't have "javadoc" as it is not java, just talk about function documentation.)

As soon as you have fixed that, please rebase onto the current `dev` branch and make sure that the commit hash mentioned in the line you have added to the NEWS.md gets updated correctly. Then we can merge this PR right away.

bockthom:
Signed-off tags are still missing in the last few commits...

bockthom (approved):

PR #201, opened 2021-05-13
Enable to read data at every point in time

nlschn:
### Prerequisites

### Description

This PR adresses #177, specifically [this comment](https://github.com/se-sic/coronet/issues/177#issuecomment-806912711)

### Changelog

See 6d5c3aba099484a490498fc405def881047fa015


nlschn:
When trying to find the many bugs I created, I just stumbled upon an interesting question.
With the new functionality, that not all data sources are read when splitting, there is a possibility that when splitting e.g. by commits, that the commits have not been read yet. This happens for example in `test_that("Check consistency of data and network time-based splitting."` (`tests-split.R`, line 1638). The test fails because there is no commit data, that it can read the dates from which results in a rather cryptic error message (missing value where TRUE/FALSE was expected). But when I call `project.data$get.commits()` before actually splitting, it works (or rather fails with a different message, but that's something else...) so now it has the data and the corresponding "date" column. 
Long story short: My question is, what should happen in this case? I think we should at least check if the data source you are trying to split by is present and print a warning if not. To go further, one could automatically read the necessary data in order to enable immediate splitting. What do you think about that?

bockthom:
> When trying to find the many bugs I created, I just stumbled upon an interesting question.

If you got stuck during bug fixing, we can have a closer look at the bugs together in our next meeting and have a joint.

> Long story short: My question is, what should happen in this case? I think we should at least check if the data source you are trying to split by is present and print a warning if not. To go further, one could automatically read the necessary data in order to enable immediate splitting. What do you think about that?

We should automatically get the data that is needed for splitting. So, for example, when we split by commits, we should call the getter for the commit data within the splitting function before the actual splitting is performed. I guess this should fix the problem in general.

Anyway, it can still be the case that the commit data actually is empty. In such a case, we could think about printing a warning but I am not sure what happens in such a case. If it leads to an error, we maybe should throw an error instead of warning. But we can discuss this on Thursday in detail.

nlschn:
> If you got stuck during bug fixing, we can have a closer look at the bugs together in our next meeting and have a joint debugging session.

At the moment the number of rows in the data of RangeData objects does not match with the expected data. I will try to find the reason for this until friday, but if I don't succeed, I would gladly take your offer.

> We should automatically get the data that is needed for splitting. So, for example, when we split by commits, we should call the getter for the commit data within the splitting function before the actual splitting is performed. I guess this should fix the problem in general.

This shouldn't be that big of a problem.

nlschn:
At last, this PR is now ready to be reviewed. All tests in all versions finally run and I brought README and NEWS up to date with the latest changes and additions.

util-misc.R
```diff
@@ -873,3 +873,33 @@ get.range.bounds = function(range) {
 
     return (range)
 }
+
+
+#' Get a single bin from a range
```
hechtlC, line 878: Please add a .

util-read.R
```diff
@@ -348,15 +348,17 @@ read.issues = function(data.path, issues.sources = c("jira", "github")) {
...
-    ## violations can happen for "commit_added" events if the commit was made before the PR was opened
-    ## the original date for "commit_added" events is stored in "event.info.2" in any case
-    commit.added.events = issue.data[["event.name"]] == "commit_added"
-    issue.data[commit.added.events, "event.info.2"] = get.date.string(issue.data[commit.added.events, "date"])
-    commit.added.events.before.creation = commit.added.events &
-        !is.na(issue.data["creation.date"]) & (issue.data["date"] < issue.data["creation.date"])
-    issue.data[commit.added.events.before.creation, "date"] = issue.data[commit.added.events.before.creation, "creation.date"]
-    issue.data = issue.data[order(issue.data[["date"]], decreasing = FALSE), ] # sort!
+    if (nrow(issue.data) > 0) {
+        ## fix all dates to be after the creation date.
```
hechtlC, line 352: could you please remove the dot here?

hechtlC (changes requested):
Thank you for the nice enhancements and fixes @nlschn! 
I have just found some minor issues. But these need fixing nevertheless. Some of these are not of your making but since you are already touching the lines you can fix them as well :)

NEWS.md
```diff
@@ -19,14 +19,18 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add R version 4.1 to test suite and adjust missing time-zone attributes on `NA` vectors or empty POSIXct vectors which are correctly added as of R version 4.1 (PR #203, 6b7fb36478325185f80f95e03dd0a0135bf44346, 98c56715502cf5140d98796dc2d6cb18f71760b2, 09d11ab631de8051ca317f36652ab0fee9cc0cce)
+- Splitting no longer loads all (additional) data sources, but only the ones that have already been cached in the 'ProjectData' (PR #201, 5894e7d5213dc56b4ac2c8ea1f44b3461284a39a, 217d9c1f08770cc82f7c3f4775305e29e4d92827, 0403964469d106166104371e69be826078b2a673).
+- Change the internal representation of empty data from `NULL` to empty data frames and adapt function `get.cached.data.sources()` of ProjectData which returns a vector of all data sources that are cached (including additional and filtered data sources) (PR #201 aec898e105ff5478fb75ba488b56c9c258a17a80, e55d088f926ae7a069761864855c179c04ed07dc); additionally, introduce new function `is.data.source.cached()` in `util-data.R` that returns a logical vector indicating which of the given data sources are cached (PR #201, b49cc5df4da04af9df2ae9bcdc7847fafb6befc9).
```
bockthom, line 33: Please update the outdated hashes...

bockthom (changes requested):
I like this PR @nlschn, really nice improvements to data and splitting functionality of coronet. Good things come to those who wait!

I only found one actual bug and a couple of minor issues regarding documentation and consistency. Please see my comments below.

In addition, if I have seen it correctly, your copyright header is missing in "test-networks-covariates.R". Please don't forget to update that.

tests/codeface-data/results/testing/test_empty_feature/feature/authors.list
```diff
@@ -0,0 +1,9 @@
+4936;"Thomas";"thomas@example.org"
+4937;"Olaf";"olaf@example.org"
+4938;"Björn";"bjoern@example.org"
+4939;"udo";"udo@example.org"
+4940;"Fritz fritz@example.org";"asd@sample.org"
+4941;"georg";"heinz@example.org"
+4942;"Hans";"hans1@example.org"
+4943;"Karl";"karl@example.org"
+4944;"Max";"max@example.org"
```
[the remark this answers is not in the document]
nlschn: I tried this at first, the read fails with "Error: Stopped due to missing authors."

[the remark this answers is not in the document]
bockthom: Ah, now I understand the problem here. Thanks for posting the error message. That's indeed not that easy to deal with:

On the one hand, why should there be authors if nobody has committed anything or mailed anything? We, indeed, have a function `create.empty.authors.list()` which would be useless if we wouldn't allow empty authors lists...

However, on the other hand, I know what the reason for this check was: If we have at least some data, we should always have author data: The author data are the only data source which we can use to check whether anything is wrong with the data, assuming that there are data in at least one of the different data sources, resulting also in author data being non-empty.

@hechtlC: We have two options: (1) Keep the check as it is and disallow empty author data as there always must be author data. Or (2) enable empty author data (which would make the empty test data consistent and also increases the value of function `create.empty.authors.list()` but may lead to a problem when author data is not present where it should be present. Let's postpone this discussion to our next meeting.

tests/test-data.R
```diff
@@ -105,6 +105,83 @@ test_that("Compare two ProjectData objects on empty data", {
     expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
 })
 
+test_that("Compare two ProjectData objects on non-empty data", {
+
+    ## initialize a ProjectData object with the ProjectConf and clone it into another one
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
```
bockthom, line 111: Just a quick question: Is the new test on non-empty data a duplication of the test on empty data with just using `CASESTUDY` instead of `CASESTUDY_EMPTY` and adding some configuration updates regarding locking? If these are the only differences, I would like to use a parameterized test instead of having two tests which are almost duplicates.

For parameterized tests, we use `patrick::with_parameters_test_that` and specify different cases afterwards via `patrick::cases(...)` and specifying a list of variables and their values in the respective case. In our test, we should have two different cases: For the one case, we set `test.casestudy` to `CASESTUDY`, and for the other case, we set `test.casestudy` to `CASESTUDY_EMPTY` and use parameter `test.casestudy` in the test. In addition, we need a second parameter `test.locking` which is `TRUE` for the non-empty casestudy and could be either true or false in the empty-case study –– thus, we can easily have a third case to test locking on empty data... –– and we just need to put all the locking-related Conf updates into an `if (test.locking) {...}`. With that, we should be able to reduce the amount of duplicated test code. (For an example on how to use `patrick`, you can have a look at [`test-networks-equal-constructions.R` (lines 90 and 140)](https://github.com/se-sic/coronet/blob/a7f4123dcb86a7a0f76adf791b66a7d9c24f7647/tests/test-networks-equal-constructions.R#L90-L140), I'd guess that should help).

NEWS.md
```diff
@@ -19,14 +19,18 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
-
+- Add functionality that allows to read any data source at any point in time, even after splitting. In this case, the read data is automatically cut to the corresponding range on the `RangeData` object (PR #201, 7f9394f528068f52957be67ab7e2efb733011c6b). Additionally, changing the configuration parameters concerning additional data sources, the environment of a `ProjectData` object is no longer reset (PR #201, 364b3a255d2124d1e07cd6cfa295209dcf3806f8).
+- Add new configuration parameters `commits.locked`, `mails.locked` and `issues.locked` to ProjectConf which, when set to `TRUE`, prevent the respective getters from triggering the read of the data if it is not present yet (PR #201, eed45ac1f97b4915ec42fb869f34359da1df4c60).
```
bockthom, line 23: Just a few additions: Commit 364b3a2 still is outdated.
And commit eed45ac is, to the best of my knowledge, not related to the locking stuff. Please update them correctly...

NEWS.md
```diff
@@ -19,14 +19,18 @@ and their respective centrality values (d3cd528609480f87658601ef13326e950a74cce7
...
 - Add R version 4.1 to test suite and adjust missing time-zone attributes on `NA` vectors or empty POSIXct vectors which are correctly added as of R version 4.1 (PR #203, 6b7fb36478325185f80f95e03dd0a0135bf44346, 98c56715502cf5140d98796dc2d6cb18f71760b2, 09d11ab631de8051ca317f36652ab0fee9cc0cce)
+- Splitting no longer loads all (additional) data sources, but only the ones that have already been cached in the 'ProjectData' (PR #201, 52a3014b3c4a1d6dda1347adcffed4e90fae1d2b, aec898e105ff5478fb75ba488b56c9c258a17a80, de1bbfe644c52200e80a769e7c0e447c3a087278).
+- Change the internal representation of empty data from `NULL` to empty data frames and adapt function `get.cached.data.sources()` of ProjectData which returns a vector of all data sources that are cached (including additional and filtered data sources) (PR #201 aec898e105ff5478fb75ba488b56c9c258a17a80, e55d088f926ae7a069761864855c179c04ed07dc); additionally, introduce new function `is.data.source.cached()` in `util-data.R` that returns a logical vector indicating which of the given data sources are cached (PR #201, b49cc5df4da04af9df2ae9bcdc7847fafb6befc9).
```
bockthom, line 33: When talking about `is.data.source.cached`, please not only reference the commit in which the usages are introduced (b49cc5df), but please also add the commit in which the function itself was added:
491e70c44dd71e4d51b40ee83bc8eb598d413bc0

bockthom (changes requested):

NEWS.md
nlschn: Must have pasted the hash at the wrong location. So many hashes...

tests/test-data.R
nlschn: > Just a quick question: Is the new test on non-empty data a duplication of the test on empty data with just using CASESTUDY instead of CASESTUDY_EMPTY and adding some configuration updates regarding locking?

No, some of the checks were changed from expect_true to expect_false

bockthom: Ok, but then the info "Two identical ProjectData objects." seems to be wrong for some of the checks...

bockthom: And also the comment in line 120 "It is always equal" needs to be adjusted then, too.

tests/README.md
```diff
@@ -0,0 +1,22 @@
+<link rel="shortcut icon" type="image/png" href="../logo/3.favicon_radius.png">
+
+![coronet](./logo/5.coronet_with_font_radiused_rectangle.png)
```
bockthom, line 3: This (line 3) does, on the one hand, not work, and on the other hand, is not necessary:

It is enough to have the favicon, we don't need to have the coronet logo displayed at the beginning of this test-README. Instead, I would suggest to have a title:

`# Overview of the Test Projects`

tests/README.md
```diff
@@ -0,0 +1,22 @@
...
+     * Authors
+     * Commit messages 
+     * Pasta
+     * Synchronicity
+2. - Casestudy: `test_empty`
+   - Selection process: `testing`
+   - Contains the following data:
+     * Authors
+
+Please note, that there cannot be a project without author data as in this case, coronet stops when reading the data. Everything else can be empty.
```
bockthom, line 22: Please highlight `coronet` :wink:

tests/test-data.R
```diff
@@ -90,10 +90,94 @@ test_that("Compare two ProjectData objects", {
...
+    proj.data.two = proj.data.one$clone()
+
+    expect_true(proj.data.one$equals(proj.data.two), info = "Two identical ProjectData objects.")
+
+    ## always change one data source in the one object, test for equality, change it in the
+    ## second object, as well, and test for equality.
+
+    ## test 'equals' on pasta
+    proj.data.two$get.pasta()
+    expect_false(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
```
bockthom, line 124: Shouldn't this be "non-identical" ?

tests/test-data.R
```diff
@@ -90,10 +90,94 @@ test_that("Compare two ProjectData objects", {
...
+
+    proj.data.one$get.pasta()
+    expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
+
+    ## test 'equals' on commits
+    proj.data.one$get.commits()
+
+    ## prevent 'equals' from triggering a read when calling the getter
+    proj.conf$update.value("commits.locked", TRUE)
+    expect_false(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
```
bockthom, line 134: Shouldn't this be "non-identical" ?

tests/test-data.R
```diff
@@ -90,10 +90,94 @@ test_that("Compare two ProjectData objects", {
...
 
     proj.data.two$get.synchronicity()
+    expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
+
+    ## test 'equals' on commit messages
+    proj.data.one$get.commit.messages()
+    expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
 
+    proj.data.two$get.commit.messages()
     expect_true(proj.data.one$equals(proj.data.two), "Two identical ProjectData objects.")
```
bockthom, line 181: Hm, this confuses me a bit..., I'm not sure if this tests what it is intended to test...
Either let's just state in the comments that both "synchronicity" and "commit messages" are not configured (and therefore we always expect identical objects here), or let's add additional checks actually using synchronicity data and commit messages (then we would need to configure "synchronicity", "synchronicity.time.window", and "commit.messages" in the ProjectConf right before adding such additional checks...)

bockthom (changes requested):
Almost done: Just some minor adjustments in the new test-README and the adjustment of one test left :smile:

tests/test-data.R
nlschn: I fixed that in both equality tests. This caused the `read.synchronicity` function to be called in the empty data test. Turns out that loading an empty file with `load(file = file)` throws an error, so I added an emptyness check besides the one checking if the file exists in the first place. 
All other functions use some try-catch-like construct that, I think, circumvents this issue, I don't know, why `read.synchronicity` doesn't use that.

bockthom: Thanks for adjusting the test descriptions. Unfortunately, I have spotted two minor inconsistencies:
- In line 45, you also should mention "(clone)" as you added it also to the corresponding line in the second test.
- In line 149, the closing bracket is missing "(authors".

In addition, could you try to add further checks to ensure that the data objects are different, for instance, when using another type of "commit.messages"? Currently you use "message", but what happens when using anything else than that? (Sorry that I did not think about that before, but let's try out what happens when "message" is changed to something else – I am not sure what happens then with the commit data or commit message data... Same for changing the synchronicity time window...We don't need to have all of these changes in the test, but we should try it out though manually).

Last but not least: The synchronicity data is a binary R file, so we cannot read a csv file there, but need to load a binary file containing R data frames instead. This is why this is the only situation where we call the `load` function. As loading could fail anyway when the loaded binary does not contain the expected content, we did not have checks for emptiness. But it is good have such a check now.

nlschn: `set.commits()` will not trigger `update.commit.messages()` if `commit.messages` are cached, meaning that changing the config parameter does, at the moment, not affect the commits, even when calling `get.commits()` even after changing the config. Is this intended or should I call the update function every time the setter is called (given the parameter is not set to `none`)?

bockthom: Hm, good question. We somehow need to handle that, but I don't know how...
Should we automatically take care of that whenever we update the ProjConf? That is, whenever we update the "commit.message" option, then the corresponding update function should be triggered (and also for the other conf parameters) - but we need to make sure that updating only affects the cached data but does not read new data...

@hechtlC Any ideas? Would you agree with calling the update functions when the corresponding parameters are changed?

nlschn: I have accidentally deleted the question, sorry
But how I handled it now, is, I reset the commits of one data object manually and re-get the commits

hechtlC: Hmm that's a tricky one. But I suppose that since you want to work with the additional data if you configure it later, we could call the update functions immediately. But like @bockthom said, I would not want to change the already cached commit (etc.) data, as it could be that it is modified already and you would not want to change the modification. That is except the adding of the new data source. So yes I agree @bockthom.

bockthom: > I have accidentally deleted the question, sorry.

Thanks to e-mails, I can re-post your question again, here it is:

> `set.commits()` will not trigger `update.commit.messages()` if commit.messages are cached, meaning that changing the config parameter does, at the moment, not affect the commits, even when calling `get.commits()` even after changing the config. Is this intended or should I call the update function every time the setter is called (given the parameter is not set to `none)?

So, this was the question to which I responded with "Hm good question..." above...

bockthom: I just had a quick look at your most recent commit on that. Let me add one more comment on that:
We should also make sure that the update function is only called once (or, at least, as less frequent as possible), for performance reasons. For instance, in 67e11f4 you adjust the setter function `set.commits` to always call the update function. I am not sure whether we really need to adjust the setter, but, anyway, as the setter might call `get.commit.messages`, which in turn might call the update function, the update function would be called twice from the setter in some of the cases. I would like to avoid that.
But I also have seen now that this problem is not trivial :cry: So what we actually need is (1) to find out in which cases the update function really needs to be called and (2) optimize such that we omit repeated update calls and though get correct results in every situation...

nlschn: I think I just forgot to turn it back, but I don't think calling the update function in the setter everytime is not necessarry _anymore_ since I added a call in `set.project.conf.entry` if the parameter is changed.

util-data.R
```diff
@@ -745,6 +755,16 @@ ProjectData = R6::R6Class("ProjectData",
                 if(c("commit.messages") %in% params) {
```
bockthom, line 755: Missing space after `if`.

util-data.R
```diff
@@ -720,6 +720,16 @@ ProjectData = R6::R6Class("ProjectData",
...
+            ## if the 'pasta' parameter has been changed, update the pasta data, since we want to
+            ## merge the data again if the parameter changed
+            if (entry == "pasta") {
+                private$update.pasta.data()
+            }
+            ## if the 'synchronicity' or 'synchronicity.time.window' parameter has been changed, update the
+            ## synchronicity data, since we want to mergee the data again if the parameter changed
+            if (entry == "synhronicity" || entry == "synchronicity.time.window") {
+                private$update.synchronicity.data()
+            }
```
bockthom, line 732: First of all, don't worry about the failing test-pipeline in R version 3.3: It seems that package `sqldf` has stopped supporting R version 3.3. (I'll check that in detail later, but I'd assume that we will also need to stop supporting R 3.3 then...)

But let me add a few other thoughts:
- "synhronicity" contains a typo (also below at the other line where you call the update function) (and also the word "mergee" is misspelled in both comments :wink:)
- I am not sure if the update function for pasta and synchronicity works correctly when setting the corresponding parameters `FALSE`. (Remember, for the commit messages, you remove the columns if the parameter is changed, but what do you do when pasta or synchronicity are set to `FALSE`? Do you also remove the columns then, or what else happens in the update function then?
- I'd only call the update functions if the environment was *not* reset (otherwise there is nothing to update). You already implemented it that way in the `update.project.conf` function below, but not in the `set.project.conf.entry` function.

Please make sure that everything mentioned above works as expected. If so, then we either merge although the pipeline fails for R 3.3. or I will send you a patch which you can apply before merging (depends on what I find out regarding sqldf in R 3.3.).

nlschn: > First of all, don't worry

Okay, thanks for taht information, I thought I broke something again...

> Do you also remove the columns then, or what else happens in the update function then?

I assumed that would be the case, but I can implement this

> You already implemented it that way in the update.project.conf function below, but not in the set.project.conf.entry function.

Yes, I meant to implement it the same way, thanks for pointing that out.

nlschn: > I am not sure if the update function for pasta and synchronicity works correctly when setting the corresponding parameters `FALSE`

When looking at the `update.commit.messages` function, I believe setting it to `none` doesn't work either. An additional check is needed there as well

bockthom: Good point! Thanks for also checking that!

I had a quick look at your changes, and, unfortunately, identified to further issues in your last commit:
(1) Please don't use `== TRUE` checks! Just remove `== TRUE` and everything should be fine. You have added several of these checks...
(2) When changing the synchronicity time window, new data has to be read, since we have separate input data for different synchronicity time windows. However, our current implementation just resets the corresponding columns in the commit data, but the private field `synchronicity` is still the old one. Hence, when one changes the synchronicity time window from 5 to 10, one would get still the values from 5. The easiest way would be to reset the environment when the synchronicity time window parameter is changed. However, maybe you find also a way to just remove and trigger re-reading the synchronicity data in such a case? Could you please check for a couple of minutes whether this is possible to achieve? If not, I would go for resetting the environment (to just come to an end here is synchronicity data is not used that often...)

I'm really sorry for spotting new issues again and again :see_no_evil: (I'm not blaming on you, it is just my fault to not think about these issues earlier) I'm really hoping that this synchronicity related issue is the last one I've spotted, maybe I just should stop thinking about this PR after this one is fixed...

bockthom: And, luckily, even the pipeline for R 3.3 seems to work again :partying_face: There may have been a problem with the availability of certain package versions. Hopefully this does not happen again until this PR will be merged so that I don't have to deal with this problem any more...

bockthom: > However, maybe you find also a way to just remove and trigger re-reading the synchronicity data in such a case?

Additional comment: This should only happen if the synchronicity data is already cached. If one changes the time window but has not read the synchronicity data yet, we should definitively not read it.

nlschn: I added another check to the functions that change config parameters (they have gotten a bit lengthy...) that resets and reloads only the synchronicity if 1. the time window changed and 2. synchronicity data is alreay cached.

bockthom (approved):

hechtlC (approved):

PR #203, opened 2021-05-22
Necessary adjustments for the recently released R version 4.1

bockthom:
### Prerequisites

### Description

A few days ago, R version 4.1.0 was released. The new R version brings a few adjustments and fixes, which in turn involve a few minor adjustments to let our test suite run successfully even with the new R version.

In particular, there are two fixes in R 4.1.0 that break our test suite:
- The base environment `baseenv()` is locked now. That is, we cannot clear the variable `last.warning` any more at the beginning of the test suite as `last.warning` belongs to the base environment. However, we can just remove the corresponding statement in our test suite as we ignore warnings anyway. (98c5671)
- In previous R versions, empty POSIXct vectors (e.g., `c()`) as well as `NA` vectors of class POSIXct did not have an `tzone` attribute (which was actually a bug in R). In the new R version, this has been fixed. As we did not take care of that `tzone` attribute when creating empty POSIXct vectors or specifying `NA` as a default value for a date (as R itself did not take care of either), some tests fail when comparing dates containing an `NA` value. This can be easily fixed by making sure that the `tzone` is set correctly when creating empty data frames, and also by calling `get.date.from.string` when specifying `NA` as the default value of a date (e.g., when computing vertex attribute "first activity"). (09d11ab)

Last but not least, consistently add R version 4.1 (as of now, 4.1.0) to our CI test pipeline and update the patch release of R version 4.0 to its most recent patch release 4.0.5 (which is the last patch release of R 4.0). (6b7fb36)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,7 +11,7 @@
 ## with this program; if not, write to the Free Software Foundation, Inc.,
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
-## Copyright 2020 by Thomas Bock <bockthom@cs.uni-saarland.de>
+## Copyright 2020-2021 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## All Rights Reserved.
 
 # Drone CI Setup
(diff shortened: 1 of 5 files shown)


hechtlC (approved):
Thanks for the fixes @bockthom ! I looked over the things and didn't find anything wrong with it. This PR can be merged.

Issue #202, opened 2021-05-14
Read bot data and filter bots


bockthom:
### Motivation &  Description

In many projects there are bots that participate in the development process. However, such bots could be highly active and can distort the network structure and corresponding analyses.

To mitigate this, we need to identify bots (which we do for issue data by an already existing tool) and provide a possibility to filter such bots in coronet. When assuming that we already have identified the bots and stored them in a separate `.list` file, we need to somehow read the bot data and filter the bots from the authors.

### Ideas for the Implementation

Today, we had a discussion with @JoJoDeveloping and @nlschn on how we could implement this. We ended up with the following solution:

1. **Read bot data when reading author data:** Within `get.authors()`, check if a bot list is present, and if so, read the bot data and merge it with the author data, that is, add an additional column (e.g. `is.bot`) where bots get `TRUE`, humans `FALSE`, and authors that don't appear in the list of bots `NA`. We decided to **not** make this configurable and always read bots if they are present as the additional amount of data is rather small but the benefit of knowing about who is a bot can always be accessed without any configuration to keep configuration options manageable. If no bot data is present, the new implementation does behave as the previous one, but we can print a warning that no bots are available.
*Advantage:* Whenever we read the author data, we actually immediately know about which authors are bots. We don't need to store this information in a separate data source, as it is just one column of additional information, so that we can directly merge the bot data with the author data and store it within the author data.
*Disadvantage:* If no bot data is available, we don't add a `is.bot` column, therefore having different columns in the author data when having bots or not having bots. However, we want to avoid having a column consisting of only `NA` values to reduce memory consumption––and this behavior is similar to adding commit messages as this also enhances the commit data with additional columns which are not present otherwise.

2. **Filter bots when configured:** We need to add a configuration option `filter.bots` (or similarly named) to the ProjectConf. When ever this option is enabled, the bots should be filtered from all data sources within `filter.commits`, `filter.issues`, and `filter.mails` to remove bots from all data sources where needed when calling the corresponding getters (e.g., `get.issues.filtered`). We agreed that all the filter functions should call `get.authors` to trigger reading author and bot data and use the information from there to remove actions triggered by a bot.
*Advantage:* We can filter a bot's actions from all data and the bots are also removed before creating networks on the data. Even if we currently have only bot data for GitHub issues, we can also remove their commits or mails (if there are such). Our implementation is also prepared for future enhancements where we could detect commit bots or mail bots.
[...]

PR #206, opened 2021-06-10
Read bot data and filter bots

JoJoDeveloping:

### Description

Aims to implement #202

### Changelog

* Added dummy test file
* Added code to read bot data and merge it into the authors data
* Implement filtering for issues and commits
* Add tests
* Fix some bugs
* Rename main data source getters and internal fields
* Remove implicit reflective field accesses in favor of explicit lists outlining the accessed fields


util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+## column names of a data frame containing bot information (see file
+## 'TODO-bots.list' and function \code{read.bot.list})
+## TODO figure out what the actual columns will be
+BOT.LIST.COLUMNS = c(
+    "author.name", "author.email", ## author
+    "comment.count", "comment.empty.count", "comment.pattern.count", "comment.dispersion", ## intermediate results
+    "is.bot" ## whether this is a bot
+)
+
+#' Read the bot classification from the 'TODO-bots.list' file.
```
bockthom, line 400: Please don't forget to update the name.

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+#'
+#' @param data.path the path to the commit-messages list
+#'
+#' @return a data frame with author.name, author.email, and a (potentially NA) boolean whether this is a bot,
+#'         or \code{NULL} if the above file is not present.
+read.bot.info = function(data.path) {
+    logging::logdebug("read.bot.info: starting.")
+
+    ## read the file with the commit messages
+    file = file.path(data.path, "TODO-bots.list")
```
bockthom, line 410: Update file name.

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+#' Read the bot classification from the 'TODO-bots.list' file.
+#'
+#' @param data.path the path to the commit-messages list
+#'
+#' @return a data frame with author.name, author.email, and a (potentially NA) boolean whether this is a bot,
+#'         or \code{NULL} if the above file is not present.
+read.bot.info = function(data.path) {
+    logging::logdebug("read.bot.info: starting.")
+
+    ## read the file with the commit messages
```
bockthom, line 409: No, we don't read commit messages here...

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+    file = file.path(data.path, "TODO-bots.list")
+
+    bot.data = try(read.table(file, header = FALSE, sep = ";", strip.white = TRUE,
+                                         encoding = "UTF-8"), silent = TRUE)
+
+    ## handle the case that the list of commits is empty
+    if (inherits(bot.data, "try-error")) {
+        logging::logwarn("There is no bot clarification available for the current environment.")
```
bockthom, line 417: What do you mean by "bot clarification" ? I don't get that. Maybe you find a noun that suits a little bit better.

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+read.bot.info = function(data.path) {
+    logging::logdebug("read.bot.info: starting.")
+
+    ## read the file with the commit messages
+    file = file.path(data.path, "TODO-bots.list")
+
+    bot.data = try(read.table(file, header = FALSE, sep = ";", strip.white = TRUE,
+                                         encoding = "UTF-8"), silent = TRUE)
+
+    ## handle the case that the list of commits is empty
```
bockthom, line 415: Nope. We don't care if the list of commits is empty here :wink:

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+        logging::logwarn("Datapath: %s", data.path)
+
+        ## return a data frame with the correct columns but zero rows
+        return(NULL)
+    }
+
+    ## set column names for new data frame
+    colnames(bot.data) = BOT.LIST.COLUMNS
+
+    bot.data = bot.data[, c("author.name", "author.email", "is.bot")]
```
bockthom, line 427: Maybe this is not necessary any more after the update of the bots.list. Please check. If you think it is still useful, we can also keep it.

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+    bot.data["is.bot"] = sapply(bot.data[["is.bot"]], function(x) switch(x, Bot = TRUE, Human = FALSE, NA))
+
+
+    logging::logdebug("read.bot.info: finished.")
+
+    return(bot.data)
```
bockthom, line 433: Why do we have so much blank lines here? One blank line before the logging statement should be enough...

util-read.R
```diff
@@ -423,6 +473,17 @@ read.authors = function(data.path) {
     }
     colnames(authors.df) = AUTHORS.LIST.COLUMNS
 
+    bot.data = read.bot.info(data.path)
+    if (!is.null(bot.data)) {
+        ## TODO keep row names consistent
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]),] # re-order after read
+        row.names(authors.df) = 1:nrow(authors.df)
```
bockthom, line 481: Please address this TODO, (even if I don't know what the TODO comment actually means...)

util-read.R
```diff
@@ -385,6 +386,54 @@ create.empty.issues.list = function() {
...
+
+
+    logging::logdebug("read.bot.info: finished.")
+
+    return(bot.data)
+}
+
+
 ## column names of a dataframe containing authors (see file 'authors.list' and function \code{read.authors})
 AUTHORS.LIST.COLUMNS = c(
```
bockthom, line 438: Shouldn't we add the `is.bot` column here? I am not sure about that, please check. Maybe there are cases where this column is not existent, then we may not need it? 
@hechtlC Please have a closer look at this and how we treat that at other column lists...

util-read.R
```diff
@@ -423,6 +473,17 @@ read.authors = function(data.path) {
...
+    bot.data = read.bot.info(data.path)
+    if (!is.null(bot.data)) {
+        ## TODO keep row names consistent
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]),] # re-order after read
+        row.names(authors.df) = 1:nrow(authors.df)
+    } else {
+        authors.df[["is.bot"]] = NA
+    }
+    authors.df = authors.df[,c("author.id", "author.name", "author.email", "is.bot")]
```
bockthom, line 485: Do we really need this?

util-conf.R
```diff
@@ -378,6 +379,12 @@ ProjectConf = R6::R6Class("ProjectConf", inherit = Conf,
                 allowed = c(TRUE, FALSE),
                 allowed.number = 1
             ),
+            filter.bots = list(
+                default = FALSE,
+                type = "logical",
+                allowed = c(TRUE, FALSE),
+                allowed.number = 1
+            ),
```
bockthom, line 387: Please don't forget to add this one in the list of configuration options in the README.md file.

bockthom (changes requested):
Hey, nice work @JoJoDeveloping. Looks already good, but I also have a few comments on that (see also my detailed comments below).

Independent from that, I would like to mention a few major issues which need some further investigations or discussions:
- What happens to unfiltered data when adding data from additional data sources? For example: If I enable PaStA, will the PaStA data be merged to just the filtered commit/mail data or also the unfiltered data? (It looks like that the current implementation in this PR only adds it to the filtered data, but I am actually not sure and unfortunately I don't have the time any more to check that. Please also then have a look at the update functions for all the additional data sources and check whether everything works well for both filtered and unfiltered cached data...)
- Can we add an example on how to filter bots manually by calling the corresponding function in the showcase file?
- Shouldn't we add a test case to test `read.bot.info`? It is already tested within the test for reading the author data. But a separate test might be useful just in the case that the procedure on how reading author calls reading the bot information will change. Then it would be useful to test reading the bot information also separately.
- I am not sure whether the default for bot filtering should be `TRUE` or `FALSE`. Please discuss with @hechtlC.

util-data.R
```diff
@@ -478,21 +534,21 @@ ProjectData = R6::R6Class("ProjectData",
             if (self$is.data.source.cached("mails")) {
 
                 ## remove previous PaStA data
-                private$mails["pasta"] = NULL
-                private$mails["revision.set.id"] = NULL
+                private$mails.filtered["pasta"] = NULL
+                private$mails.filtered["revision.set.id"] = NULL
```
hechtlC, line 539: If we only do this for the filtered mails (which I am not quite sure if we decided to do this can you remind me if we decided so?) then you have to change the comment of the function in line 528 from \code{mails} to \code{mails.filtered}

util-data.R
```diff
@@ -854,48 +911,50 @@ ProjectData = R6::R6Class("ProjectData",
...
+        #' Return the commits retrieved by the method \code{get.commits.unfiltered} by removing untracked files and removing the
         #' base artifact (see parameters).
         #'
         #' This method does not use caching. If you want to use caching, please use the method
-        #' \code{get.commits.filtered} instead.
+        #' \code{get.commits} instead.
         #'
         #' @param remove.untracked.files flag whether untracked files are kept or removed
         #' @param remove.base.artifact flag whether the base artifact is kept or removed
+        #' @param filter.bots flag commits by bots should be removed
```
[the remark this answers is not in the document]
hechtlC: There is also a 'whether' missing here

util-data.R
```diff
@@ -1543,9 +1635,28 @@ ProjectData = R6::R6Class("ProjectData",
...
+                "mails"   = "get.mails.unfiltered",
+                "issues"  = "get.issues.unfiltered"
+            )
+
+            DATASOURCE.TO.ADDITIONAL.ARTIFACT.FUNCTION = list(
+                "authors" = "get.authors",
+                "commit.messages" = "get.commit.messages",
+                "synchronicity" = "get.synchronicity",
+                "pasta" = "get.pasta"
+            )
```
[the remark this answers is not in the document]
hechtlC: And if not then please align at the equals sign.

util-read.R
hechtlC: I guess that really depends on whether we always read the bot data in addition to the author data and merge them or not. If so then yes we should add it. If not then leave it out (just like we do with the pasta columns in mail and commit data)

hechtlC (dismissed):
Nice work @JoJoDeveloping 
I have only found a couple more things in addition to the comments by @bockthom.

One thing I am not really sure about is: In the function `update.pasta.data()`, we check if mails/commits are in the cached data sources. But the functions `update.pasta.commit.data()` and `update.pasta.mail.data` are then working with the filtered data sources. So should this check not rather check for the filtered mail and commit data to be there? Or am I missing something?

Please also don't forget to add you copyright header in `util-networks.R`.

And also please don't forget to rebase your PR to the current dev branch as there are changes.

util-data.R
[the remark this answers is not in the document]
hechtlC: So we discussed this and we think (even though the base artifact filtering is always TRUE as default) that it would be best to have FALSE as default for the bot filtering.
But we can discuss this further after our vacation @bockthom

util-read.R
JoJoDeveloping: We do not, since this global list describes the file format, and the bot data is not part of the `authors.list` file

JoJoDeveloping:
The latest commits should have adressed all comments, change request, remainibg TODOs.

util-data.R
```diff
@@ -486,7 +544,7 @@ ProjectData = R6::R6Class("ProjectData",
                 if (private$project.conf$get.value("pasta")) {
                     ## merge PaStA data
                     private$mails = merge(private$mails, private$pasta.mails,
-                                          by = "message.id", all.x = TRUE, sort = FALSE)
+                                                   by = "message.id", all.x = TRUE, sort = FALSE)
```
bockthom, line 547: Wrong indentation. Previously it was correct.

util-data.R
```diff
@@ -475,7 +533,7 @@ ProjectData = R6::R6Class("ProjectData",
             logging::logdebug("update.pasta.mail.data: starting.")
 
             ## return immediately if no mails available
-            if (self$is.data.source.cached("mails")) {
+            if (self$is.data.source.cached("mails.unfiltered")) {
```
bockthom, line 536: We actually need to do this twice: Once for "mails.unfiltered" and once for "mails". The content of this if-block contains the "mails" variant (as opposed to your change here). So please copy the whole block and adjust it correctly for the other variant.
For consistency reasons, the first one should be "mails.unfiltered" and the second one "mails.filtered".
This comment may sound unintuitive. In case of questions, don't hesitate to ask or step by.

util-data.R
```diff
@@ -604,15 +662,15 @@ ProjectData = R6::R6Class("ProjectData",
...
+                self[[ DATASOURCE.TO.UNFILTERED.ARTIFACT.FUNCTION[[source]] ]]()
             }
         },
 
         #' Extract the earliest and the latest date from the specified data source
         #' and store it to the timestamps data.frame.
         #'
         #' @param source the specified data source
-        extract.timestamps = function(source) {
+        extract.timestamps = function(name, source) {
```
bockthom, line 673: The new "name" parameter is not documented.
When adjusting the documentation, please clearly state what the difference between "name" and "source" is and which of the two parameters is used for what

util-data.R
```diff
@@ -683,12 +741,13 @@ ProjectData = R6::R6Class("ProjectData",
...
         reset.environment = function() {
             private$authors = create.empty.authors.list()
+            private$commits.unfiltered = create.empty.commits.list()
             private$commits = create.empty.commits.list()
-            private$commits.filtered = create.empty.commits.list()
             private$commit.messages = create.empty.commit.message.list()
             private$data.timestamps = data.frame(start = numeric(0), end = numeric(0))
-            private$issues.filtered = create.empty.issues.list()
             private$issues = create.empty.issues.list()
+            private$issues.unfiltered = create.empty.issues.list()
```
bockthom, line 749: Please swap these two lines for consistency reasons.

util-data.R
```diff
@@ -1227,8 +1288,20 @@ ProjectData = R6::R6Class("ProjectData",
...
+            return(private$mails.unfiltered)
+        },
+
+        #' Get the mail data.
+        #' If it does not already exist call the read method.
+        #' Call the setter function to set the data and add PaStA
+        #' data if configured in the field \code{project.conf}.
+        #'
+        #' @return the mail data
+        get.mails = function() {
```
bockthom, line 1302: This is just copy-paste from the unfiltered getter. Please add some information that this gets the filtered mail data (in a similar way as for the getter of unifiltered and filtered commits...)

util-read.R
```diff
@@ -423,6 +466,16 @@ read.authors = function(data.path) {
...
 
+    bot.data = read.bot.info(data.path)
+    if (!is.null(bot.data)) {
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]),] # re-order after read
+        row.names(authors.df) = 1:nrow(authors.df)
+    } else {
+        authors.df[["is.bot"]] = NA
+    }
+    authors.df = authors.df[,c("author.id", "author.name", "author.email", "is.bot")]
```
bockthom, line 477: Please add a comment to this line that states that we need this line to sort columns correctly :smile: 
And: Missing space in front of and after the comma

util-read.R
```diff
@@ -423,6 +466,16 @@ read.authors = function(data.path) {
     }
     colnames(authors.df) = AUTHORS.LIST.COLUMNS
 
+    bot.data = read.bot.info(data.path)
+    if (!is.null(bot.data)) {
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]),] # re-order after read
```
bockthom, line 472: Missing space after the comma

bockthom (changes requested):
Thanks @JoJoDeveloping  for addressing our reviews.
@hechtlC and I today had a joint look over your PR again, and––unfortunately––we spotted a bunch of problems that need to be fixed before we can merge this PR.

First of all, here are some general issues to be fixed:
- [x] Copy `test_feature/feature/bots.list` in the test data also to `test_proximity/proximity/bots.list`
- [ ] Copy `test_feature/feature/issues-github.list` also to `test_proximity/proximity/issues-github.list`
- [ ] Add an example of using the function `filter.bots(...)` to the `showcase.R` (Here we explicitly mean the function `filter.bots(...)`, not the parameter (for the parameter, see the individual comments below.)
- [ ] In the `README.md`, there is a section "Folder structure of the input data". In this section, there is an overview of all `.list` files in alphabetical order. Please add the `bots.list` file there.
- [ ] Add `"is.bot"` to `AUTHOR.LIST.COLUMN` in `util-read.R` (see more details in separate comment)

In addition, please also have a look at the comments below. Don't hesitate to ask or step by in case of questions.

util-read.R
bockthom: > We do not, since this global list describes the file format, and the bot data is not part of the `authors.list` file

No, that is wrong. For issue data, there is an "event.id" in this global list whereas the event id is not part of the `issues.list` file. Or in the mail data, we have an "artefact.type" column mentioned in the global list, but the "artefact.type" is not part of the "emails.list". For that reason, we should also add the `is.bot` column to the `AUTHORS.LIST.COLUMNS` as this column is always present (regardless of having a `bots.list` file or not). This is different to pasta related columns in the commit or mail data, which are only present when configured. So, lets add `is.bot` as it is always present and also for consistency reasons. @hechtlC also agreed on this.

util-data.R
```diff
@@ -1486,6 +1584,38 @@ ProjectData = R6::R6Class("ProjectData",
...
+            sources = self$get.cached.data.sources(source.type = "only.additional")
+            additional.to.external.name = list(
+                "authors" = "authors",
+                "commit.messages" = "commit.messages",
+                "synchronicity" = "synchronicity",
+                "pasta" = "pasta"
+            )
+            return(additional.to.external.name[sources])
+        },
+
```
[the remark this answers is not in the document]
JoJoDeveloping: This one has indeed be tricky.

In my opinion, `get.cached.data.sources` should be private. It is highly dependant on the names of private fields and thus its behavior should not be relied on by other, external functions. The reason it is not private is that it is invoked on the "`other`" argument of `ProjectData::equals`. In sane object-oriented languages, this would not be an issue since one is usually able to invoke private member functions on other objects of one's own type. In R, however, this is not the case, hence this function is public.

There have been a few other places where this function was used, which led to issues during the field renaming process, because again, this function exposes the names of private fields to classes that should not concern themselves with that. (This functionality, however, is useful inside the class since a lot of other methods in `util-data` rely on this duality and I dod not want to rename them all). Thus I added these helper functions, which can be used to ensure external behavior stays consistent when one changes implementation details (which is why I explicitly spell out the possible values, so that future programmers are able to quickly find and fix issues).

Your proposed changes would affect all callsites of `is.data.source.cached`, and I would rather avoid going over all of them.

Ideally, we should mark this function as "Do Not Use".

tests/codeface-data/results/testing/test_feature/feature/bots.list
```diff
@@ -0,0 +1,4 @@
+"Thomas";"thomas@example.org";Bot
+"Björn";"bjoern@example.org";Human
+"udo";"udo@example.org";Unknown
+"CLAassistant";"nomail";Bot
```
bockthom, line 4: Can you please put the last column into quotation marks? e.g. "Bot"
Holds for both "bots.list" files (feature and proximity).

NEWS.md
```diff
@@ -21,6 +21,8 @@
...
 - Add new configuration parameters `commits.locked`, `mails.locked` and `issues.locked` to `ProjectConf` which, when set to `TRUE`, prevent the respective getters from triggering the read of the data if it is not present yet (PR #201, 382167790e7d096b539eeed6464efeb9bb501d43)
 - Add support for classifying developers on the basis of more count-based classification metrics, including mail-count, mail-thread count, issue-count, issue-comment count, issue-commented-in count, and issue-created count (issue #70, PR #209, d7b2455ece9840caba7c3a46704d2a62a11bee57, 6f737c8859519867928acc4e545c89bfd91b1e2e)
+- Add bot filtering mechanism, which allows removing issues/mails/commits made by bots (838855ffbe6d404bcb447d470a97870b238754ae, dcce82de58e25c6a0f3c27c9a7c46fb35a47fced)
+- Ignore the "deleted user" as wel as the author having an empty name "" (1a08140884f30366995352af315218cfd315b958, 24c222a21238cf0d21dfa750d8c951f39b4d0500)
```
bockthom, line 25: There is a typo, should be as we**ll** as

util-read.R
bockthom: The "is.bot" column still is missing in `AUTHORS.LIST.COLUMNS`...

util-read.R
```diff
@@ -423,6 +481,20 @@ read.authors = function(data.path) {
...
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]), ] # re-order after read
+        row.names(authors.df) = 1:nrow(authors.df)
+    } else {
+        ## if bot data is not available, add NA data, which is what would have happened
+        ## if the file was empty
+        authors.df[["is.bot"]] = NA
+    }
+    ## re-order the columns
+    authors.df = authors.df[,c("author.id", "author.name", "author.email", "is.bot")]
```
bockthom, line 495: Here still the space after the first comma is missing

util-read.R
```diff
@@ -423,6 +481,20 @@ read.authors = function(data.path) {
     }
     colnames(authors.df) = AUTHORS.LIST.COLUMNS
 
+    bot.data = read.bot.info(data.path)
+    if (!is.null(bot.data)) {
+        authors.df = merge(authors.df, bot.data, by = c("author.name", "author.email"), all.x = TRUE, sort = FALSE)
+        authors.df = authors.df[order(authors.df[["author.id"]]), ] # re-order after read
+        row.names(authors.df) = 1:nrow(authors.df)
```
bockthom, line 488: Please use `seq_len(nrow(authors.df)` instead of `1:nrow(authors.df)`

util-data.R
bockthom: Please swap these two lines for consistency reasons: For commits and mails, you first mention the unfiltered data and then the main data. Please also adjust the order similarly in lines 153–163, where the fields are defined, to have a consistent order fields everywhere.

util-data.R
```diff
@@ -1437,34 +1572,75 @@ ProjectData = R6::R6Class("ProjectData",
 
         #' Get the names of all data sources that are currently cached in the
         #' ProjectData object. The possible data sources are:
-        #' 'commits', 'mails', 'issues', 'commits.filtered', 'issues.filtered', commit.messages', 'authors',
-        #' 'synchronicity', and 'pasta'.
+        #' 'commits', 'mails', 'issues', for 'only.main' and 'only.unfiltered'
+        #' 'commit.messages', 'authors', 'synchronicity', and 'pasta', for 'only.additional'
         #' 'data.timestamps' are tested implicitly every time as they only contain
         #' the earliest and latest date of one data source.
```
bockthom, line 1578: Does this comment regard "data.timestamps" still make sense for the new function? Please check.

util-data.R
```diff
@@ -1437,34 +1572,75 @@ ProjectData = R6::R6Class("ProjectData",
...
         #'
         #' @param source.type character vector indicating which data sources should be checked for.
         #'                    - 'only.main' filters only main data sources, i.e. commits, mails and issues.
         #'                    - 'only.additional' filters only additional data sources and authors.
-        #'                    - 'only.filtered' filters only filtered data sources (i.e. filtered commits and filtered
-        #'                      mails).
+        #'                    - 'only.unfiltered' filters only unfiltered data sources (i.e. unfiltered commits, mails, issues).
+        #'                    [default: "all"]
```
bockthom, line 1584: There is no "all" here, so how can "all" be the default?

bockthom (commented):
Good work @JoJoDeveloping. Almost done. I just spotted a few very minor issues you may have overlooked in my previous comments and some newly introduced issues. But no worries, I have added them below again.

And please don't forget to apply the patch @hechtlC has sent you via e-mail.

When these issues are fixed, I will merge this PR right away :smile:

bockthom (approved):
Looks good now. And CI pipeline also succeeds. So, I will merge right away.
Thanks for your endurance and your great work @JoJoDeveloping. We are glad to be able to use the new features of coronet soon :smile:

Issue #214, opened 2021-10-14
Filter "deleted user" correctly


bockthom:
### Description

Currently, when we filter the deleted user, we check whether `data["author.name"] != "deleted user"`, but the deleted user can also be spelled with upper-case characters, e.g. "Deleted user" or "Deleted User".

https://github.com/se-sic/coronet/blob/a65602663efbe208e201e39418a42d8151903bbb/util-read.R#L50-L51

To filter the deleted user in all their spelling variants, I suggest to use `tolower["author.name"] != "deleted user"`.
In addition, we could think whether we also filter users whose e-mail is "ghost@github.com". According to what I have seen up to now, this will not be necessary, but maybe it could be helpful for future changes...

### Versions

This affects version `4.0` of coronet.

PR #216, opened 2021-10-27
Add functionality to read and merge gender data to author data

miriyusifli:
### Prerequisites

### Description

The pull request adds functionality to read and merge gender data to the author data

### Changelog

- Add gender related configurations to `util-conf.R` file (ebe4899dad80f37cf2ced20ed1591b9b7fd2f2f4)
- Add function `read.gender` to read the gender data (27bf76288525a75e0f948570f43e16f947874ee3)
- Add the base functions `get.data.path.gender`, `get.gender`, `set.gender`, and `update.gender.data` to manage the gender data (20a4e9013b401361b831af1ecafbddb3710253f3)
- Add test, test data, and sample data to test the `read.gender` function (9c9d2612bb481d0f6ddbf36b5b22d918fa2d0cab)


bockthom (changes requested):
Thanks for your nice additions to `coronet` @miriyusifli! I have had a quick look at your changes and they look good in general. However, I have identified a couple of inconsistencies and minor issues which have to be fixed. Please have a look at my detailed comments below.

In addition, the documentation of your changes in several README files is missing. So, please consider the following comments:
1. `coronet/tests/README.md`: You have added gender data to one of these projects, so please mention this here (directly below bots).
2. `README.md`: Please update the readme regarding the new data source. That is, update section `Data sources` and enhance the list of `additional (orthogonal) data sources` by adding a new list item between "commit messages" and "PaStA", describing the gender data. In addition, there is also a section "Configuration Classes" containing a subsection "ProjectConf", containing a subsection "(Configurable) Data-retrieval-related parameters". Please add the gender parameter in this sub-sub-section (at the same place, i.e., the same order, as in the code, see also the comment below.)

In case of any questions (i.e., if something is unclear to you), please don't hesitate to ask!

util-data.R
```diff
@@ -1249,6 +1282,33 @@ ProjectData = R6::R6Class("ProjectData",
...
+            } else {
+                logging::logwarn("You have not set the ProjectConf parameter 'gender' to 'TRUE'! Ignoring...")
+
+                ## mark gender data as empty
+                self$set.gender(NULL)
+            }
+
+            return(private$gender)
+        },
+
```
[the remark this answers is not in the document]
hechtlC: I totally agree. So @miriyusifli please add a method `cleanup.gender.data` where you remove all lines from the gender data that contain author names that do not appear in the author data.

hechtlC (changes requested):
First of all thanks for the nice changed @miriyusifli.
I have not found any new issues here as of now (except for the failing test which has to be fixed). But I totally agree with @bockthom on the points he raised. So please change this.

util-read.R
```diff
@@ -387,6 +388,64 @@ read.issues = function(data.path, issues.sources = c("jira", "github")) {
...
+#'
+#' @param data.path the path to the gender data
+#'
+#' @return the read and parsed gender data
+read.gender = function(data.path) {
+    # constant for seperating key and value
+    SEPERATOR = ";"
+
+    ## get file name of gender data
+    filepath = file.path(data.path, "gender")
```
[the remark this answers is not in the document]
hechtlC: I agree with @bockthom. Please add the '.list' ending to the gender files. I will update that on the cluster then accordingly.

tests/codeface-data/results/testing/test_gender/gender
```diff
@@ -0,0 +1,9 @@
+Thomas;male
+Olaf;female
+Karl;male
+Max;male
+udo;female
+Björn;male
+georg;male
+Hans;male
+Fritz fritz@example.org;
```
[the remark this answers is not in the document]
miriyusifli: @bockthom The implementation is a little bit different from your idea. As you know, there are different gender detection tools. [GenderAPI](https://gender-api.com/) is just one of them. I think it would be better to keep the implementation as it is instead of adding `GENDER.LIST.VALUES` to restrict possible values of the gender column since different gender detection tools use different labels for genders. Some of them use many labels such as `mostly_male`, `mostly_female` and etc. The current implementation can be used with any tool because there is no restriction about labels. Some tools do not use the `unknown` label. They just leave the cell empty. Therefore, I added this line to the test to be sure that also this line can be parsed successfully.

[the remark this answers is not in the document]
hechtlC: @miriyusifli and I just discussed this and we decided to go with the way you suggested, @bockthom. If we ever encounter other labels, we can simply add them without any big problems.

tests/test-data.R
```diff
@@ -40,6 +40,7 @@ test_that("Compare two ProjectData objects on empty data", {
     ## initialize a ProjectData object with the ProjectConf and clone it into another one
     proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY_EMPTY, ARTIFACT)
     proj.conf$update.value("pasta", TRUE)
+    proj.conf$update.value("gender", TRUE)
```
bockthom, line 43: Please add you copyright header to this file `tests/test-data.R`, this is the only file where it is still missing.

NEWS.md
```diff
@@ -2,6 +2,15 @@
 
 # coronet – Changelog
 
+## Unversioned
+
+### Added
+- Incorporate gender data, i.e., add a configuration entry to the project configuration, add function `read.gender` for reading gender data, add functions `get.gender` and `set.gender` and corresponding utility functions to automatically merge gender data to the author data (PR #216, ebe4899dad80f37cf2ced20ed1591b9b7fd2f2f4, 27bf76288525a75e0f948570f43e16f947874ee3, 20a4e9013b401361b831af1ecafbddb3710253f3, 9c9d2612bb481d0f6ddbf36b5b22d918fa2d0cab)
```
bockthom, line 8: Please check whether these commit hashes still exist. (I think it should be still ok, but as you and @hechtlC have rebased, it is worth to double-check.)

tests/test-read.R
```diff
@@ -251,6 +252,24 @@ test_that("Read the author data.", {
...
 
+test_that("Read and parse the gender data.", {
+
+    ## configuration object for the datapath
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+
+    ## read the actual data
+    gender.data.read = read.gender(proj.conf$get.value("datapath.gender"))
+    
+    rownames(gender.data.read) = NULL
```
bockthom, line 263: Is it really necessary to remove the rownames here? This is also not done in the other "read" tests. Hence, I don't understand why this is necessary, and therefore, I would like to remove this statement here. Please check whether this can be removed without letting the test fail. If it fails after removing, then this should be fixed somewhere else, but not in the test, I guess. We also don't do this in the other tests here...

bockthom (changes requested):
@miriyusifli Thanks a lot for your changes, this PR looks quite good now.
I had a look at your changes and spotted a few newly introduced inconsistencies which should be easy to fix, please find my comments below. In case of questions, don't hesitate to ask.
We are really close to finishing this PR, not much work left :smile: 

(Regarding the failing CI test regarding eigenvector centrality: Don't worry about that, @hechtlC is currently figuring out why this test fails. We'll let you know as soon as we have identified what's going wrong there. This should be independent from your changes. Hence, you can continue working on the remaining issues of your PR.)

tests/test-read.R
miriyusifli: @bockthom I think it is necessary since we order gender data after reading it as described in the [comment](https://github.com/se-sic/coronet/pull/216#discussion_r755158721). Default rownames (1, 2, 3 ...) are added when data is read. After ordering it, also the order of the rownames is changed (9, 1, 3 ...). There are 2 possible solutions here: 
1. Setting rownames in necessary order for gender.data.expected
2. Setting rownames to NULL for gender.data.read.

I think the first solution is not quite good. Since if a new row is added to the list, the order of rownames will change. Therefore, solution 2 is better in my opinion.

NEWS.md
miriyusifli: @bockthom I have made many requested changes since PR creation. Therefore, we have more than 4 commits. Should I add all of the commits to the NEWS.md?

tests/test-read.R
bockthom: Ah I see. Thanks for the detailed explanation, this totally makes sense.
Nevertheless, I would like to move this statement to the read function, such that the row names are already correct when returned from `read.gender` and we don't need to address this problem with in the test.
I'd suggest to move this statement regarding the rownames to the end of the `read.gender`, directly after re-ordering:

```R
    gender.data = gender.data[order(gender.data[["author.name"]]), ] # re-order after read
    rownames(gender.data) = NULL
```
Would this be possible? Then we can remove it from the test. I'd assume that this should work and the test still runs successfully. Please try whether this works. In the case that this does not work, then we keep it as it is now.

README.md
```diff
@@ -557,6 +558,10 @@ There is no way to update the entries, except for the revision-based parameters.
 - `filter.bots`
     * Remove all commits, issues, and mails made by bots. Bots are identified using the `bots.list` file.
     * [`TRUE`, *`FALSE`*]
+- `gender`
+    * Read and add gender data to authors (column `gender`)
+    * [`TRUE`, *`FALSE`*]
+    * **Note**: To include PaStA-based edge attributes, you need to give the `"gender"` edge attribute for `edge.attributes`.
```
bockthom, line 564: I am very sorry, I overlooked this one yesterday: There are no edge attributes for `gender`, so please remove the `**Note**` line here as it does not make sense. Sorry for bothering you here again.

util-read.R
```diff
@@ -511,6 +511,85 @@ create.empty.authors.list = function() {
...
+        ## find authors who have undefined gender labels
+        undefined.labels.authors = filter(gender.data, gender %in% undefined.labels)
+        logging::logwarn(sprintf("Undefined gender labels. %s cannot be used. Only %s are allowed.
+                                 The following authors have undefined labels: %s ",
+                                 paste(shQuote(undefined.labels), collapse = ","),
+                                 paste(shQuote(GENDER.LIST.VALUES), collapse = ", "),
+                                 paste(shQuote(undefined.labels.authors[["author.name"]]), collapse = ",")))
+
+        ## replace all undefined labels with 'unknown'
+        gender.data[["gender"]][gender.data[["gender"]] %in% undefined.labels] = 'unknown'
```
bockthom, line 570: In this line, please use double quotation marks for `"unknown"`. (This is the only line where this is wrong, you did it correctly everywhere else.)

util-read.R
```diff
@@ -511,6 +511,85 @@ create.empty.authors.list = function() {
...
+)
+
+## declare the datatype for each column in the constant 'GENDER.LIST.COLUMNS'
+GENDER.LIST.DATA.TYPES = c(
+    "character", "character"
+)
+
+## declare predefined values for the gender column
+GENDER.LIST.VALUES = c(
+    "male", "female","unknown"
```
bockthom, line 529: Please add a space after the second comma.

NEWS.md
```diff
@@ -2,6 +2,16 @@
...
+
+## Fixed
+- Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`.
```
bockthom, line 11: Here the commit hash is still missing, so instead of the period at the end just add ([this is the commit hash])

NEWS.md
bockthom: No, the requested changes just deal with minor issues (coding style, moving functions etc.) - this is not what should appear in the NEWS. We only need to mention the 4 commits already present there, which contain the basic functionality you have added. And please add one more commit, namely the commit which introduces the `GENDER.LIST.VALUES` as this is also kind of basic functionality you have added regarding gender data. So, having 5 commits here should be enough then.

However, make sure that you update/double-check the commit hashes after you finished rebasing the remaining implementation stuff. Updating the commit hashes should be the very last step :wink:

bockthom (changes requested):
@miriyusifli Great, I like your new changes. There are just three minor issues left (one I have overlooked yesterday, and two newly introduced ones), and the two comments you replied to. Please see my answers below.

As soon as the six remaining issues are addressed, we are ready to merge. Thanks for your patience.

miriyusifli:
@bockthom and @hechtlC all request changes have been applied. I think PR is ready to be merged.

NEWS.md
```diff
@@ -5,11 +5,11 @@
...
 
 ## Fixed
-- Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`. 
-- Fix the filtering of the deleted user in `util-read.R` to always be lowercase as the deleted user can appear with different spellings (#214, cb1615600a8f7ed5801c87ede08cf0e1c3e91e48)
+- Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`. (#214, 25fb86277c7cc15b94ca0327bff4bb7e818ca09b)
```
bockthom, line 11: @miriyusifli Thanks again, your changes to the code look very good to me. I only spotted a minor error in the `README.md`: 
Please remove the `#214, ` before the commit hash here in this line (only in this line it is wrong, for the subsequent line it is correct), as this number is wrong here - there is nothing we should reference in this line, because this fix is not related to any numbered issue. So, in this line, it just should be `(25fb86277c7cc15b94ca0327bff4bb7e818ca09b)`. Please also remove the period directly before the opening parenthesis. Everything else looks fine, it is only this single line targeting the eigenvector centrality fix which needs to be adjusted.

miriyusifli:
@bockthom Done :)

bockthom (approved):
Thanks a lot @miriyusifli. Thanks for your endurance and thanks for contributing to `coronet`. Good work, this is a nice and beneficial enhancement for `coronet`.

We (@hechtlC and me) will merge this PR soon (and deal with the CI failure regarding eigenvector-centrality afterwards, as this failure is not related to your changes but affects `coronet` in general). So, now you're really done with this PR :partying_face: 

We wish you all the best for the future.

hechtlC (approved):
Thanks a lot @miriyusifli for the endurance while working on this PR! Ans also thanks for the nice additions to coronet.

@bockthom you can merge this once you are ready.

Issue #217, opened 2021-12-02
Missing data in `get.first.activity.data` leads to an error


bockthom:
### Description

When getting the first activity data via `get.first.activity.data`, using multiple activity types (`activity.type = c("mails", "commits", "issues")`), 
we end up in an error when there is no activity for several of these activity types.

In particular, assume that there is no "mail" activity in the range data and no "commit" activity in the range data, but only "issue" activity. In such a case, when reducing `activity.by.type`, we get an error here:
https://github.com/se-sic/coronet/blob/a65602663efbe208e201e39418a42d8151903bbb/util-networks-covariates.R#L1031
In this case, both `x` ("mails") and `y` ("commits") are empty lists (`list()`) in the first step of the reduction, making the reduction fail as there are no names in `x` and `y`.

#### Quick and dirty solution: 
Check if `keys` is `NULL` before https://github.com/se-sic/coronet/blob/a65602663efbe208e201e39418a42d8151903bbb/util-networks-covariates.R#L1040, and if so, return and empty list instead of combining the values for the keys. Disadvantage: When someone expects that there is a "commits" and "emails" named element in the resulting list for each developer, then this will lead to problems as there is no "commits" or "emails" element in the resulting list.

#### Preferred solution: 
Make sure that the input lists for the reduction (i.e., `x` and `y`) are not empty. So, check before the reduction whether multiple lists in `activity.by.type` are empty, and if there are some, just copy one of the non-empty lists and set all its values to `NA`. Then reducing the lists should work without any problem and the resulting list in the end should have elements named "commits" and "emails" for each developer though commit data and e-mail data are empty. 

### Versions

This affects all versions in which the function `get.first.activity.data` exists.

PR #220, opened 2022-02-02
Add check to get.first.activity.data to look for missing activity types

nlschn:
Adresses issue #217

Signed-off-by: Niklas Schneider <s8nlschn@stud.uni-saarland.de>

### Prerequisites

### Description

Add check to `get.first.activity.data` function in `util-networks-covariates.R` to check whether there are activity types that configured but not present in the given data.

If there are no activities at all, print a warning and return an empty list immediately (this is referred to as the "quick and dirty solution" in #217...). If there are activity types missing but also some present, copy an entry from a present key, replace its elements with `NA`, and add that list for each missing key . Also print a warning in this case.

This pull request addresses #217.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@
 ### Fixed
 - Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`.(25fb86277c7cc15b94ca0327bff4bb7e818ca09b)
 - Fix the filtering of the deleted user in `util-read.R` to always be lowercase as the deleted user can appear with different spellings (#214, 1b4072c7ec0e33a595e31d9e9d27bb5c133b1556)
-
+- Add check to `get.first.activity.data` to look for missing activity types. If no activities are in the RangeData, the function will print a warning and return an empty list (PR #220, #217, 5dc975f51320fd80b79f533d9a86a1916a0562a2, 01f43a82777ded0a639d6089b2c884407e1a2af1, f3c6c305edb1a524abf70f40e8cdf0559e5c9798)
 
 ## 4.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@
 ### Fixed
 - Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`.(25fb86277c7cc15b94ca0327bff4bb7e818ca09b)
 - Fix the filtering of the deleted user in `util-read.R` to always be lowercase as the deleted user can appear with different spellings (#214, 1b4072c7ec0e33a595e31d9e9d27bb5c133b1556)
-- Add check to `get.first.activity.data` to look for missing activity types. If no activities are in the RangeData, the function will print a warning and return an empty list (PR #220, #217, 5dc975f51320fd80b79f533d9a86a1916a0562a2, 01f43a82777ded0a639d6089b2c884407e1a2af1, f3c6c305edb1a524abf70f40e8cdf0559e5c9798)
+- Add check to `get.first.activity.data` to look for missing activity types. If no activities are in the RangeData, the function will print a warning and return an empty list (PR #220, #217, 5707517600c5579095c245b63c745d01cde02799, 42a4befb36e7fd9830924dc7fb2e04ecdf86e209,  d6424c03baff05562448df1b6b87828ca9a37b88, ca8a1b4c628261dcb471e1da3603439e75e4cc56, f6553c6106e5fec3837c6edb906a4d0960c5c5fb)
 
 ## 4.0
 
--- a/tests/test-networks-covariates.R
+++ b/tests/test-networks-covariates.R
@@ -1484,9 +1484,7 @@ test_that("Test get.first.activity.data with missing commits and mails for all a
                                        bins = c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]),
                                        split.basis = c("issues"))[[1]]
 
-    ## assert that warnings are thrown for the missing data sources (making the test pass instead of result in warning).
-    ## here, using '=' instead of '<-' leads to an error.
-    expect_warning(first.activity.data <- get.first.activity.data(range.data))
+    first.activity.data = get.first.activity.data(range.data)
 
     ## leave the authors without any data out
     expected.activity.data = list(
--- a/tests/test-networks-covariates.R
+++ b/tests/test-networks-covariates.R
@@ -19,7 +19,7 @@
 ## Copyright 2018-2019 by Klara Schlüter <schluete@fim.uni-passau.de>
 ## Copyright 2018-2019 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
 ## Copyright 2021 by Johannes Hostert <s8johost@stud.uni-saarland.de>
-## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
+## Copyright 2021-2022 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -839,7 +839,6 @@ test_that("Test add.vertex.attribute.issue.comment.count", {
     })
 })
 
-
 #' Test the add.vertex.attribute.author.email method
 test_that("Test add.vertex.attribute.author.email", {
 
@@ -1406,4 +1405,58 @@ test_that("Test addition of attributes despite of non-captured vertices", {
     expect_true("commit.count.committer.and.author" %in% igraph::list.vertex.attributes(net.commit.count))
     expect_identical(igraph::get.vertex.attribute(net.commit.count, "commit.count.committer.and.author"), 0L)
 
+})
+
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Unit tests for helper functions------------------------------------------
+#' Test addition of attributes despite of non-captured vertices
+test_that("Test get.first.activity.data with missing commits and mails", {
+
+    ## initialize a ProjectData object with the ProjectConf
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.data.base = ProjectData$new(project.conf = proj.conf)
+
+    ## create a RangeData object with the same data sources as proj.data.base
+    ## here, some data is already missing; this is used to test whether the function correctly sets missing data as 'NA'
+    ## lists
+    timestamps = proj.data.base$get.data.timestamps(outermost = TRUE)
+    range.data = split.data.time.based(proj.data.base, bins =
+                                                c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]))[[1]]
+
+    first.activity.data = get.first.activity.data(range.data)
+
+    expected.activity.data = list(
+        list(commits = get.date.from.string("2016-07-12 15:58:59"),
+             mails = get.date.from.string("2004-10-09 18:38:13"),
+                     issues = get.date.from.string("2013-05-05 21:46:30")),
+        list(commits = get.date.from.string("2016-07-12 16:00:45"),
+             mails = get.date.from.string(" 2016-07-12 15:58.50"),
+                    issues = get.date.from.string("2013-05-25 03:25:06")),
+        list(commits = get.date.from.string("2016-07-12 16:06:32"),
+             mails = get.date.from.string("2016-07-12 16:04:40"),
+             issues = get.date.from.string("2013-04-21 23:52:09")),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string("2010-07-12 11:05:35"),
+             issues = get.date.from.string(NA)),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string("2010-07-12 12:05:34"),
+             issues = get.date.from.string(NA)),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string("2010-07-12 12:05:40"),
+             issues = get.date.from.string(NA)),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string("2010-07-12 10:05:36"),
+             issues = get.date.from.string(NA)),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string(NA),
+             issues = get.date.from.string("2016-07-12 15:59:59")),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string(NA),
+             issues = get.date.from.string("2016-07-15 20:07:47"))
+    )
+    names(expected.activity.data) = c("Björn", "Olaf", "Thomas", "Fritz fritz@example.org", "georg", "Hans",
+                                      "udo", "Karl", "Max")
+
+    expect_equal(first.activity.data, expected.activity.data)
 })
--- a/tests/test-networks-covariates.R
+++ b/tests/test-networks-covariates.R
@@ -1410,53 +1410,101 @@ test_that("Test addition of attributes despite of non-captured vertices", {
 
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Unit tests for helper functions------------------------------------------
-#' Test addition of attributes despite of non-captured vertices
-test_that("Test get.first.activity.data with missing commits and mails", {
+
+#' Test get.first.activity.data with missing commits, mails, and issues.
+#' Not all authors are expected to have activities in all data sources
+test_that("Test get.first.activity.data with missing commits, mails, and issues", {
 
     ## initialize a ProjectData object with the ProjectConf
     proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
     proj.data.base = ProjectData$new(project.conf = proj.conf)
 
-    ## create a RangeData object with the same data sources as proj.data.base
-    ## here, some data is already missing; this is used to test whether the function correctly sets missing data as 'NA'
-    ## lists
+    ## create a RangeData object with the same data sources as 'proj.data.base'.
+    ## some data is already missing; this is used to test whether the function correctly sets missing data as 'NA'.
     timestamps = proj.data.base$get.data.timestamps(outermost = TRUE)
-    range.data = split.data.time.based(proj.data.base, bins =
-                                                c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]))[[1]]
+    range.data = split.data.time.based(proj.data.base,
+                                       bins = c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]))[[1]]
 
     first.activity.data = get.first.activity.data(range.data)
 
     expected.activity.data = list(
-        list(commits = get.date.from.string("2016-07-12 15:58:59"),
-             mails = get.date.from.string("2004-10-09 18:38:13"),
-                     issues = get.date.from.string("2013-05-05 21:46:30")),
-        list(commits = get.date.from.string("2016-07-12 16:00:45"),
-             mails = get.date.from.string(" 2016-07-12 15:58.50"),
-                    issues = get.date.from.string("2013-05-25 03:25:06")),
-        list(commits = get.date.from.string("2016-07-12 16:06:32"),
-             mails = get.date.from.string("2016-07-12 16:04:40"),
-             issues = get.date.from.string("2013-04-21 23:52:09")),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string("2010-07-12 11:05:35"),
-             issues = get.date.from.string(NA)),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string("2010-07-12 12:05:34"),
-             issues = get.date.from.string(NA)),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string("2010-07-12 12:05:40"),
-             issues = get.date.from.string(NA)),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string("2010-07-12 10:05:36"),
-             issues = get.date.from.string(NA)),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string(NA),
-             issues = get.date.from.string("2016-07-12 15:59:59")),
-        list(commits = get.date.from.string(NA),
-             mails = get.date.from.string(NA),
-             issues = get.date.from.string("2016-07-15 20:07:47"))
+        "Björn" = list(commits = get.date.from.string("2016-07-12 15:58:59"),
+                       mails = get.date.from.string("2004-10-09 18:38:13"),
+                       issues = get.date.from.string("2013-05-05 21:46:30")),
+        "Olaf" = list(commits = get.date.from.string("2016-07-12 16:00:45"),
+                      mails = get.date.from.string("2016-07-12 15:58:50"),
+                      issues = get.date.from.string("2013-05-25 03:25:06")),
+        "Thomas" = list(commits = get.date.from.string("2016-07-12 16:06:32"),
+                        mails = get.date.from.string("2016-07-12 16:04:40"),
+                        issues = get.date.from.string("2013-04-21 23:52:09")),
+        "Fritz fritz@example.org" = list(commits = get.date.from.string(NA),
+                                         mails = get.date.from.string("2010-07-12 11:05:35"),
+                                         issues = get.date.from.string(NA)),
+        "georg" = list(commits = get.date.from.string(NA),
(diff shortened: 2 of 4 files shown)


nlschn:
There seems to be a problem when `present.keys` has no elements. In these test cases the result (returning an empty list as there is no data for the configured activity types) does not match the expected result.

bockthom:
> There seems to be a problem when `present.keys` has no elements. In these test cases the result (returning an empty list as there is no data for the configured activity types) does not match the expected result.

How can it happen that `present.keys` has no elements? This could only happen when there is no data available, as far as I understand. The failing tests call the function `get.network.covariates.test.networks`, which returns a list containing a `networks` and a `project.data` element. Within this function, there also is a variable [`input.data`](https://github.com/se-sic/coronet/blob/a65602663efbe208e201e39418a42d8151903bbb/tests/test-networks-covariates.R#L84) which contains the data for separate ranges. It could be the case, that some of the data of interest (e.g. "mails") may be empty in one the ranges, which could cause the data to be empty when we choose `activity.by.type = "mails"`. This might be a case in which no `present.keys` are available, but I am not sure how this case was handled before. Therefore, we need to find out what happens now and what we have expected.

To find out where exactly this problem actually occurs, I suggest to debug one of the failing tests, e.g., line 1028 in `test-networks-covariates.R`:
https://github.com/se-sic/coronet/blob/a65602663efbe208e201e39418a42d8151903bbb/tests/test-networks-covariates.R#L1028-L1037
As we are iterating over several lists in this test, it is not that easy to find out where exactly the problem occurs. In a first step, we need to find out for which `AGGREGATION.LEVELS` it happens and then for which list element of the attributes list.
At the moment, I only  have two ideas in mind, how to come to an end fast: (1) Print the `level` to see in which level we get the (first) error, and (2) print the actual and expected attributes list to see where exactly there is a difference in the list (maybe using a diff-tool to be able to spot the difference in the string representations of the two lists...)

nlschn:
> How can it happen that present.keys has no elements? This could only happen when there is no data available, as far as I understand.

It is also possible if there is data available, but not for the types given in the parameter `activity types`.

bockthom:
> It is also possible if there is data available, but not for the types given in the parameter `activity types`.
This is true. So with "no data" I meant "no data for the chosen activity types".

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
         )
         return(minima.per.person)
     })
+    names(activity.by.type) = activity.types
+
+    ## if activity.by.type is null or the list is empty, print a warning and return an empty list
```
bockthom, line 1027: Could you please use `'` around variable names in comments? So here it should be `'activity.by.type'`.

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+    if (is.null(activity.by.type) || length(activity.by.type) == 0) {
+        logging::logwarn("There were no activities in the given RangeData at all")
+        return(list())
+    }
+
+    ## first, get the keys that are entirely missing or present
+    missing.keys = setdiff(names(activity.by.type), activity.types)
+    present.keys = intersect(names(activity.by.type), activity.types)
```
bockthom, line 1035: Does this part of the implementation still make sense? As we set `names(activity.by.type) = activity.types` a view lines above, the `setdiff` will always be empty by construction and `intersect` will always end up in `activity.types`.

So, what I see from now is that we do not even need to to set  `names(activity.by.type) = activity.types` as we don't need the names. Instead of `present.keys`, we could just use `activity.types` in the statement below:

```R
present.keys = present.keys[!is.na(activity.by.type[present.keys]) & length(activity.by.type[present.keys]) > 0]
append(missing.keys,
        present.keys[is.na(activity.by.type[present.keys]) | length(activity.by.type[present.keys]) == 0])
```
In addition, you append the missing keys after removing some present keys from `present.keys` here. The order of these statements should be switched, shouldn't it?
This should work then (if I don't have overlooked anything...)

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+
+    ## now check for the present keys whether their list is actually empty; in this case, the key is also declared
+    ## missing
```
bockthom, line 1038: Could you please use just one line for the comment? We allow 120 characters per line, so it should fit into one line, I guess. (And don't forget to update the comment if "present keys" will not exist any more...)

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+    present.keys = present.keys[!is.na(activity.by.type[present.keys]) & length(activity.by.type[present.keys]) > 0]
+    append(missing.keys,
+        present.keys[is.na(activity.by.type[present.keys]) | length(activity.by.type[present.keys]) == 0])
+
+    ## if there are no keys left, that are present, again, print a warning and return an empty list as there is no data
+    ## for the configured activity types
+    if (length(present.keys) == 0 || is.null(present.keys)) {
+        logging::logwarn("There were no activities in the given RangeData that were configured")
+        return(list(NA))
```
bockthom, line 1047: What is the reason for returning `list(NA)` here? A few lines above, you return just `list()` after the null check. Is this difference intended? If so, please explain somewhere in the comments and also update the function documention `@return` as there you only mentioned returning an empty list.

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+    }
+
+    ## else, take the first present item, copy its list and replace all entries with NA so that the reduce and apply
+    ## can work it out
```
bockthom, line 1051: This comment should also fit into one line, I guess (but my guess can also be wrong :wink:)

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+    na.list = lapply(activity.by.type[present.keys[[1]]], function (x) NA)
```
bockthom, line 1052: Could you please remove the space between `function` and the opening parenthesis? 

In addition, according to our contribution guide, we "always use the function `return()` to return a value from a function", so please use `return(NA)` here :wink:

util-networks-covariates.R
```diff
@@ -1019,6 +1022,41 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+
+    for (missing.key in missing.keys) {
+        logging::logwarn(paste(c("The type", missing.key, "was configured",
+                                 "but the RangeData did not cotain any",
+                                 "activities of that type"), sep = " "))
```
bockthom, line 1057: I guess we can also put this into two lines instead of three :wink:

bockthom (changes requested):
Thanks for this PR @nlschn. During my review I identified that the change we have applied together yesterday (namely, adding `names(activity.by.type) = activity.types`) breaks the rest of the implementation as the identification of missing keys does not work any more since we have added the names before... Sorry for my wrong suggestion yesterday, yesterday I did not check what we do afterwards...
However, I have a suggestion which might circumvent this problem; please see my comments below. (In addition, there are also just a few formatting inconsistencies.)

As soon as you have changed your implementation such that it works again, I will try whether the implementation also works in the situation in which I spotted the bug, just to make sure that everything works as expected then.

util-networks-covariates.R
nlschn: No, it has 133 chars. 😂

util-networks-covariates.R
nlschn: I need the `present.keys` to find the candidates I can copy the list from, if I remember correctly. But apart from that, you're right.
Did I understand you correctly that new assignment of the names `names(activity.by.type) = activity.types` should be kept?

util-networks-covariates.R
bockthom: Damn! Then my visual estimation was far off.
Maybe it would be a better idea if I abstain from such visual estimations in the future...:joy:

util-networks-covariates.R
bockthom: Right. We don't need the names, so we could remove them. But it may help for future debugging to keep the names... Remember that we've struggled with the example in the comments above as there we had names but did not have the names in the resulting list. So, to prevent struggling with the names again, we can just keep it.

util-networks-covariates.R
```diff
@@ -1024,36 +1024,31 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
         logging::logwarn("There were no activities in the given RangeData at all")
         return(list())
     }
 
-    ## first, get the keys that are entirely missing or present
-    missing.keys = setdiff(names(activity.by.type), activity.types)
-    present.keys = intersect(names(activity.by.type), activity.types)
+    ## check for keys whose member lists are not empty or na
+    missing.keys = activity.types[is.na(activity.by.type[activity.types]) |
+                                      length(activity.by.type[activity.types]) == 0]
```
bockthom, line 1035: Please spell `NA` in the comment with upper case letters.
And please fix the indentation in the second line such that the `l` of length and the `i` of is.na align.

The rest looks good so far. That's all for today. I will think about the correctness of the new implementation tomorrow with a fresh mind :wink:

bockthom: I had a closer look at the new implementation again, and I take the view that it now should work as intended :partying_face: 

The only issue I spotted targets the comment again: "check for keys whose member lists are not empty or NA" - but aren't you looking for empty keys here? So the "not" in the comment is wrong, isn't it?

Before merging, I will check out the new implementation and apply it to to the cases where the previous implementation failed. Let's see if it solves the issue, but I am pretty confident in that. Though, I cannot predict when I will find time to try it out, but @hechtlC can already start reviewing this PR.

bockthom (changes requested):
I am very very sorry, but I have bad news :astonished: :speak_no_evil: 

After spending more than an hour on figuring out in which project the problem this PR deals with occurs, I had to notice that the new implementation still does not prevent the error that it should fix...

I identified two lines in the implementation which make the new fix ineffective:  (1) The check for missing keys does not cover the missing keys in the situation in which the bug occurs. (2) The replacement with `NA` replaces a complete list by `NA` instead of replacing each element by `NA`. See details on that below.

I have to admit that I did not spot these problems in the code and, moreover, that I did not remember all details of the problematic behavior. It would have been better to store a error dump containing the whole environment in which the bug occurs when creating issue #217 to have some test data available for easily dealing with the situation. :see_no_evil:  Next time, I'll do better, hopefully...

@nlschn I will try to derive real input data from the real situation in which the bug occurs and provide it to you, such that you can use the provided data to play around with and fix the currently ineffective statements. (I don't know why didn't I do this earlier...) However, I am not sure whether I can do this tomorrow, but I'll get back to you next week at latest. But there is no rush, you might also be busy with other, more important, tasks at the moment anyway.

util-networks-covariates.R
```diff
@@ -1019,6 +1022,36 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+    ## if there are no keys left that are present, again, print a warning and return an empty list as there is no data
+    ## for the configured activity types
+    if (length(present.keys) == 0 || is.na(present.keys) || is.null(present.keys)) {
+        logging::logwarn("There were no activities in the given RangeData that were configured")
+        return(list())
+    }
+
+    ## else, take the first present item, copy its list and replace all entries with NA so that the reduce and apply can
+    ## work it out
+    na.list = lapply(activity.by.type[present.keys[[1]]], function(x) return(NA))
```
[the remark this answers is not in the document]
bockthom: Unfortunately, `na.list` still is wrongly. This ends up in a completely weird result:
If mails and commits are empty, we expect something like 
```R
list(
    authorA = list(commits = NA, mails = NA, issues = <date-retrieved-from-issues>),
    authorB = list(commits = NA, mails = NA, issues = <date-retrieved-from-issues>), 
    ...
)
```
However, what we currently get is the following:
```R
list(
    authorA = list(commits = <date-retrieved-from-issues>, mails = NA),
    authorB = list(commits = <date-retrieved-from-issues>, mails = NA, 
    ...
)
```
I don't know how this can happen, but the result is ill-formated (issues are missing) and wrong (the `issue` date occurs at `commits` although there should be an `NA`). Although no error is thrown during the function call, accessing the returned data afterwards still ends up in an error...

So, the problem still seems to be the creation of the `na.list`: Currently when setting `activity.by.type[missing.key] = na.list` we get `list(mails=NA)`, but what we need is `list(authorA = list(mails = NA), authorB = list(mails = NA))`. So, when creating `na.list`, we need to create a `list(mails = NA)` for every author entry...

bockthom (changes requested):
Thanks for the fix @nlschn. 

Fortunately, determining the missing keys works as expected now. Since `missing.keys` are determined correctly now, the corresponding warning is printed correctly, but I spotted a formatting issue when the warning is to be printed. Please see my suggestion on how to prevent this below.

Unfortunately, computing the `na.list` still does not work, ending up in wrong results and an error when accessing the result afterwards. Please see the details below.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
     expect_identical(igraph::get.vertex.attribute(net.commit.count, "commit.count.committer.and.author"), 0L)
 
 })
+
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Unit tests for helper functions------------------------------------------
+#' Test addition of attributes despite of non-captured vertices
```
bockthom, line 1413: Could you please add a blank line between `## Unit tests for ...` and `#' Test addition...` as one comment describes the section and the other comment describes the test function.
In addition, both @hechtlC and I don't understand the description of the test function "Test addition of attributes despite of non-captured vertices". So, fist of all, there are no vertices in this test as it does not test networks but just data aggregation. Hence, the second part of this sentence could be something like "without creating networks". And please also rephrase the beginning "Test addition of attributes". There is no need to say that this is a test addition, it actually is a stand-alone test. So, please just describe what is tested here.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+test_that("Test get.first.activity.data with missing commits and mails", {
+
+    ## initialize a ProjectData object with the ProjectConf
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.data.base = ProjectData$new(project.conf = proj.conf)
+
+    ## create a RangeData object with the same data sources as proj.data.base
+    ## here, some data is already missing; this is used to test whether the function correctly sets missing data as 'NA'
+    ## lists
```
bockthom, line 1422: Please re-format this comment. As it is a multi-sentence comment, please add periods to mark the end of the sentences.
In addition, the single word in the last line looks a bit lost. Please adjust that. And also remember to add single quotation marks when using variable names, so it should be `'proj.data.base'`.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+    timestamps = proj.data.base$get.data.timestamps(outermost = TRUE)
+    range.data = split.data.time.based(proj.data.base, bins =
+                                                c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]))[[1]]
```
bockthom, line 1425: Please adjust the indentation of the second line to the ( of the function call.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+
+    first.activity.data = get.first.activity.data(range.data)
+
+    expected.activity.data = list(
+        list(commits = get.date.from.string("2016-07-12 15:58:59"),
+             mails = get.date.from.string("2004-10-09 18:38:13"),
+                     issues = get.date.from.string("2013-05-05 21:46:30")),
```
bockthom, line 1432: Please adjust the indentation.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+        list(commits = get.date.from.string("2016-07-12 16:00:45"),
+             mails = get.date.from.string(" 2016-07-12 15:58.50"),
+                    issues = get.date.from.string("2013-05-25 03:25:06")),
```
bockthom, line 1435: Please adjust the indentation

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+                                                c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]))[[1]]
+
+    first.activity.data = get.first.activity.data(range.data)
+
+    expected.activity.data = list(
+        list(commits = get.date.from.string("2016-07-12 15:58:59"),
+             mails = get.date.from.string("2004-10-09 18:38:13"),
+                     issues = get.date.from.string("2013-05-05 21:46:30")),
+        list(commits = get.date.from.string("2016-07-12 16:00:45"),
+             mails = get.date.from.string(" 2016-07-12 15:58.50"),
```
bockthom, line 1434: Please remove the space at the beginning of the character string and also replace the dot which is separating minute and second by a colon :wink:

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+             issues = get.date.from.string(NA)),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string(NA),
+             issues = get.date.from.string("2016-07-12 15:59:59")),
+        list(commits = get.date.from.string(NA),
+             mails = get.date.from.string(NA),
+             issues = get.date.from.string("2016-07-15 20:07:47"))
+    )
+    names(expected.activity.data) = c("Björn", "Olaf", "Thomas", "Fritz fritz@example.org", "georg", "Hans",
+                                      "udo", "Karl", "Max")
```
bockthom, line 1459: Could we even put the names directly in the list above? For instance, as follows:
```R
"Björn" = list(commits = ...
```
This would make understanding the test easier.

tests/test-networks-covariates.R
```diff
@@ -1407,3 +1406,57 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+
+    expect_equal(first.activity.data, expected.activity.data)
```
bockthom, line 1461: This test is correct and very useful, but it does not yet test the corner case for which we have fixed the bug in the function `get.first.activity.data`.
So, please add a second test case below, for which one or even two of the data sources need to be empty for every developer. So, for instance, call `set.issues(create.empty.issues.list())` and `set.mails(create.empty.mails.list())` before calling `get.first.activity.data`. Only in these cases we test the new behavior. The test above would also have worked before the bug fix yet.

bockthom (changes requested):
Thanks for the new test @nlschn. The test looks good and correct, but it does not test the corner cases we need to test. So, I would like to keep this test and add another one addressing the corner case. See the details in the comments below (beside some comments addressing formatting issues).

tests/test-networks-covariates.R
```diff
@@ -1410,53 +1410,101 @@ test_that("Test addition of attributes despite of non-captured vertices", {
...
+    ## create a RangeData object with the same data sources as 'proj.data.base'.
+    ## some data is already missing; this is used to test whether the function correctly sets missing data as 'NA'.
+    range.data = split.data.time.based(proj.data,
+                                       bins = c(timestamps[["start"]][[1]], timestamps[["end"]][[1]]),
+                                       split.basis = c("issues"))[[1]]
+
+
+    ## assert that warnings are thrown for the missing data sources (making the test pass instead of result in warning).
+    ##here, using '=' instead of '<-' leads to an error.
+    expect_warning(first.activity.data <- get.first.activity.data(range.data))
```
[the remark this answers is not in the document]
nlschn: But why does the test still pass instead of output the result "Warning"?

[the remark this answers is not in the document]
bockthom: Unfortunately, the test does not pass (at least, in the CI pipeline). The test ends up in an error as no warning is thrown. So, I don't really understand your question.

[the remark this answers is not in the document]
nlschn: This is strange as it passes on my computer with the `expect_warning` line and without it I get
```
Warning (test-networks-covariates.R:1490:5): Test get.first.activity.data with missing commits and mails for all authors
one argument not used by format 'The type mails was configured but the RangeData did not cotain any activities of that type'
```
and another one similar to this for the second warning.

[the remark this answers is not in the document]
bockthom: Hm, strange. Let me try this out also on my computer to see what happens there, but I don't know when I will find some time to do that, maybe tonight, but I cannot promise that. Anyway, I'd assume that we will end up removing this check as it won't work. But still let's give it a try.

As all the other (really important) issues seem to be fixed now, I will request a review for this PR from @hechtlC now.

bockthom:
@nlschn  As PR #221 has been merged to `dev` in the meanwhile, please rebase your branch onto the current `dev` branch, and don't forget to update your already added commit hashes in the `NEWS.md`.

NEWS.md
```diff
@@ -13,7 +13,7 @@
 ### Fixed
 - Fix values in test for the eigenvector centrality as igraph has changed the calculation of this with version 1.2.7. Also put a warning that we recommend version 1.2.7 in `install.R` and document it in the `README.md`.(25fb86277c7cc15b94ca0327bff4bb7e818ca09b)
 - Fix the filtering of the deleted user in `util-read.R` to always be lowercase as the deleted user can appear with different spellings (#214, 1b4072c7ec0e33a595e31d9e9d27bb5c133b1556)
-- Add check to `get.first.activity.data` to look for missing activity types. If no activities are in the RangeData, the function will print a warning and return an empty list (PR #220, #217, 5dc975f51320fd80b79f533d9a86a1916a0562a2, 01f43a82777ded0a639d6089b2c884407e1a2af1, f3c6c305edb1a524abf70f40e8cdf0559e5c9798)
+- Add check to `get.first.activity.data` to look for missing activity types. If no activities are in the RangeData, the function will print a warning and return an empty list (PR #220, #217, 5707517600c5579095c245b63c745d01cde02799, 42a4befb36e7fd9830924dc7fb2e04ecdf86e209, f3c6c305edb1a524abf70f40e8cdf0559e5c9798, d6424c03baff05562448df1b6b87828ca9a37b88)
```
bockthom, line 16: It looks like that you have omitted to replace the third hash, as it is still the same as before.

In addition, please add the following two commits (which you have authored after your previous update to the changelog):
ca8a1b4c628261dcb471e1da3603439e75e4cc56
f6553c6106e5fec3837c6edb906a4d0960c5c5fb

util-networks-covariates.R
```diff
@@ -1014,11 +1017,50 @@ get.first.activity.data = function(range.data, activity.types = c("commits", "ma
...
+
+    ## else, take the first present item, copy its list and replace all entries with NA so that the reduce and apply can
+    ## work it out
+    na.list = lapply(activity.by.type[present.keys[[1]]], function(x) {
+        authors = lapply(x, function(author) return(NA))
+        return(authors)
+    })
+
+    for (missing.key in missing.keys) {
+        logging::logwarn(paste("The type", missing.key, "was configured but the RangeData did not cotain any",
```
hechtlC, line 1060: contain not cotain :)

hechtlC (approved):
The pull request looks good now. I just found one typo. Please fix that and then apply the patch that I will send you. Then it can be merged.

tests/test-networks-covariates.R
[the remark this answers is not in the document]
bockthom: Just a final note on this discussion about warnings: I debugged and researched on that. There actually seems to be no way to handle warnings as expected by 'testthat'.
- First of all, the `expect_warning` call only is supposed to work on `warning` calls but not on any warning logs such as `logging::logwarn`.
- Second, there is no possibility to prevent `testthat` from counting and reporting warnings. So, warnings cannot be ignored. The only way for warnings not being reported by `testthat` would be a `suppressWarnings(...)` call around each statement that is expected to print a warning. However, this has two drawbacks: When calling `suppressWarnings`, the warnings will also be removed from the log. And `suppressWarnings` also breakes the `CheckReporter` of `testthat`, resulting in not reporting any information about failed and successful tests. Hence, if we want the test results to be properly reported, we must not used `suppressWarnings`.

Summary: We neither can check for warnings printed via `logwarn`, nor can we `testthat` make ignore the warnings in the report. Consequently, we have to stick to the current implementation (although I would have liked to find another solution).

bockthom (approved):
Everything looks good now. Thanks again @nlschn for your efforts.
I have also run the current implementation on the projects for which calling `get.first.activity.data` formerly broke, and now it runs successfully!

So, I will merge this PR soon, as the CI test runs are successful (even though the CI pipeline, for most R versions, fails - but it fails for three different reasons which I had debugged last night. I will probably fix these CI issues on various R versions next week, after this PR has been merged, hopefully, such that we will see the green check mark soon again :smile: )

PR #223, opened 2022-04-08
Fix CI pipeline

bockthom:
### Prerequisites

### Description

Some tests in our CI pipeline failed for multiple reasons:
- The steps for `R` version `3.3` failed as some of the used packages require version `3.4` in the most recent package version and are, therefore, not available any more for `3.3.`. Hence, we exclude this `R` version from the pipeline.
- The steps for `R` version `4.1` and higher failed as the used `r-base` docker container uses some buggy version of Debian which uses `glibc` version >= 2.33 which prevents R to be started within the buggy docker container. Therefore, we switched to the `r-ver` docker container in our CI pipeline to prevent this problem (and manually set the package repository to CRAN).
- Some tests regarding eigenvector centrality potentially fail on some machines do to misordering of data.frame rows due to inaccurracies in floating point operations. To have a potential fix if such a problem occurs again, we now have a workaround for this test.

Last but not least: As `igraph` has released a new package version which contains a lot of bugfixes, we now recommend to use, at least, version `1.3.0` of `igraph`.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,7 +11,7 @@
 ## with this program; if not, write to the Free Software Foundation, Inc.,
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
-## Copyright 2020-2021 by Thomas Bock <bockthom@cs.uni-saarland.de>
+## Copyright 2020-2022 by Thomas Bock <bockthom@cs.uni-saarland.de>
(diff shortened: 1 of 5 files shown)


hechtlC (dismissed):
Thank you very much for the fix of the Pipeline @bockthom! Looks good to me so that can be merged once you are ready.

PR #225, opened 2022-05-08
Cross product of patrick test cases

joba00002:
This will be a useful utility for several new tests that are planned.

Signed-off-by: Jonathan Baumann <joba00002@stud.uni-saarland.de>

### Prerequisites

### DIFF
--- a/tests/testing-utils.R
+++ b/tests/testing-utils.R
@@ -24,6 +24,27 @@ requireNamespace("patrick")
 #' test names are joined with a comma.
 #' The variable names in the two cases objects are assumed to be disjoint.
 #'
+#' Example:
+#'
+#' joined.cases = cases.cross.product(patrick::cases(
+#'      "sliding window: FALSE" = list(test.sliding.window = FALSE),
+#'      "sliding window: TRUE" = list(test.sliding.window = TRUE)
+#' ), patrick::cases(
+#'      "PaStA: FALSE" = list(test.pasta = FALSE),
+#'      "PaStA: TRUE" = list(test.pasta = TRUE)
+#' ))
+#'
+#' yields the following tibble:
+#'
+#' .test_name test.sliding.window test.pasta
+#' 1 sliding window: FALSE, PaStA: FALSE               FALSE      FALSE
+#' 2  sliding window: FALSE, PaStA: TRUE               FALSE       TRUE
+#' 3  sliding window: TRUE, PaStA: FALSE                TRUE      FALSE
+#' 4   sliding window: TRUE, PaStA: TRUE                TRUE       TRUE
+#'
+#' Which can be used like this:
+#' patrick::with_parameters_test_that( {...}, joined.cases)
+#'
 #' @param cases.1 a patrick::cases object (or compatible tibble).
 #' @param cases.2 a patrick::cases object (or compatible tibble).
 #'
--- a/tests/testing-utils.R
+++ b/tests/testing-utils.R
@@ -24,6 +24,8 @@ requireNamespace("patrick")
 #' test names are joined with a comma.
 #' The variable names in the two cases objects are assumed to be disjoint.
 #'
+#' Calls to this function can be nested to generate cross products of more cases.
+#'
 #' Example:
 #'
 #' joined.cases = cases.cross.product(patrick::cases(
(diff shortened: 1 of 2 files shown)


tests/testing-utils.R
```diff
@@ -0,0 +1,63 @@
...
+## with this program; if not, write to the Free Software Foundation, Inc.,
+## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+##
+## Copyright 2022 by Jonathan Baumann <joba00002@stud.uni-saarland.de>
+## All Rights Reserved.
+
+
+#' Construct the 'cross product' of two patrick::cases objects.
+#' Each case of the left object is combined with each case of the right,
+#' test names are joined with a comma.
```
hechtlC, line 20: Left and right could be confusing as the actual parameters are called `cases.1` and `cases.2`. So please either change the comment to e.g., first and second case or change the parameters to `cases.left` and `cases.right`.

tests/testing-utils.R
```diff
@@ -0,0 +1,63 @@
...
+        for (id.2 in seq_along(cases.2)){
+            row.2 = cases.2[id.2,]
+
+            test.name.1 = row.1[[".test_name"]]
+            test.name.2 = row.2[[".test_name"]]
+            ## The new test name should reference both previous names.
+            ## Joining with a comma is an obvious option
+            test.name.combined = paste(test.name.1, test.name.2, sep = ", ")
+
+            ## Select everthing from row.1 and everything but '.test_name' from row.2.
```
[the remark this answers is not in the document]
hechtlC: And additionally typo in 'everthing'.

hechtlC (commented):
Looks good so far @joba00002. But there are a couple of things that need to be changed as @bockthom already indicated.

bockthom:
> Where exactly should I put such an example? If by function documentation, you mean the documentation comment above it, then I'm not sure if I would put an example there. It would take up a fair amount of space, although it would be easy to find there.

My idea was to put it in the roxygen2 function documentation, that is, directly *before* explaining the parameters via `@param`.
We don't care about how much space it takes. More importantly, such an example would tell the user how this function can be practically used---and the function documentation would be the place where I would look for examples.

If you would think that such an example would be way too long (what I cannot imagine yet), you can also post the example here first before integrating it into the codebase.

tests/testing-utils.R
```diff
@@ -24,6 +24,25 @@ requireNamespace("patrick")
...
+#' joined.cases = cases.cross.product(patrick::cases(
+#'      "sliding window: FALSE" = list(test.sliding.window = FALSE),
+#'      "sliding window: TRUE" = list(test.sliding.window = TRUE)
+#' ), patrick::cases(
+#'          "PaStA: FALSE" = list(test.pasta = FALSE),
+#'          "PaStA: TRUE" = list(test.pasta = TRUE)
+#'      ))
+#'
+#'  yields the following tibble:
+#' .test_name test.sliding.window test.pasta
```
bockthom, line 37: Thanks for the example, it looks good! :+1: 

But please increase the readability be adjusting the spacing:
- Add blank line between "Example:" and "joined.cases = ..."
- Add another blank line between "yields the following tibble" and ".test_name ...."
- Adjust the indentation in the example: Is there a reason for the doubled indentation in the second `patrick::cases`? Especially also the position of the `))` at the end looks weird.

Afterwards, as soon as @hechtlC also agrees, we can merge.

bockthom: And could you please add a statement to the `Added` section of the NEWS.md stating that parameterized tests are now possible, for example "Add testing utillity file `tests/testing-utils.R` to make parameterized tests on the cross-product of multiple parameter values possible" or something similar? (This helps tracking the new functionality back, even if it is just testing functionality. When just enhancing, moving, or adding tests, this is not necessary to add, of course.)

---

And just an additional remark (which you can just ignore for this PR, but which might be worth thinking about): This utility function works for two parameters only, if I saw it correctly. What happens if there would be three parameters? Can I just call `cases.cross.product` nested then? For example: 

`cases.cross.product(patrick::cases({...}, cases.cross.product(patrick::cases({...},{...})`

Would this work, or would I need another workaround in this case?

hechtlC: I agree with the formatting issues brought up by @bockthom. Also the Added section in the NEWS.md file.

hechtlC (approved):
Looks good @joba00002! Thanks for the addition. As soon as the issues brought up by @bockthom are resolved, I feel good about merging this.

tests/testing-utils.R
joba00002: > And just an additional remark (which you can just ignore for this PR, but which might be worth thinking about): This utility function works for two parameters only, if I saw it correctly. What happens if there would be three parameters? Can I just call `cases.cross.product` nested then? For example:
> 
> `cases.cross.product(patrick::cases({...}, cases.cross.product(patrick::cases({...},{...}))`
> 
> Would this work, or would I need another workaround in this case?

Yes, nested calls work.

bockthom (approved):
Looks good now. Thanks for testing the "nesting" and fixing the iteration bugs. I also played around with these iterating functions and the solution now seems to work :+1:

Issue #170, opened 2019-07-11
Add further vertex attributes for artifact vertices


clhunsen:
### Description

Related to the idea to add further vertex attributes (see issue #92), we should provide some further vertex attributes for artifact vertices. Currently, we only provide vertex attributes for commits, but we should definitely provide functions to add attributes for e-mails and issues. 

The idea for this originates [here](https://github.com/se-passau/codeface-extraction-r/pull/67#discussion_r162930063).

### Motivation

The motivation for this enhancement is basically the proper alignment of functionality for all types of artifact vertices. If we have any information that would fit, we should provide corresponding functionality by means of new vertex-attribute functions.

### Ideas for the Implementation

Some ideas for further vertex attributes proposed in this issue:
- Mail threads
   - [x] Number of contributors to this thread
   - [x] Number of e-mails in this thread
   - [x] Start and end date
   - [x] Originating mailing list
- Issue threads
   - [x] Number of contributors to this issue
   - [x] Number of events or comments in this issue
   - [x] Start and end date
   - [x] Issue Title
   - [x] Is Pull Request
   - [x] Maybe it could also be interesting for pull requests to know whether they are merged/closed/open?

What are your thoughts?

PR #229, opened 2022-08-05
Vertex attributes for artifacts

joba00002:
Addresses #170

### Prerequisites

### DIFF
--- a/tests/test-networks-covariates.R
+++ b/tests/test-networks-covariates.R
@@ -20,6 +20,7 @@
 ## Copyright 2018-2019 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
 ## Copyright 2021 by Johannes Hostert <s8johost@stud.uni-saarland.de>
 ## Copyright 2021-2022 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
+## Copyright 2022 by Jonathan Baumann <joba00002@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -59,9 +60,11 @@ myranges.since.2010 = construct.ranges(mybins.since.2010, sliding.window = FALSE
 #' @return Tuple containing project data and list of networks
 get.network.covariates.test.networks = function(network.type = c("author", "artifact"), issues = FALSE,
                                                 author.relation = c("cochange", "issue", "mail"),
+                                                artifact.relation = c("cochange", "issue", "mail"),
                                                 bins = mybins) {
 
     author.relation = match.arg(author.relation)
+    artifact.relation = match.arg(artifact.relation)
     network.type.function = switch(match.arg(network.type),
                                    "author" = "get.author.network",
                                    "artifact" = "get.artifact.network")
@@ -72,7 +75,8 @@ get.network.covariates.test.networks = function(network.type = c("author", "arti
     proj.conf$update.value("commits.filter.untracked.files", TRUE)
     proj.conf$update.value("issues.only.comments", FALSE)
     net.conf = NetworkConf$new()
-    net.conf$update.values(list(author.relation = author.relation, simplify = FALSE))
+    net.conf$update.values(list(author.relation = author.relation, artifact.relation = artifact.relation,
+                                simplify = FALSE))
 
     ## retrieve project data
     project.data = ProjectData$new(proj.conf)
@@ -1311,6 +1315,64 @@ test_that("Test add.vertex.attribute.artifact.first.occurrence", {
     })
 })
 
+#' Test the add.vertex.attribute.artifact.last.edited method
+test_that("Test add.vertex.attribute.artifact.last.edited", {
+
+    ## Test setup
+
+    networks.and.data = get.network.covariates.test.networks("artifact")
+
+    expected.attributes = list(
+        range = network.covariates.test.build.expected(
+            c("2016-07-12 15:58:59 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        ),
+        cumulative = network.covariates.test.build.expected(
+            c("2016-07-12 15:58:59 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        ),
+        all.ranges = network.covariates.test.build.expected(
+            c("2016-07-12 16:00:45 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        ),
+        project.cumulative = network.covariates.test.build.expected(
+            c("2016-07-12 15:58:59 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        ),
+        project.all.ranges = network.covariates.test.build.expected(
+            c("2016-07-12 16:00:45 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        ),
+        complete = network.covariates.test.build.expected(
+            c("2016-07-12 16:00:45 UTC"), c("2016-07-12 16:00:45 UTC"),
+            c("2016-07-12 16:06:32 UTC", "2016-07-12 16:06:32 UTC")
+        )
+    )
+
+    ## convert date strings to POSIXct
+    expected.attributes = lapply(expected.attributes, function(times) {
+        lapply(times, function(date.vector) {
+            get.date.from.string(date.vector)
+        })
+    })
+
+    ## Test
+
+    lapply(AGGREGATION.LEVELS, function(level) {
+        networks.with.attr = add.vertex.attribute.artifact.last.edited(
+            networks.and.data[["networks"]], networks.and.data[["project.data"]],
+            aggregation.level = level
+        )
+
+        actual.attributes = lapply(networks.with.attr, igraph::get.vertex.attribute, name = "last.edited")
+
+        ## convert UNIX timestamps to POSIXct
+        actual.attributes = lapply(actual.attributes, get.date.from.unix.timestamp)
+
+        expect_equal(expected.attributes[[level]], actual.attributes)
+    })
+})
+
 #' Test the add.vertex.attribute.artifact.change.count method
 test_that("Test add.vertex.attribute.artifact.change.count", {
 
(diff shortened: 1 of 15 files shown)


util-data-misc.R
```diff
@@ -629,3 +629,42 @@ get.issue.last.activity.date = function(proj.data, type = c("all", "issues", "pu
...
+#' @param type which issue type to consider (see \code{preprocess.issue.data}).
+#'             One of \code{"issues"}, \code{"pull.requests"} or \code{"all"}
+#'             [default: "all"]
+#'
+#' @return a named list of dates, where the name is the issue ID.
+get.issue.title = function(proj.data, type = c("all", "issues", "pull.requests")) {
+    type = match.arg(type)
+    logging::logdebug("get.issue.title: starting.")
+    df = preprocess.issue.data(proj.data, type = type, retained.cols = c("issue.id", "issue.title"))
+    issue.id.to.title = get.key.to.value.from.df(df, "issue.id", "title")
```
bockthom, line 648: Does this really work? In `preprocess.issue.data` you retain column `issue.title`, but in `get.key.to.value.from.df` you ask for column `title`.

util-data-misc.R
```diff
@@ -629,3 +629,42 @@ get.issue.last.activity.date = function(proj.data, type = c("all", "issues", "pu
...
+    issue.id.to.title.only = lapply(issue.id.to.dates, function(df) {
+        df[[1,"data.vertices"]] # data frames resulting from get.key.to.value.from.df always have at least one row
+    })
```
bockthom, line 651: I don't get what this statement actually does. 
- First of all, there is no variable `issue.id.to.dates`. Should be `issue.id.to.titles`, I guess.
- Second, `df[[1, data.vertices]]` sounds strange–what actually does this and why is it needed? Maybe some more documentation helps understand what's going on here.

util-data-misc.R
```diff
@@ -629,3 +629,42 @@ get.issue.last.activity.date = function(proj.data, type = c("all", "issues", "pu
...
+#'
+#' @param proj.data the \code{ProjectData} containing the issue data
+#'
+#' @return a named list of logical values, where the name is the issue ID.
+get.issue.is.pull.request = function(proj.data) {
+    logging::logdebug("get.issue.is.pull.request: starting.")
+    issue.id.to.type = get.key.to.value.from.df(proj.data$get.issues(), "issue.id", "issue.type")
+    issue.id.to.is.pr = lapply(issue.id.to.type, function(df) {
+        "pull request" %in% unlist(df[["data.vertices"]])
+    })
```
bockthom, line 667: Would it be possible to use helper function `mask.pull.requests` here? I don't like the code duplication here. If we will change anything in the future regarding how to identify pull requests, one would need to adjust to different statements.

bockthom (commented):
Thanks for your additions to this PR @joba00002. Sorry for the late reply, we were very busy during the last weeks.
I had a quick look at your recent changes and spotted a few possible copy-paste errors or statements that need more documentation, etc., please see my individual comments below.

In addition, I checked the list of vertex attributes in #170. Could you please tick all the attributes you already have implemented? From what I have seen now, only two vertex attributes are still missing, right?
- number of comments in an issue (there is an attribute for events already, but maybe someone would like to count the comments only even if all events are configured to be considered for network construction)
- originating mailing list

Please let us know if you have further questions regarding that.

util-data-misc.R
underscore-j: Yes, it should actually be called `issue.title` in both cases.

What's weird is that I thought I tested these functions manually (I will also write tests for them), so that shouldn't have worked…

util-data-misc.R
underscore-j: `df[[1, data.vertices]]` should just be the issue title of the first row.
My thinking is that the title should be the same in all rows, so I just need one.

util-data-misc.R
underscore-j: I think I decided against that because I want a named list, not a vector.
However, I can also use `mask.pull.requests` and then add the issue IDs as names.

underscore-j:
Regarding the originating mailing list:
When we discussed this in a meeting, you said that the thread ID would contain a number that indicates the originating mailing list.
However, this does not appear to be the case, at least in the sample data.

At the moment, I still don't see any way how a thread in the project data can be linked back to its originating mailing list.

util-data-misc.R
joba00002: I thought that I had tested all these functions manually (I will also write tests for them), but clearly, I missed this one.
It should indeed be `issue.title` in both cases.

util-data-misc.R
joba00002: The list contains data frames that may have multiple rows.
In each of these, I think the name should be unique, so I just pick it from the first row.

util-data-misc.R
joba00002: Not directly, as that returns a vector and I want a named list. However, I can use it and then add the names after to avoid duplication.

joba00002:
I wanted to tick the ones I have implemented, however, GitHub doesn't seem to allow me to do so. I might not have the necessary permissions to edit that issue.

Regarding the originating mailing lists of mail threads, I thought you said in a meeting that the thread ID contains the index of the mailing list as listed in the config file.
However, at least in the sample data, this does not appear to be the case, so I currently don't see a way of linking a thread to its originating mailing list.

util-data-misc.R
```diff
@@ -447,3 +451,275 @@ get.author.issue.comment.count = function(proj.data, type = c("all", "issues", "
...
+#' @return a named list of dates, where the name is the issue ID.
+get.pr.open.merged.or.closed = function(proj.data) {
+    logging::logdebug("get.issue.title: starting.")
+    df = preprocess.issue.data(proj.data, type = "pull.requests", use.unfiltered.data = TRUE,
+                               retained.cols = c("issue.id", "issue.state", "event.name"))
+    issue.id.to.events = get.key.to.value.from.df(df, "issue.id", "event.name")
+    issue.id.to.state = lapply(issue.id.to.events, function(df) {
+        return (if ("open" %in% df[["issue.state"]]) "open"
+                else if ("merged" %in% df[["event.name"]]) "merged"
+                else "closed")
```
bockthom, line 706: The state of a GitHub issue could also be "reopened" (in theory, I don't know whether his really happens for pull requests). I guess we should treat a "reopened" PR as "open" here, shouldn't we?

util-data-misc.R
```diff
@@ -447,3 +451,275 @@ get.author.issue.comment.count = function(proj.data, type = c("all", "issues", "
...
+    return(issue.id.to.title.only)
+}
+
+#' Get whether a PR is open, has been merged, or has been closed without merging.
+#'
+#' @param proj.data the \code{ProjectData} containing the issue data
+#'
+#' @return a named list of dates, where the name is the issue ID.
+get.pr.open.merged.or.closed = function(proj.data) {
+    logging::logdebug("get.issue.title: starting.")
```
bockthom, line 699: Wrong logging message.

tests/test-networks-covariates.R
```diff
@@ -1347,6 +1410,992 @@ test_that("Test add.vertex.attribute.artifact.change.count", {
...
+        actual.attributes = lapply(actual.attributes, get.date.from.unix.timestamp)
+
+        expect_equal(expected.attributes[[level]], actual.attributes)
+    })
+})
+
+## Unit tests for issue artifact networks
+
+#' Helper function for add.vertex.attribute.issue.* tests
+merge.expected.attributes = function(expected.attributes.issues.only, expected.attributes.prs.only) {
```
bockthom, line 1606: Could you please document the parameters and return values of this helper function?
And could you please document what this helper function actually does. I don't get what get replaced and added here and why.

joba00002: To be fair, this function really is an ugly hack, but I did not want to spend too much time on it.

It takes two lists of lists of vectors (`expected.attributes.issues.only` and `expected.attributes.prs.only`), where the vectors contain `NA` whenever the corresponding vertex is not an issue/PR. I want to merge them in such a way that I get the data for issues and PRs, no `NA`s.

I should really rewrite this functions, because I don't like how this one works.

util-data-misc.R
joba00002: Yes, I did not think of this case.

util-data-misc.R
```diff
@@ -447,3 +451,275 @@ get.author.issue.comment.count = function(proj.data, type = c("all", "issues", "
...
+get.issue.event.count = function(proj.data, type = c("all", "issues", "pull.requests"), use.unfiltered.data = FALSE) {
+    type = match.arg(type)
+    logging::logdebug("get.issue.event.count: starting.")
+    df = preprocess.issue.data(proj.data, type = type, retained.cols = c("issue.id", "author.email"),
+                               use.unfiltered.data = use.unfiltered.data)
+    issue.id.to.events = get.key.to.value.from.df(df, "issue.id", "author.email")
+    issue.id.to.event.count = lapply(issue.id.to.events, function(df) {
+        length(df[["data.vertices"]])
+    })
+    logging::logdebug("get.issue.event.count: finished")
```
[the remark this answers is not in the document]
joba00002: True.
I somehow missed that `event.id` exists, so I just treated every line as its own event.

util-data-misc.R
```diff
@@ -359,13 +362,14 @@ get.author.mail.thread.count = function(proj.data) {
...
+#' @param use.unfiltered.data whether to use unfiltered issue data (see \code{preprocess.issue.data}) [default: FALSE]
 #'
 #' @return a dataframe consisting of two columns, the first of which holding the authors' names and the second holding
 #'         their respective issue counts
-get.author.issue.count = function(proj.data, type = c("all", "issues", "pull.requests")) {
+get.author.issue.count = function(proj.data, type = c("all", "issues", "pull.requests"), use.unfiltered.data = FALSE) {
     type = match.arg(type)
     logging::logdebug("get.author.issue.count: starting.")
-    df = preprocess.issue.data(proj.data, type = type)
+    df = preprocess.issue.data(proj.data, type = type, use.unfiltered.data = use.unfiltered.data)
```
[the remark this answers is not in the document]
bockthom: That sounds reasonable.
I talked with @hechtlC right now and we came to the conclusion that we would like to have this parameter, though. Just for extendability reasons: If filtering issue data would change in the future and change more, it would be good have this parameter here available (to know where something could change then). So could you just add the parameter to both functions and add a comment in the description that they should have no effect for now. As there is a default value, this parameter should not be used anywhere then. Thanks!

util-data-misc.R
```diff
@@ -447,3 +451,275 @@ get.author.issue.comment.count = function(proj.data, type = c("all", "issues", "
...
+get.issue.last.activity.date = function(proj.data, type = c("all", "issues", "pull.requests"),
+                                        use.unfiltered.data = FALSE) {
+    type = match.arg(type)
+    logging::logdebug("get.issue.last.activity.date: starting.")
+    df = preprocess.issue.data(proj.data, type = type, retained.cols = c("issue.id", "date"),
+                               use.unfiltered.data = use.unfiltered.data)
+    issue.id.to.dates = get.key.to.value.from.df(df, "issue.id", "date")
+    issue.id.to.end.date = lapply(issue.id.to.dates, function(df) {
+        max(df[["data.vertices"]])
+    })
```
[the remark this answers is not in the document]
bockthom: We decided to keep everything as it is implemented now. No need for a change.

util-data-misc.R
```diff
@@ -523,6 +523,34 @@ get.mail.thread.end.date = function(proj.data) {
...
 
+#' Get the identifier of the mailing list from which a threat originates.
+#' This identifier is part of the thread ID as produced by codeface, e.g. if the thread ID is "13#37", then 13 is the ID
+#' of the mailing list.
+#'
+#' Older versions of codeface did not include this identifier. If the identifier is not included in the data used, a
+#' warning is produced and the list will contain \code{NA} for each thread.
+#'
+#' @param proj.data the \code{ProjectData} containing the mail data
+#' @return a named list of mailing list identifiers, where the name is the thread.
```
bockthom, line 534: Please add a blank line between `@param` and `@return`. This also holds for the documentation of several other functions in this file.

util-data-misc.R
```diff
@@ -523,6 +523,34 @@ get.mail.thread.end.date = function(proj.data) {
     return(thread.to.end.date)
 }
 
+#' Get the identifier of the mailing list from which a threat originates.
+#' This identifier is part of the thread ID as produced by codeface, e.g. if the thread ID is "13#37", then 13 is the ID
```
bockthom, line 527: Please also add a comma after e.g.

util-data-misc.R
```diff
@@ -523,6 +523,34 @@ get.mail.thread.end.date = function(proj.data) {
...
+#' Older versions of codeface did not include this identifier. If the identifier is not included in the data used, a
+#' warning is produced and the list will contain \code{NA} for each thread.
+#'
+#' @param proj.data the \code{ProjectData} containing the mail data
+#' @return a named list of mailing list identifiers, where the name is the thread.
+get.mail.thread.originating.mailing.list = function(proj.data) {
+    logging::logdebug("get.mail.thread.originating.mailing.list: starting.")
+    thread.ids = unique(proj.data$get.mails()[["thread"]])
+    thread.to.list = lapply(thread.ids, function(thread.name) {
+        thread.id = substr(thread.name, 9, nchar(thread.name)-1) # remove '<thread-' '>'
```
bockthom, line 539: Please add spaces around the `-` sign.

tests/test-networks-covariates.R
```diff
@@ -1603,12 +1603,30 @@ test_that("Test add.vertex.attribute.mail.thread.end.date", {
...
+#' vertices of the other type.
+#' The data is given as lists of lists of vectors, and merged by replacing the \code{NA}-values of the first list
+#' with the non-\code{NA} values of the second. Therefore, it only works if all vertices have a non-\code{NA} value in
+#' exactly one of the two lists.
+#'
+#' @param expected.attributes.issues.only a list of lists of vectors, containing some value for each issue vertex and
+#'                                        \code{NA} for PR vertices
+#' @param expected.attributes.prs.only a list of lists of vectors, containing some value for each PR vertex and
+#'                                     \code{NA} for issue vertices
+#' @return a list of lists of vectors, containing values for issue and PR vertices
```
bockthom, line 1644: Please also add a blankl line here before `@return`.

bockthom (commented):
Thanks @joba00002 for all the efforts, looks good to me!

I just spotted a couple of spacing inconsistencies. Please fix them and add your changes to the `NEWS.md`, and then this PR is ready to merge.

bockthom (approved):

PR #231, opened 2023-01-10
Add input validation for setter and read functions

MaLoefUDS:
### Prerequisites

### Description

* Add input validation into setter functions and functions that read input from files.
* Fix check for empty input files in almost all read functions

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,20 @@
 
 # coronet – Changelog
 
+## 4.3
+
+### Added
+
+- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #208 d1d9a039f50480ec5b442dc7e8b518648d1f9d9d).
+
+### Changed/Improved
+
+- Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #208 b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
+
+### Fixed
+
+- Fix check for empty input-files in utility read functions. Compared to unpresent files, empty files do not throw an error when reading them, a check for `nrow(commit.data) < 1` is therefore required (PR #208 ecfa643cbc15975c3062af95c50ead02730b580f).
+
 ## 4.2
 
 ### Added
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,15 +6,15 @@
 
 ### Added
 
-- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #208 d1d9a039f50480ec5b442dc7e8b518648d1f9d9d).
+- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #231, d1d9a039f50480ec5b442dc7e8b518648d1f9d9d).
 
 ### Changed/Improved
 
-- Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #208 b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
+- Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
 
 ### Fixed
 
-- Fix check for empty input-files in utility read functions. Compared to unpresent files, empty files do not throw an error when reading them, a check for `nrow(commit.data) < 1` is therefore required (PR #208 ecfa643cbc15975c3062af95c50ead02730b580f).
+- Fix check for empty input-files in utility read functions. Compared to unpresent files, empty files do not throw an error when reading them, a check for `nrow(commit.data) < 1` is therefore required (PR #231, ecfa643cbc15975c3062af95c50ead02730b580f).
 
 ## 4.2
 
--- a/tests/test-misc.R
+++ b/tests/test-misc.R
@@ -14,7 +14,7 @@
 ## Copyright 2017 by Felix Prasse <prassefe@fim.uni-passau.de>
 ## Copyright 2017-2018 by Claus Hunsen <hunsen@fim.uni-passau.de>
 ## Copyright 2017-2018 by Thomas Bock <bockthom@fim.uni-passau.de>
-## Copyright 2023 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
+## Copyright 2022-2023 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -109,10 +109,10 @@ test_that("Match argument or take default.", {
 })
 
 ##
-## Check presence and datatype of column.
+## Check presence and datatype of data frame columns.
 ##
 
-test_that("Check presence and datatype of column.", {
+test_that("Check presence and datatype of data frame columns.", {
 
     user.names = c("John", "Peter", "Maria", "Susanne")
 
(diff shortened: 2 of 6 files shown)


tests/test-misc.R
```diff
@@ -14,6 +14,7 @@
 ## Copyright 2017 by Felix Prasse <prassefe@fim.uni-passau.de>
 ## Copyright 2017-2018 by Claus Hunsen <hunsen@fim.uni-passau.de>
 ## Copyright 2017-2018 by Thomas Bock <bockthom@fim.uni-passau.de>
+## Copyright 2023 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
```
bockthom, line 17: Please write `2022-2023` as you have already started to work on that in 2022. (Also holds for the other files on which you have worked in 2022 already.)

bockthom (commented):
Thanks @MaLoefUDS, this PR looks really good so far! I only spotted a few inconsistencies in comments that should be fixed, please see my individual comments below.

In addition, there is one data source we have not talked about yet: The custom event timestamps. We also should validate the inputs for its setter and read functions, although it is a little bit different than the other data sources:

`read.custom.event.timestamps`: This function reads a data.frame with two columns – one is of type character, the other one of type POSIXct. We should check this directly after reading it. However, there are no constants and no column names, as this data format is not used outside of the read function. So there should be only hardcoded checks within this function. Because at the end, the read data.frame is transferred to a named list prior to returning.

`set.custom.event.timestamps`: This function just takes a list as parameter. So we should just check whether the input is a list.

So, could you please add the validation for these two special cases?

And please don't forget to add the description for this PR (you don't need to add a changelog there).

util-read.R
```diff
@@ -925,6 +918,25 @@ read.custom.event.timestamps = function(data.path, file.name) {
...
     names(timestamps) = custom.event.timestamps.table[[1]]
 
+    ## check if first column of input has type character
+    if (class(names(timestamps)) != "character") {
+        error.message = sprintf(
+            "The first column of custom event timestamps should be of type character but is '%s'",
+            class(names(timestamps)))
+        logging::logerror(error.message)
+        stop(error.message)
+    }
```
bockthom, line 928: At this stage, the character check is superfluous: `names` always returns a character vector. And when the names are set, whatever the input is, `names` transfers it to characters. 

Therefore, we either remove this particular check (as it is always a character, we don't need to check here) - or we perform the check before calling `names` for the first time. But it could also be that the names are integers at the time of reading it, so we would also have to cast it to characters in this case. So, maybe is a better option to not check the names column at all.

util-read.R
```diff
@@ -925,6 +918,25 @@ read.custom.event.timestamps = function(data.path, file.name) {
...
+
+    ## check if second column of input is in POSIXct format by converting 
+    ## all elements to POSIXct and catch possible errors
+    tryCatch(sapply(timestamps, function(time) as.POSIXct(time)),
+            error = function(e) {
+                error.message = sprintf("The second column of custom event timestamps should be in POSIXct format")
+                logging::logerror(error.message)
+                stop(error.message)
+            })
```
bockthom, line 937: We have a function that covers the conversion to POSIXct already and does all the error handling and handling of special cases etc. 
Therefore, this `sapply` call is not necessary, as the function `get.date.from.string` does this already and is already called below when sorting the timestamps.

So, it should be enough to check at the end of this function (right after the sorting, within the if-block) whether `class(unlist(timestamps))` is `"POSIXct"` (only if the list is not empty).

bockthom (changes requested):
I just spotted two issues regarding the new checks for the custom event timestamps. Please fix them, everything else looks good to me.

(And don't worry about the failing CI pipeline on GitHub. There are some temporary problems at our CI provider that let the CI pipeline always run into a timeout before starting the tests. Just ignore the pipeline for the moment and just make sure that the tests run successfully on your end.)

util-misc.R
```diff
@@ -139,6 +140,78 @@ match.arg.or.default = function(arg, choices, default = NULL, several.ok = FALSE
...
+#'                   without regarding column types. Otherwise this vector must be of the
+#'                   same length as the vector of \code{columns}
+#'                   [default: NULL]
+verify.data.frame.columns = function(data, columns, data.types = NULL) {
+
+    ## every column of the data frame must be one to one mapped to a datatype expected in the column
+    ## therefore if there aren't as many datatypes provided in \code{data.types} as column names have
+    ## been provided in \code{columns} we can stop here already
+    if (!is.null(data.types) && length(columns) != length(data.types)) {
+        error.message = sprintf("If specified, the length of the two given vectors 'columns' and 'data.types' must match.")
```
hechtlC, line 159: This line has more than 120 allowed characters. Please split the line

hechtlC (approved):
Looks good, thanks @MaLoefUDS . I just found one line that has more than 120 characters. Once this is fixed and the annotations by @bockthom are processed, we can merge this.

util-read.R
MaLoefUDS: As discussed I will replace the original check by a check for `NA` as `get.date.from.string` will return a `NA`  on a wrongly formatted string :).

bockthom (commented):
Looks good, thanks!

(I only spotted one minor indentation inconsistency)

util-misc.R
```diff
@@ -156,7 +156,8 @@ verify.data.frame.columns = function(data, columns, data.types = NULL) {
     ## therefore if there aren't as many datatypes provided in \code{data.types} as column names have
     ## been provided in \code{columns} we can stop here already
     if (!is.null(data.types) && length(columns) != length(data.types)) {
-        error.message = sprintf("If specified, the length of the two given vectors 'columns' and 'data.types' must match.")
+        error.message = sprintf(paste("If specified, the length of the two given vectors",
+        "'columns' and 'data.types' must match."))
```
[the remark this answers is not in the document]
MaLoefUDS: Will do directly!

NEWS.md
```diff
@@ -2,6 +2,20 @@
...
+
+- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #208 d1d9a039f50480ec5b442dc7e8b518648d1f9d9d).
+
+### Changed/Improved
+
+- Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #208 b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
+
+### Fixed
+
+- Fix check for empty input-files in utility read functions. Compared to unpresent files, empty files do not throw an error when reading them, a check for `nrow(commit.data) < 1` is therefore required (PR #208 ecfa643cbc15975c3062af95c50ead02730b580f).
```
bockthom, line 17: I just noticed that the PR number is 231 and not 208. Please fix that in the news.md (and add a comma after the PR number).

MaLoefUDS: Oh I mistook it for the issue number, my bad.

bockthom (dismissed):

tests/test-misc.R
```diff
@@ -107,6 +108,73 @@ test_that("Match argument or take default.", {
...
+    age = c(42, 50, NaN, 66)
+
+    ## contains NA to verify functionality does not break
+    is.male = c(TRUE, TRUE, FALSE, NA)
+
+    ## construct simple testing dataframe
+    data.frame = data.frame(user.names, age, is.male)
+
+    ## 1) Check base functionality (benign use-case)
+    expect_no_error(verify.data.frame.columns(
```
hechtlC, line 129: Please change this to `expect_error(..., NA, message(...))`.
The `expect_no_error` function is a new feature from testthat but is still experimental. So for now we want to stay at the "old" version. Also please put a comment behind the check with `Expect no error`.

hechtlC (changes requested):
I have just tried to run the test suite on my machine since the CI is still stuck. And I noticed that you are using an experimental feature of testthat in `tests-misc.R`. See my comment.
Once this is fixed for all four occurences, I will merge the PR. Thanks again for the great work!

tests/test-misc.R
MaLoefUDS: I didn't know about this, I will do the replacement 😇

hechtlC (approved):

Issue #213, opened 2021-09-22
Handle empty user correctly


bockthom:
### Description

Currently, we remove the empty user at several places while reading data (see `util-read.R`). However, there might be locations where removing the empty user is not done yet (e.g., committer of a commit?, referenced author of an issue?, ...)

We need to check that we remove the empty user consistently. We also have to check this again after issue se-sic/codeface-extraction#44 has been fixed.

### Versions

This affects only versions >= 4.0 of `coronet`.

PR #235, opened 2023-01-24
Make removal of empty/deleted users consistent

nlschn:
The function in util-read.R now uses a vector containing the columns to check for deleted and ampty users. This is important for commits, because there, the column "committer.name" is also relevant.

Also use the function in the setter methods of util.data for mails, issues, commits, and authors right after input validation.

This addresses #213.

### Prerequisites

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 ### Changed/Improved
 
 - Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
+- Include removal of empty and deleted users in the setters of mails, commits, issues, and authors (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@
 ### Changed/Improved
 
 - Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
-- Include removal of empty and deleted users in the setters of mails, commits, issues, and authors (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
+- Include removal of empty and deleted users in the setters of mails, commits, issues, and authors.For commits, also the "committer.name" column is now checked for deleted or empty users. (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
 
 ### Fixed
 
--- a/tests/codeface-data/results/testing/test_feature/feature/commits.list
+++ b/tests/codeface-data/results/testing/test_feature/feature/commits.list
@@ -9,5 +9,5 @@
 32716;"2016-07-12 16:06:30";"Thomas";"thomas@example.org";"2016-07-12 16:06:30";"Thomas";"thomas@example.org";"d01921773fae4bed8186b0aa411d6a2f7a6626e6";1;1;0;1;"";"";"";0
 32711;"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"0a1a5c523d835459c42f33e863623138555e2526";1;1;0;1;"test2.c";"Base_Feature";"Feature";1
 32711;"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"0a1a5c523d835459c42f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
-32711;"2016-07-12 16:06:33";"Thomas";"thomas@example.org";"2016-07-12 16:06:33";"";"thomas@example.org";"0a1a5c523d835459c40f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
-32711;"2016-07-12 16:06:34";"Thomas";"thomas@example.org";"2016-07-12 16:06:34";"deleted user";"thomas@example.org";"0a1a5c523d835459c41f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
+31711;"2016-07-12 16:06:33";"Thomas";"thomas@example.org";"2016-07-12 16:06:33";"";"thomas@example.org";"0a1a5c523d835459c40f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
+30711;"2016-07-12 16:06:34";"Thomas";"thomas@example.org";"2016-07-12 16:06:34";"deleted user";"thomas@example.org";"0a1a5c523d835459c41f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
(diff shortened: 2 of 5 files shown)


util-read.R
```diff
@@ -43,15 +43,23 @@ requireNamespace("sqldf") # for SQL-selections on data.frames
...
-#'
+#' @param columns the columns in which to search for the "deleted user" and author with empty name.
+#'        The default value is \code{c("author.name")}.
+#' 
 #' @return the data frame without the rows in which the author name is "deleted user" or ""
-remove.deleted.and.empty.user = function(data) {
-    return(data[tolower(data[, "author.name"]) != "deleted user" & data["author.name"] != "", ])
+remove.deleted.and.empty.user = function(data, columns = c("author.name")) {
+    ## create a copy of the original data frame
+    df = data.frame(data)
```
hechtlC, line 57: Why exactly are you doing this instead of just working on `data` as it was in the old function?

util-read.R
```diff
@@ -43,15 +43,23 @@ requireNamespace("sqldf") # for SQL-selections on data.frames
...
+    ## loop over the given columns and remove all rows in which the author name is "deleted user" or ""
+    for (column in columns) {
```
hechtlC, line 59: Could you please add a check whether the column names in `columns` actually exist in the data?

util-read.R
```diff
@@ -43,15 +43,23 @@ requireNamespace("sqldf") # for SQL-selections on data.frames
...
 
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
-## Helper functions ---------------------------------------------------------------
+## Helper functions --------------------------------------------------------
 
 #' Remove the "deleted user" or the author with empty name "" from a data frame.
 #'
 #' @param data the data from which to remove the "deleted user" and author with empty name
-#'
+#' @param columns the columns in which to search for the "deleted user" and author with empty name.
```
hechtlC, line 51: Please change the description of the default value. We do this by adding [default: "author.name"] ath the end of the description. As a reference take a look at the function `set.project.conf` in util-data.R.

NEWS.md
```diff
@@ -11,6 +11,7 @@
 ### Changed/Improved
 
 - Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f).
+- Include removal of empty and deleted users in the setters of mails, commits, issues, and authors (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
```
hechtlC, line 14: Please add the information, that we now also check the `committer.name` for the empty and deleted users.

hechtlC (changes requested):
Thanks for the good work @nlschn !
I just have a few small remarks. Also could you please add a commit entry in the commits.list file in the test data (proximity and feature alike) where you have a "normal" user name as the author and the empty user as the committer.name? This way, if the tests still run we implicitly test that the empty user and the committer.name are filtered correctly. So no need to write a new test.

util-read.R
nlschn: Because the function now takes a vector that can hold mulitple column names and I have not found any other way than to loop through the column names and check them seperately. On the other hand, the function is called with two arguments at max right now. The check itself has remained the same.

hechtlC: This should not matter in this case. You can safely remove this line and just work on `data`. That should be fine.

util-read.R
```diff
@@ -47,14 +47,20 @@ requireNamespace("data.table") # for faster data.frame processing
...
 #'
-#' @param data the data from which to remove the "deleted user" and author with empty name
+#' @param data the data from which to remove the "deleted user" and author with empty name.
 #' @param columns the columns in which to search for the "deleted user" and author with empty name.
-#'        The default value is \code{c("author.name")}.
+#'                [default: c("author.name")]
 #' 
 #' @return the data frame without the rows in which the author name is "deleted user" or ""
 remove.deleted.and.empty.user = function(data, columns = c("author.name")) {
+    if (!all(columns %in% names(data))) {
```
hechtlC, line 56: Please use `colnames` instead of `names` here for clarity reasons.

hechtlC (dismissed):

util-data.R
```diff
@@ -1638,6 +1647,9 @@ ProjectData = R6::R6Class("ProjectData",
                 verify.data.frame.columns(data, ISSUES.LIST.COLUMNS, ISSUES.LIST.DATA.TYPES)
             }
 
+            ## remove deleted and empty users
+            data = remove.deleted.and.empty.user(data)
```
bockthom, line 1651: This only affects the active users that are the author of an issue event, but not passive users that are mentioned/subscribed/(un)assigned/review_requested/review_request_removed to an issue and named in the event info columns of the issue data.

@hechtlC and I have discussed about this whether we also should remove the empty/deleted user from the event info columns in issue data. However, for the moment, we come to the conclusion, that we should not remove it from there, for the following two reasons:
(1) If a commit is referenced in an issue, we also don't check whether the commit hash exists in our data, so why should we check if a referenced user does exist?
(2) It would make the current remove function more complex, as the `event.info.1` column contains different data, depending on the `event.type`. Therefore, we cannot just run the removement on the complete column, but would need to select only the lines which belong to a mentioned/subscribed/(un)assigned/review_requested/review_request_removed the event.

However, as not removing the empty/deleted user in this column introduces some kind of inconsistency, we should document this in the code, to be aware of that the empty/deleted user can still appear in the event info columns.

So, @nlschn could you please add a comment here in the setter function as well as also in the corresponding read function, which makes clear that the empty/deleted user is not removed from the event info columns, or to be more precise, the corresponding event is not removed even if the user in the event info column is empty or deleted.

bockthom (changes requested):
Looks also good to me.
I just have spotted one mistake in the newly added test data, and I've added one more comment regarding another inconsistency with regard to removing the empty/deleted user that is still present, please see them below.

tests/codeface-data/results/testing/test_feature/feature/commits.list
```diff
@@ -9,5 +9,5 @@
...
 32711;"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"0a1a5c523d835459c42f33e863623138555e2526";1;1;0;1;"test2.c";"Base_Feature";"Feature";1
 32711;"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"2016-07-12 16:06:32";"Thomas";"thomas@example.org";"0a1a5c523d835459c42f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
-32711;"2016-07-12 16:06:33";"Thomas";"thomas@example.org";"2016-07-12 16:06:33";"";"thomas@example.org";"0a1a5c523d835459c40f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
-32711;"2016-07-12 16:06:34";"Thomas";"thomas@example.org";"2016-07-12 16:06:34";"deleted user";"thomas@example.org";"0a1a5c523d835459c41f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
+31711;"2016-07-12 16:06:33";"Thomas";"thomas@example.org";"2016-07-12 16:06:33";"";"thomas@example.org";"0a1a5c523d835459c40f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
+30711;"2016-07-12 16:06:34";"Thomas";"thomas@example.org";"2016-07-12 16:06:34";"deleted user";"thomas@example.org";"0a1a5c523d835459c41f33e863623138555e2526";1;1;0;1;"test2.c";"foo";"Feature";1
```
bockthom, line 13: Unfortunately, just changing the commit id is not enough, as also the commit hash should be unique. So, please also generate a new commit hash for these two new commits. Thanks!

bockthom: And please also keep in mind: The commit ids differ between "feature" and "proximity" data (as you did correctly), but both "feature" and "proximity" use the same commit hash :wink: 
Even if it may not be intuitive, there is a reason: the commit id directly comes from the database, where one commit (having a specific commit hash) is stored in both "feature" and "proximity" data.

bockthom (approved):
Looks good now!

PR #237, opened 2023-03-18
Fix R warnings and replace deprecated R function calls

bockthom:
### Prerequisites

### Description

In this PR, the deprecated function call `default.stringsAsFactors()` gets replaced, and some warnings are fixed that are caused by vectorized `is.na` calls when configuration parameters have a vector of values. 

### Changelog

1. As of R version 4.2.0, the function call `default.stringsAsFactors()` is deprecated. In the near future (probably with R version 4.3.0), this call will become defunct. Therefore, we replace the single call to this function in coronet by `FALSE`, which is also the current default value of the global option `stringsAsFactors`.

2. Currently, we get a couple of warnings when updating configuration values that consist of more than one values, for example in the following case:
The default value of the configuration option of "issues.from.source" is the following vector: `c("github", "jira")`. Then, the following statements leads to a couple of warnings:
`project.configuration$update.value("issues.from.source", "github")`
_Warnings:_
`length(x) = 2 > 1' in coercion to 'logical(1)`
These warnings are caused by the default value of "issues.from.source" in several if conditions, as 
`if (is.na(private[["attributes"]][[name]][["default"]]))` returns more than one boolean, actually two: one for "github" and one for "jira". But this is not expected here, as the `is.na` check here should just find out whether the overall default value is `NA` and not for all elements of a vector whether they are `NA`. To fix this, we need to make sure that the `is.na` check is only performed when the length of the variable to check is exactly 1. As this check is needed multiple times, and since we also need its negated version, I have added an additional helper function `is.single.na`, that checks whether an element is of length 1, and if so, returns the result of `is.na`, otherwise returns `FALSE`. This makes sure that we always have a single boolean value in this condition, no matter whether the input is a vector of multiple elements or not.
This prevents the warning being printed. As the the warning will be turned into an error in future R versions, this fix also prevents errors that could be caused by this problem.

3. When accessing a data.frame column via `[` instead of `[[`, this could lead to the warning 
`In xtfrm.data.frame(x) : cannot xtfrm data frames` 
in certain situations. This was the case for the patchstack mail filtering in function `filter.patchstack.mails` in `util-data.R`. Using `[[` instead of `[` prevents this warning.

### Disclaimer
~~The CI pipeline will potentially fail for this PR now, not because of the changes in this PR, but because of the issues mentioned in se-sic#236 (caused by igraph version 1.4.0 and later). The changes of this PR here should be totally unrelated to these issues. The tests pass with the changes here when running them locally with igraph version 1.3.5.~~ *(this has already been fixed in #240 in the meantime)*

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,18 +6,22 @@
 
 ### Added
 
-- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #231, d1d9a039f50480ec5b442dc7e8b518648d1f9d9d).
-- Add CI support for GitHub Actions (PR #234, fa1fc4af65751402ae6b23298dd4ed821930c6d2).
+- Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #231, d1d9a039f50480ec5b442dc7e8b518648d1f9d9d)
+- Add helper function `is.single.na` to check whether an element is of length 1 and is `NA` (ddff2b8bbca6405f5c7c1cf4e7e97374fb1426ca, ccfc2d12a68dfa412f05159e8e3b03118694e748)
+- Add CI support for GitHub Actions (PR #234, fa1fc4af65751402ae6b23298dd4ed821930c6d2)
 
 ### Changed/Improved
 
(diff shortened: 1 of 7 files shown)


hechtlC (approved):
Looks good thanks for the work @bockthom !

PR #238, opened 2023-03-23
Remove 'artifact' attribute from edges of artifact networks, append 'author.name' instead

MaLoefUDS:
### Prerequisites

### Description

As artifacts should not be able to create edges between artifacts there should not be a `artifact` attribute on edges in artifact networks. Authors can cause edges in artifact networks so `author.name` is the replacement for `artifact` in that case.

### Changelog

Add argument to `construct.edge.list.from.key.value.list` function which differentiates if constructed edges are supposed to be artifact edges, in which case we check if the `artifact` attribute is present for edges and replace it by `author.name`.

Fix tests accordingly.


tests/README.html
bockthom, line 1: Where does this file come from and why did you push it? I would like to remove it completely from the history, so maybe rebase the last commit such that this file does not appear in this PR any more.

bockthom (changes requested):
Thanks @MaLoefUDS. Looks good, I only have a view minor remarks. Could you please fix them?

---
In addition, I wonder whether we have similar problems for other kinds of artifacts (there are not only artifact networks using cochange relation, but also artifact networks using other relations, see functions `get.artifact.network.*` and, hopefully, corresponding tests). Could you please check whether we have similar problems (duplicated edges or meaningless "artifact" attributes) also for these networks? Then we can discuss in our next meeting whether we need to fix this also or whether this is not necessary for the other relations.

tests/test-networks-multi-relation.R
```diff
@@ -262,48 +262,49 @@ test_that("Construction of the multi network for the feature artifact with autho
...
+                                 "0a1a5c523d835459c42f33e863623138555e2526", # artifact cochange
                                  "72c8dd25d3dd6d18f46e2b26a5f5b1e2e8dc28d0", "5a5ec9675e98187e1e92561e1888aa6f04faa338", # bipartite cochange
                                  "3a0ed78458b3976243db6829f63eba3eead26774", "1143db502761379c2bfcecc2007fc34282e7ee61",
                                  "0a1a5c523d835459c42f33e863623138555e2526", "0a1a5c523d835459c42f33e863623138555e2526",
                                  rep(NA, 21)),                                                                           # bipartite issue
                         file = c("test.c", "test.c", "test2.c", "test3.c", "test2.c", "test2.c", "test3.c", "test2.c", # author cochange
                                  NA, NA, NA, NA,
-                                 "test2.c", "test2.c",                                                                 # artifact cochange
+                                 "test2.c",                                                                             # artifact cochange
```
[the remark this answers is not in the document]
MaLoefUDS: Oh, I inserted whitespace to keep the `# artifact cochange` somewhat aligned to the other descriptors in the lines above and below ^^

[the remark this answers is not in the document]
bockthom: Ah, I see. I did not recognize from the diff view in GitHub. (It looked like the white spaces where at the end of the line.) So, then it is fine to keep the white spaces here.

util-networks.R
```diff
@@ -970,7 +975,7 @@ construct.edge.list.from.key.value.list = function(list, network.conf, directed
     keys = names(list)
     keys.number = length(list)
 
-    if (respect.temporal.order) {
+    if (respect.temporal.order || artifact.edges) {
```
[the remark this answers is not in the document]
MaLoefUDS: It's certainly possible to move the. `artifact` - `author.name` replacement before the selection of algorithms. I've just tried it and still, all tests seem to be running. I've additionally checked if now maybe we could use the other algorithm again, but this was not possible.

bockthom:
Just to document our thoughts regarding the other types of artifact networks: The artifact networks that coronet creates based on relation `mail` or `issue` don't have edges, at the moment. Callgraph networks are not related to the other edge-construction algorithms at all. Hence, in this PR, we only need to adjust the construction of the `cochange` artifact networks.

@MaLoefUDS Could you please update the README.md and add a note on the artifact networks on `issues`, stating that we don't have edges there (same note as for `mail`)? Thanks!

tests/README.html
MaLoefUDS: Despite some efforts I was not able to delete the file completely. I tried:
```
$ git rebase <commit-before-by-pr>
... put an 'edit' to the commit which introduced README.html ...
$ git rm README.html
$ git commit --amend
$ git rebase --continue
$ git pull
``` 
After these steps the file did still exist in my git history and im not sure how to continue.

MaLoefUDS: Oh I pulled because it told me I was commits behind 😅. I’m not very familiar with rebasing in general, I‘ll try force pushing 👍

bockthom: By pulling, you "re-added" the old commits you had already pushed before. So there might be some duplicated commits in your history now and maybe also a merge commit. Please make sure to fix this and push a linear commit history. This way, you will get even more experience with rebases :wink:

MaLoefUDS: I think I just did that ^^. However I think one commit hash in NEWS needs to be updated.

bockthom: Seems to be true^^. When you update the NEWS, please also fix the typo your introduced in your statement regarding the README.

bockthom:
I finally managed to fix the issues regarding the failing test pipeline that were caused by igraph version 1.4.0 (see PR #240).

@MaLoefUDS: Could you please check whether your changes still work with the changes I made in #240? (Please try it only locally, on a separate local branch; there will probably some rebases before everything is merged into dev)

bockthom (approved):

hechtlC (approved):
Looks good thanks @MaLoefUDS ! I will merge it now.

Issue #236, opened 2023-03-18
Test suite fails when using "igraph" versions >= 1.4.0


bockthom:
In igraph version 1.4.0, they have introduced a breaking change regarding attribute types (see here: https://github.com/igraph/rigraph/blob/main/NEWS.md#igraph-140). **_This breaking change in igraph 1.4.0 breaks our test suite seriously, hundreds of tests fail._**

igraph 1.4.0 was released on 22nd of February 2023. There are multiple reasons how the recently introduced breaking changes of igraph 1.4.0 affect our test suite. For the moment, I just looked at one failing test and spotted two of these reasons, but there can also be more than these two, of course:

---

**(1) Using the `AsIs` class to generate the expected test output lets tests fail**

igraph now respects all types and classes (which have been ignored before). That is, creating data.frames containing `I(...)` calls and creating networks from the data.frame via `graph.data.frame` now creates networks that are different than the networks coronet creates, due the `AsIs` class introduced by `I(...)` which has been ignored in previous igraph versions.
However, this can be easily fixed by calling `unclass` for each data.frame column that has been created via `I(...)`.

Example: Consider the following test for author networks which uses `I(...)` to create the expected data.frame.
https://github.com/se-sic/coronet/blob/b7db5cdc1b0cab76849c78b333b6cb69dfa66669/tests/test-networks-author.R#L425-L443

For each of the columns that uses `I(...)` we just need to call `unclass` afterwards. So, in the example above, we need to add the following lines below the data.frame creation:

```R
 data[["date"]] = unclass(data[["date"]])
 data[["artifact.type"]] = unclass(data[["artifact.type"]])
 data[["hash"]] = unclass(data[["hash"]])
 data[["file"]] = unclass(data[["file"]])
 data[["artifact"]] = unclass(data[["artifact"]])
```
This fix does not affect previous versions of igraph. Hence, for all igraph versions, this problem can be fixed by just adding these `unclass` calls where applicable.

---

**(2) igraph now respects types in edge attributes**

[...]

PR #240, opened 2023-04-02
Thomas fixes for igraph 1.4.0

bockthom:
### Prerequisites

### Description

This pull requests contains fixes that are necessary due to a breaking change in igraph as of igraph version 1.4.0 (see #236).

All the fixes in this PR are implemented in such a way that coronet works with both, igraph versions < 1.4.0 and igraph versions >= 1.4.0. I have run the complete test suite with igraph version 1.3.5 and with igraph version 1.4.1, and the test suite succeeds for both igraphs versions.

### Changelog

- Fix the class of date attributes in a test for a simplified author network
- Fix the classes of the default values of vertex and edge attributes (this is important when adding attributes to an empty network)
- Fix the order of edge and vertex attributes (this is relevant when an attribute is added which was already present before, as from igraph version 1.4.0 on, the already present attribute is not re-added, and so the expected order would not be respected any more)
- Fix the class of the date attribute in multi networks (as `igraph::disjoint_union` contains a bug that converts the date attributes back to numeric)
- Fix edge attributes of bipartite edges (previously, we also added attributes for non-existent edges; now we check which edges are present and only add the corresponding attributes)


hechtlC (approved):
Looks good thanks for the fix @bockthom !

PR #244, opened 2023-09-05
Introduce issue-based artifact-network creation, new tests and fix current tests + minor bug fixes

MaLoefUDS:
### Prerequisites

#### Added

- Add issue-based artifact-networks
- Add a new `split.data.by.bins` function (not to be confused with a previously existing function that had the same name and was renamed in this context), which splits data based on given activity-based bins

#### Changed/Improved

- Enhance testing data by adding `add_link` and `referenced_by` issue events which connect issues to form edges in issue-based artifact-networks
- Add input validation for the `bins` parameter in `split.data.time.based` and `split.data.by.bins`
- Rename `split.data.by.bins` into `split.dataframe.by.bins` as this it what it does

#### Fixed

- Reformat `event.info.1` column of issue data according to the <issue-%source-%id> format, if the content of the `event.info.1` field references another issue
- Fix an issue in activity-based splitting where elements close to the border of bins might be assigned to the wrong bin. The issue was caused by the usage of `split.data.time.based` inside `split.data.activity.based` to split data into the previously derived bins, when elements close to bin borders share the same timestamps. It is fixed by replacing `split.data.time.based` by `split.data.by.bins`
- Remove the last range when using a sliding-window approach and the last range's elements are fully contained in the second last range
- Rename vertex attribute `IssueEvent` to `Issue` in multi-networks, to be consistent with bipartite-networks


util-split.R
```diff
@@ -63,6 +64,52 @@ requireNamespace("lubridate") # for date conversion
...
                                  number.windows = NULL, split.basis = c("commits", "mails", "issues"),
                                  sliding.window = FALSE, project.conf.new = NULL) {
+    split = split.data.by.time.or.bins(project.data, splitting.length = time.period, bins, split.by.time = TRUE,
+                                        number.windows, split.basis, sliding.window, project.conf.new)
+    return(split)
+}
+
+#' Split project data in activity-bin-based ranges as specified
+#'
+#' @param project.data the *Data object from which the data is retrieved
```
bockthom, line 74: Please write " project data" instead of "*Data" - RangeData is a subclass of ProjectData, and what we want here is any object of type ProjectData, regardless of being a subclass of it or not.

util-split.R
```diff
@@ -63,6 +64,52 @@ requireNamespace("lubridate") # for date conversion
...
+#' @param activity.amount the amount of data elements with unique ids to be considered in a bin, an integer.
+#' @param bins the date objects defining the start of ranges (the last date defines the end of the last range, in an
+#'             *exclusive* manner), augmented with a bin vector mapping unique ids to bins.
+#'             [default: NULL]
```
bockthom, line 78: This default value does not occur in the function declaration.

util-split.R
```diff
@@ -63,6 +64,52 @@ requireNamespace("lubridate") # for date conversion
...
+    split = split.data.by.time.or.bins(project.data, activity.amount, bins, split.by.time = FALSE,
+                                        sliding.window = sliding.window, split.basis = split.basis)
+    return(split)
+}
+
+#' Split project data in time-based or activity-bin-based ranges as specified
+#'
+#' @param project.data the *Data object from which the data is retrieved
+#' @param splitting.length either \code{time.period} from \code{split.data.time.based}
+#'                         or \code{splitting.length} from\code{split.data.by.bins.vector}
```
bockthom, line 95: `split.data.by.bins.vector` does not have a parameter `splitting.length`. Do you mean `activity.amount` instead, or something else?

util-split.R
```diff
@@ -63,6 +64,52 @@ requireNamespace("lubridate") # for date conversion
...
+#' @param bins either \code{bins} from \code{split.data.time.based}
+#'             or \code{bins} from\code{split.data.by.bins.vector}
+#' @param split.by.time logical indicating whether splitting is done time-based or by activity-bins-based,
```
bockthom, line 98: Please remove the "by" in the description and the comma at the end of the parameter description

util-split.R
```diff
@@ -129,7 +182,7 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
     bins.ranges = construct.ranges(bins)
     names(bins.ranges) = bins.ranges
 
-    if ((length(bins.ranges) <= 1) && sliding.window) {
+    if (split.by.time && (length(bins.ranges) <= 1) && sliding.window) {
```
bockthom, line 185: Why does this warning only hold for `split.by.time`? Shouldn't it be the same warning for activity-bin-based splitting?

util-split.R
```diff
@@ -140,13 +193,16 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
 
-    if (!sliding.window) {
+    if (!sliding.window || !split.by.time) {
         ## split data
         data.split = parallel::mclapply(data.to.split, function(df.name) {
             logging::logdebug("Splitting %s.", df.name)
             ## identify bins for data
             df = data[[df.name]]
-            df.bins = findInterval(df[["date"]], bins.date, all.inside = FALSE)
+            df.bins = if (!split.by.time && (df.name == split.basis))
```
bockthom, line 202: What does the check `df.name == split.basis` do? I am a little bit confused here. Maybe a comment would be helpful here.

util-split.R
```diff
@@ -1102,13 +1165,18 @@ split.get.bins.time.based = function(dates, time.period, number.windows = NULL)
...
+#'                              in the creation of the bins. This should! not change bin borders
+#'                              as entries with the same id should! share the same \code{date} attribute.
+#'                              [default: FALSE]
 #'
 #' @return a list,
-#'         the item 'vector': the bins each row in 'df' belongs to (increasing integers),
+#'         the item 'vector': the bins each row in 'df' belongs to (increasing integers),q
 #'         the item 'bins': the bin labels,  described by dates, each bin containing
-#'         'acitivity.amount' many unique items; each item in the vector indicates
+#'         'activity.amount' many unique items; each item in the vector indicates
```
bockthom, line 1176: You fixed one typo and added one - please remove the "q" here.

bockthom (changes requested):
Thanks @MaLoefUDS for your work on this PR and fixing all the bugs you encountered during adjusting all the test data. Your changes look good to me, I only have a couple of questions regarding some implementation details, some of which need further investigations prior to discussing it - some my comments below.

Notice: I did not have a look at the adjusted tests yet, and I also will need another pass on the changes to the splitting functionality to be able to completely understand all of them...

util-networks.R
```diff
@@ -475,17 +475,73 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+            # edges in artifact networks can not have the 'artifact' attribute but should instead have
+            # the 'author.name' attribute as events caused by authors connect issues
+            edge.attributes = private$network.conf$get.value("edge.attributes")
+            artifact.index = match("artifact", edge.attributes, nomatch = NA)
+            if (!is.na(artifact.index)) {
+                edge.attributes = edge.attributes[-artifact.index]
+                edge.attributes = c(edge.attributes, c("author.name"))
+            }
+
+            ## connect the current item to all previous ones
```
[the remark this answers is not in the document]
MaLoefUDS: The comment was taken from the `construct.edge.list.from.key.value.list` method line 1094 which this edge construction was based on. Should I make it more descriptive in both?

util-networks.R
```diff
@@ -475,17 +475,73 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+                cols.which = edge.attributes %in% colnames(from)
+                edge.attrs = from[ , edge.attributes[cols.which], drop = FALSE]
+
+                ## construct edge
+                to = subset(referenced.bys,
+                            event.info.1 == from[["issue.id"]] & 
+                            author.name == from[["author.name"]] & 
+                            date == from[["date"]])
+                if (!all(is.na(to))) {
+                    combination = list("Var1" = from[["issue.id"]], "Var2" = to[["issue.id"]])
```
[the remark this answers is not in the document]
MaLoefUDS: Because that's the internal names for the elements in the I presume igraph data structures I found while debugging. Its more or less arbitrary

util-split.R
MaLoefUDS: Yes, my bad!

util-split.R
MaLoefUDS: Oh yeah this is most likely the result of me mechanically adding the `split.by.time` bool to every condition which was in the original function but which I removed for simplicity from the activity-vector based splitting.

util-networks.R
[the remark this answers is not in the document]
bockthom: In line 1094 the comment is fine, as there we actually have a temporal order of activities where it is clear what "current" and "previous" means. However, here we don't have a temporal order, we just link to issues. Therefore "current" and "previous" does not really apply here.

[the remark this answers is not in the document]
MaLoefUDS: I just assumed it was meant for a previously assigned and a currently assigned variable, I did not pay attention to the temporal order. In this case, sure I will replace the comment

util-networks.R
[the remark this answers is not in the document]
bockthom: Thanks for the explanation. If this are exactly the names that are used in the other network-construction functions, then it is fine, to stay consistent. Otherwise, it would be more readable to use appropriate names here.
@hechtlC What do you think?

util-networks.R
```diff
@@ -475,17 +475,73 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+                ## get edge attributes
+                cols.which = edge.attributes %in% colnames(from)
+                edge.attrs = from[ , edge.attributes[cols.which], drop = FALSE]
+
+                ## construct edge
+                to = subset(referenced.bys,
+                            event.info.1 == from[["issue.id"]] & 
+                            author.name == from[["author.name"]] & 
+                            date == from[["date"]])
+                if (!all(is.na(to))) {
```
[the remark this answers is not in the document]
MaLoefUDS: Regarding your first paragraph: I could take an arbitrary `add_link` or `referenced_by` event and look in the `event.info.1` column to see the issue which is linked, but for the edge construction im not connecting directly to an issue but to an issue-event, this is how its done for the other artifact networks as well. So I take the corresponding other event to connect to. If there is none the input data is probably malformated.

I could add an epsilon time in which these events still connect like 1 second

util-networks.R
```diff
@@ -475,17 +475,73 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+                ),
+                edges = edge.list
+            )
+
+            ## construct network from obtained data
+            artifacts.net = construct.network.from.edge.list(
+                artifacts.net.data[["vertices"]],
+                artifacts.net.data[["edges"]],
+                network.conf = private$network.conf,
+                directed = FALSE,
```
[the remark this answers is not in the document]
MaLoefUDS: I think we can take the parameter from the config, and it will not change anything. I think we decided on the order of `add_link` -> `referenced_by` for edges and that's the way I implemented it, so it should not matter indeed.

util-networks.R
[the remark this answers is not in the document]
bockthom: > If there is none the input data is probably malformated.
> 
> I could add an epsilon time in which these events still connect like 1 second

Let's perform some empirical investigations on real-world data before we decide on each of these two concerns. I don't trust in the completeness of the data in such a way as you do :wink: You will receive an e-mail with real-world data soon...

util-networks.R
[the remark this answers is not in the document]
bockthom: Great, that sounds good. Nevertheless, let's also wait for the outcomes of our investigations on real-world data mentioned above.

util-split.R
MaLoefUDS: So you do the splitting based on one specific activity, like mails for example. So you compute bins based on the mails data and a corresponding bins vector (e.g., 1 1 1 2 2 2 3 telling you that the first 3 elements of mail data should go in bin 1, then the next 3 into bin 2 and the last into bin 3). But you can only use this vector for the mail data now, not for all the other data. That is why you need to have a distinction here

bockthom: Thanks for the explanation. Now I understand what's going on there. I think we don't need a comment there then. But I will do another review on the splitting changes anyway. So, let's wait whether there are additional questions when I try to comprehend the other changes.

tests/test-networks-artifact.R
```diff
@@ -101,3 +101,91 @@ test_that("Network construction of the undirected artifact-cochange network", {
...
+                                   "<issue-github-1>",
+                                   "<issue-github-3>",
+                                   "<issue-github-4>",
+                                   "<issue-jira-ZEPPELIN-332>",
+                                   "<issue-github-6>"),
+                          kind = "Issue",
+                          type = TYPE.ARTIFACT)
+    ## 2) edges
+    edges = data.frame(
+        from = character(), to = character(), date = as.POSIXct(character(0), tz = "UTC"), artifact.type = character(),
```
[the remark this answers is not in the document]
MaLoefUDS: Your suggested change does not break any tests, so I suppose it suffices.

util-networks.R
```diff
@@ -475,17 +475,73 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+            # the 'author.name' attribute as events caused by authors connect issues
+            edge.attributes = private$network.conf$get.value("edge.attributes")
+            artifact.index = match("artifact", edge.attributes, nomatch = NA)
+            if (!is.na(artifact.index)) {
+                edge.attributes = edge.attributes[-artifact.index]
+                edge.attributes = c(edge.attributes, c("author.name"))
+            }
+
+            ## connect the current item to all previous ones
+            for (issue.no in seq_len(nrow(add.links))) {
```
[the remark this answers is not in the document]
MaLoefUDS: I could make it work like this:

```
edge.list.raw = parallel::mclapply(split(add.links, 1:nrow(add.links)), function(from) {
...
})
edge.list = do.call(rbind, edge.list.raw)
```

Is this better performance wise?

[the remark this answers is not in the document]
bockthom: This should already be a performance improvement, yes. But if you use `plyr::rbind.fill` instead of `do.call`, the performance improvement would be even bigger :wink: And you also don't need to use two variables, you can just rbind the elements in-place.

And I suggest to use `seq_along` instead of `1:nrow`, as `seq_along` is fail-safe.

util-networks.R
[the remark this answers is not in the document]
bockthom: It turned out that we automatically add the `referenced.by` events for all `add.link` events that have `event.info.2 == "issue"` during in the codeface-extraction:
https://github.com/se-sic/codeface-extraction/blob/7bd8d0080055ed7266177fac0a28e3aeec4cac6e/issue_processing/issue_processing.py#L331-L339 

Therefore, also the dates are also exactly the same, so we can assume for sure that the corresponding `add.link` and `referenced.by` events are always present and have the same date.

To make this consistent for GitHub and JIRA issues, we also need to automatically add the `referenced_by` events for the JIRA issue extraction in https://github.com/se-sic/codeface-extraction/blob/master/issue_processing/jira_issue_processing.py" 
If we do so, the implementation here in this PR should work without any changes.

util-networks.R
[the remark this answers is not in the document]
bockthom: When we take the `directed` option from the network configuration, we also should have a test for the directed issue artifact network.

util-read.R
```diff
@@ -388,6 +388,13 @@ read.issues = function(data.path, issues.sources = c("jira", "github")) {
     issue.data[["creation.date"]] = get.date.from.string(issue.data[["creation.date"]])
     issue.data[["closing.date"]] = get.date.from.string(issue.data[["closing.date"]])
 
+    ## if other issues are referenced convert names to ID format
+    matches = issue.data[issue.data[["event.name"]] %in% c("add_link", "remove_link", "referenced_by") &
+                         issue.data[["event.info.2"]] == "issue", ]
+    formatted.matches = sprintf("<issue-%s-%s>", matches[["issue.source"]], matches[["event.info.1"]])
```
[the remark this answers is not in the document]
bockthom: Just a reminder for later: Fixing the format of `event.info.1` when it contains issue ids should also be mentioned in the changelog.

util-split.R
```diff
@@ -99,26 +146,32 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
+        if (split.by.time) {
+            split.basis = NULL
+            split.by.bins = TRUE
+            sliding.window = FALSE
+            bins = get.date.from.string(bins)
+            bins = get.date.string(bins)
+        } else {
+            bins.vector = bins[["vector"]]
+            bins = bins[["bins"]]
+        }
```
bockthom, line 173: Why is `split.by.bins` not set to `TRUE` in the `else` case? If this is not a bug, then we should comment on this - either here - or maybe also already in the comment `initiate variable` as it is not completely clear what the variable `split.by.bins` is used for in this method.

util-split.R
```diff
@@ -192,10 +248,10 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
         ## perform different steps for sliding-window approach
```
bockthom, line 248: The comment here seems to be outdated. Shouldn't it be "for sliding-window approach of time-based splitting" ? Because sliding-window for activity-based splitting should already be covered by the previous if-case...

util-split.R
```diff
@@ -214,7 +270,7 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
 
     ## add splitting information to project configuration
     project.conf.new$set.splitting.info(
-        type = "time-based",
+        type = if (split.by.time) "time-based" else "activity-based",
```
bockthom, line 273: (Not related to this PR, but still interesting): Do we set the type for the splitting info also in other functions? Maybe it would be an idea to introduce string constants for the types, if there are just these two. (Does not need to be done in this PR, we can open another issue for that in take of this later. So just having a look whether these types appear also somewhere else be enough for now....

util-split.R
```diff
@@ -63,6 +64,52 @@ requireNamespace("lubridate") # for date conversion
...
+#'
+#' @param project.data the *Data object from which the data is retrieved
+#' @param splitting.length either \code{time.period} from \code{split.data.time.based}
+#'                         or \code{splitting.length} from\code{split.data.by.bins.vector}
+#' @param bins either \code{bins} from \code{split.data.time.based}
+#'             or \code{bins} from\code{split.data.by.bins.vector}
+#' @param split.by.time logical indicating whether splitting is done time-based or by activity-bins-based,
+#' @param number.windows see \code{number.windows} from \code{split.data.time.by.bins.vector}
+#'                       [default: NULL]
+#' @param split.basis the data name to use as the basis for split bins, either 'commits', 'mails', or 'issues'
```
bockthom, line 101: "data name" --> "data source"
Maybe we can also add a reference to "activity.amount" for activity-based splitting. This is also what I would like to briefly discuss in the next meeting, to make the initial sentence of this function description a little bit more precise.

util-split.R
```diff
@@ -435,7 +498,7 @@ split.data.activity.based = function(project.data, activity.type = c("commits",
         ## and the data of the last regular range is contained in the last sliding-window range, then:
         ## remove the last regular range as it is not complete and we don't loose data when removing it
         last.regular.range = cf.data[[length(cf.data)]]
-        last.sliding.range = cf.data[[length(cf.data) - 1]]
+        last.sliding.range = cf.data.sliding[[length(cf.data.sliding) - 1]]
```
bockthom, line 501: Wait. If I am not mistaken, the comment specifies another scenario than the code here.

What the comment describes (each number represents one activity in the respective range):
regular: 444455
sliding:  334444
Then regular range 5 can be removed.

What the new code describes (each number represents one activity in the respective range):
regular: 445555
sliding:  333344
Then sliding range 4 can be removed.

Don't we actually need to handle both cases? Maybe the implementation below takes care of both scenarios (I did not check for that), but currently code and comment seem to be contradicting...

util-split.R
```diff
@@ -1120,11 +1188,23 @@ split.get.bins.activity.based = function(df, id, activity.amount, remove.duplica
...
     )
+
+    ## pad bins with entries for all duplicate ids
+    if (include.duplicate.ids) {
+        bins.activity.padded = c()
+        for (i in seq_along(ids)) {
+            ## create an extra entry for every duplicate id in the same bin as
+            ## the first occurance of the id
+            current.bin = bins.activity[ which(ids.unique == ids[i]) ]
+            bins.activity.padded = c(bins.activity.padded, current.bin)
```
bockthom, line 1199: Can we add an example (as comment) describing in which situation this is/could be necessary? Currently I understand what's going on, but don't completely get in which situation this was a problem in the previous implementation?
Maybe let's discuss this first. I would also like to know in which situations of the same method call we don't need this padding and in which we do.

bockthom (commented):
As promised, I had a closer look at the changed splitting functionality, and I also started to have a look at the tests (these were the two parts that were still missing from my last review.)

For the splitting functionality, I have a couple of questions that I would like to discuss in our next meeting to get a better understanding of what's going on there. Please have a look at my individual comments (most of them address just comments, but we should discuss them though). In general, the new functionality looks good to me, but I would like to make sure that I understand everything correctly and that we consider special corner cases correctly.

For the tests, I decided to skip reviewing the tests for now as I identified an unexpected change in the test data - we should discuss this in our next meeting before I will have a closer look at the tests themselves. Again, see my detailed comments below.

util-split.R
MaLoefUDS: I have a small problem with what we discussed in our most recent meeting. There we mentioned to include a comment which explains why we need the padded bin vector here. 
I reviewed the code again and the place where the padding is performed is inside a helper function so it would be strange to mention the use case for a function calling this helper function inside the helper function I think. I think we could add the comment where the helper function is called with the padding (as specified in the function parameters) as this is the only place where the padding is needed as of now 🤷‍♂️.

bockthom: Agreed. Please add the comment where the function is called with the respective parameter.

util-split.R
MaLoefUDS: I have reviewed this case again and as I suspected both edge cases are already covered. 

1) This case can be achieved if we have 18 elements and an activity amount of 4. In this case we calculate offset.start as floor(activity.amount / 2) = 2, offset end as (items.unique.count - offset.start) %% activity.amount = 0 and realize that the data will not be cut off from the end. This means that the last sliding range will end really at the end of the 334444 and the last regular range really ends at the end of 444455 anyways. Now the endpoints are equal and we run into the case which you highlighted here in the PR and remove the last regular range (the 55).

2) Now we have 20 elements and an activity amount of 4. offset.start is still 2 and offset.end = (items.unique.count - offset.start) %% activity.amount = 2. This means that the last 2 elements will be cut off before calculating sliding windows, and the 55 will be removed :).

bockthom: Sounds good, thanks a lot for checking that. But then we should think about adjusting the comments:

Shouldn't we move the comments in lines 497-499 without any changes down to line 504/507?
And in addition keep a copy of the comments in lines 497-499 in lines 497-499 in which we replace all occurrences of "regular" and "sliding window" by each other?

MaLoefUDS: About moving the comments I dont have any complaints because line 504-507 really are more fitting. Im wondering why we should leave a copy of the comment with sliding window and regular changed there as line 500-501 have nothing to do with removing the last sliding window (as suggested).

util-split.R
MaLoefUDS: They are only set in one place (except in tests where expected results are built). I dont think its worth creating constants for them *yet*.

util-split.R
bockthom: > Im wondering why we should leave a copy of the comment with sliding window and regular changed there as line 500-501 have nothing to do with removing the last sliding window (as suggested).

What does this line do then?
```R
last.sliding.range = cf.data.sliding[[length(cf.data.sliding) - 1]]
```
It extracts the second-last sliding range as "last sliding range". This means, the actual last one is not considered to be the last one. So the last one is kind of "removed" here. I agree that here no actual removement takes place, as no element is removed from the list, but something similar happens there...

So, I would like to have a comment describing what the cited line is used for (when not removing the last sliding range). And I would like to describe somewhere in the code where the following scenario is handled (for which the previous comment in 497-499 with changed sliding/regular ranges would fit best):

    regular: 444455
    sliding: 334444
    Then regular range 5 can be removed.

util-split.R
bockthom: Ok. Then let's ignore that _for now_.

util-split.R
MaLoefUDS: Lets just say I have discovered a funny series of bugs previously unnoticed when debugging on the issue data with the readded events previously deleted and uh yeah its weird

MaLoefUDS: Now im not sure anymore if we really should remove 5 in the case you mentioned in your most recent comment. Shouldn't regular ranges stay as they are?

bockthom: Wait, I mixed the to examples up. The 444455 example is the one which is handled in lines 504-507, right? Which is why wanted the comment to be moved down.

So the other case which I was looking for is this example:

    regular: 445555
    sliding: 444455
    Then sliding range 5 can be removed.

Maybe this is already cut correctly by default when calculating offsets etc.?

MaLoefUDS: The main problem is that I dont have an idea why the code even works with taking the second last element of the sliding windows, the subsequent if always is false therefore and debuggin it yields more bugs, oops.

MaLoefUDS: I think a simplification to decide which ranges to drop is to drop the last incomplete range. I did some sketches and think this should work out fine.

bockthom: I am not sure about whether removing the last incomplete range is always correct. If the last range is incomplete and contains data that is not already part of the previous regular or sliding range, then we want to keep the incomplete range. Only if everything is already contained in a previous range, we want to remove the incomplete range.

bockthom: Your explanations sound reasonable. However, the mandatory condition for this approach is: You only remove incomplete ranges after having reached x. Otherwise, this would not work, as the following two examples show:

    (I)
            __
    .....__|xx|........x
        |xx|xx|
        |xx|xx|
        |xx|__|
        |xx|  |
        |__|  |
        |RR|SS|

    (II)
            __
    .......|xx|........x
        |__|xx|
        |xx|xx|
        |xx|xx|
        |xx|__|
        |xx|  |
        |__|  |
        |RR|SS|

In these examples, the last range (SS) is incomplete but has to be kept as the last element (in I) or the last two elements (in II) are missing in RR. So, you cannot remove the last range. But if you use an additional constraint to check whether you have included the number of total activities, this might work. But I am not sure how complicated it is to determine x...

MaLoefUDS: * As I see it in (I) the last sliding range can be removed as x has been reached by the last regular range as well, or am I wrong?
* In (II) the last regular block which starts below x would be the highest / last and therefore removed leaving the last S range intact. S and R ranges alternate.
* The total number of elements x is already present as an integer inside the function ^^ (`items.unique.count`). I think this refactoring can greatly simplify the code (even if its not that easy to "understand" why its valid)

bockthom: > * As I see it in (I) the last sliding range can be removed as x has been reached by the last regular range as well, or am I wrong?

I was not sure about the concrete syntax, whether xx and ...x need to be in the same line to reach x or whether __ and ...x need to be in the same line to reach x. In my examples, ....|xx| ....x means that we reach the total number of x exactly in this line. What I might have misunderstood in your syntax is the meaning of "xx". So in my example "xx" means that a value is present and an empty lines means that it is not present in this range - which might be opposed to how you have used the syntax. Sorry for the confusion. So in (I), the last range of SS has three elements but is incomplete as it should have four. However, one of the three elements (namely the last one) is not included in RR, so the last range of SS cannot be removed although it is incomplete.

> * In (II) the last regular block which starts below x would be the highest / last and therefore removed leaving the last S range intact. S and R ranges alternate.

The last S range is later than the last R range. As in (I), also in (II) the last S range is incomplete but cannot be removed, as the last value of R is below x.

> * The total number of elements x is already present as an integer inside the function ^^ (`items.unique.count`). I think this refactoring can greatly simplify the code (even if its not that easy to "understand" why its valid)

Great. Then it might actually work to check for x and remove an incomplete range if x has already been reached by another range that has started earlier (interpreting earlier in terms of number of reached elements, such that it also works for activity-based splitting where multiples ranges can have the same time stamps).

MaLoefUDS: I've done a poc implementation and it seems to work however as we recently decided that the last regular range can be dropped as well if totally included in the last sliding window range, which was not clear before, I have to adapt some tests yet again 😂

util-split.R
```diff
@@ -99,26 +151,35 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
+            bins = get.date.from.string(bins)
+            bins = get.date.string(bins)
+            ## remove sliding windows
+            sliding.window = FALSE
+        } else {
+            ## sliding windows do not need to be removed here, as sliding windows and bins
+            ## are not contradicting in activity-based splitting
+            bins.vector = bins[["vector"]]
+            bins = bins[["bins"]]
+        }
```
bockthom, line 181: Please allow me to be forgetful. We already have discussed about these lines of code some weeks ago, but when I look at them again, I again don't understand what's going on here:

This "else" case handles the case that we want to split data activity-based by given bins, right?
How does this work? How can we split activity-based by bins? I could imagine that we can count the activities between two bins and use this activity amount for activity-based splitting. but what do we do when we have more than two bins given? This confuses me. And then the comment also states that we can have sliding windows in addition - how could this work if I cannot even imagine it without sliding windows. Maybe I am too tired now to be able to understand this little piece of code, but I am really confused now. 

Can you enlighten me in this matter @MaLoefUDS ?

util-split.R
```diff
@@ -480,12 +469,34 @@ split.data.activity.based = function(project.data, activity.type = c("commits",
...
+        ##
+        ## The conditional below inspects whether the very last item is in the first or the second
+        ## half of the last regular range. If it is in the first half, there will be a sliding
+        ## window which covers all items of the last regular range which makes the last regular
+        ## range obsolete.
+        ## Similarely if the last item is in the second half of the last regular range, there
+        ## will be a sliding range (which started at the half of the last regular range) which
+        ## contains only items also included in the last regular range, which makes the sliding
+        ## range obsolete.
+        if (((items.unique.count - 1) %% (activity.amount)) >= (offset.start)) {
```
bockthom, line 488: I am wondering whether this is off by one or not. Why do you subtract 1 form the unique item count? This might be correct in combination with `>=` (I am not sure). I am just wondering why you get the number of the second last time...

bockthom (changes requested):
Sorry for the delay. I still haven't managed to look at the tests. But I have had a look at the new changes of the main implementation. In general, the new changes look good. I only spotted a few comments that are still hard to understand and one implementation detail which I don't understand, making it hard for me to judge its validity. See my individual comments below.

 I will have a look at the tests in the next iteration.

util-split.R
bockthom: The longer I think about this, the more I think that it is correct. Your statement 
```R
if (((items.unique.count - 1) %% (activity.amount)) >= (offset.start)) {
```
can be transferred to the following equivalent one (which is more in line with your comments and looks more intuitive to me):
```R
if ((items.unique.count %% activity.amount) > offset.start) {
```
And if I understand the cropping at the beginning correctly, the offset.start-th element is still cropped, meaning that the offset.start+1st element is the first element of the sliding range. Please correct me if I am wrong.
If my argumentation is correct, then I would like you to replace your condition with the shorter condition that I proposed -- if you agree that it is easier to understand.

util-split.R
MaLoefUDS: Sure I think I can explain it.

Your first assumption is correct. When doing activity based splitting, we split all our data by a given `split.basis`, like `issue`, into sections with n unique entries. Now we expand the vectors to include all entries with duplicate ids as well. This means that the bins are now not all the same size. Those bins are then passed into the referenced `split.data.by.time.or.bins` method. This method then just splits the data by the given bins (not only the data of the given `split.basis` but *all* data. 

The role of `sliding.window` in this function when used for activity based splitting is a bit odd. The parameter does in this case **not** decide the splitting algorithm or similar, so the splitting stays the same whether this parameter is true or false however this method can be called with bins representing regular bins or representing sliding window bins and depending on this the splitting info in the end needs to be adjusted to fit. The comment actually refers to a change you requested. As in the then case of the if we set `sliding.window` to false because we cannot have given bins in time-based splitting and use sliding windows at the same time. In the case of activity based splitting these two do not contradict, in fact the parameter is not allowed to be overwritten there as we need it in the end for the splitting info. 

Hope that explains it :)

util-split.R
```diff
@@ -63,6 +64,57 @@ requireNamespace("lubridate") # for date conversion
...
+    return(split)
+}
+
+#' Split project data in time-based or activity-bin-based ranges as specified
+#'
+#' @param project.data the *Data object from which the data is retrieved
+#' @param splitting.length either \code{time.period} from \code{split.data.time.based}
+#'                         or \code{activity.amount} from\code{split.data.by.bins}
+#' @param bins either formatted as the \code{bins} parameter of \code{split.data.time.based}
+#'             or as the \code{bins} parameter of \code{split.data.by.bins}
```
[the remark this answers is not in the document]
MaLoefUDS: As I have just learned, the format differs. I copy pasted the description to both methods, I think it needs an update overall on the side of the activity bin based splitting 😅

util-split.R
MaLoefUDS: Your argumentation is solid, I also think it is easier to read so I will just exchange it ^^

util-split.R
bockthom: Ah, I get it. Thanks for the detailed explanations. Then everything is fine. I just did not question where the bins come from in this case. But if they are already given as function parameter correctly, then everything makes sense.
Can we maybe add an additional comment at the beginning of the else block? Something like
"the activity-based bins are already passed to this function in the correct format"
But as we do not make sure that the parameter is in the correct format, maybe "are" is wrong - this is just an assumption which may not hold when someone would use this function incorrectly. So maybe we should go with "should be" instead?

Should we check whether format is correct and print a warning if not? What is your opinion on that?

util-split.R
```diff
@@ -63,6 +64,57 @@ requireNamespace("lubridate") # for date conversion
...
+    return(split)
+}
+
+#' Split project data in activity-bin-based ranges as specified
+#'
+#' @param project.data the project data object from which the data is retrieved
+#' @param activity.amount the amount of data elements with unique ids to be considered in a bin, an integer.
+#' @param bins the date objects defining the start of ranges (the last date defines the end of the last range, in an
+#'             *exclusive* manner), including a vector which maps elements of the \code{split.basis} column of
+#'             \code{project.data} to bins, as produced by \code{split.get.bins.activity.based}.
```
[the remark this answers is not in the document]
MaLoefUDS:
```R
$vector
[1] 1 1 1 2 2 2 3 3

$bins
[1] "2016-07-12 15:58:59" "2016-07-12 16:06:10" "2016-07-12 16:06:32" "2016-07-12 16:06:33"
```

This is the format of the `bins` variable. Here is my suggestion of a descriptive comment:

```
@param bins the bins by which data should be split. Comprises of two components:
            \code{vector}: Assigns elements of the \code{split.basis} column of \code{project.data} to bins.
            \code{bins}: Dates defining the start of bins (the last date defines the end of the last bin, in an 
            *exclusive* manner).
            The expected format of \code{bins} is produced by  \code{split.get.bins.activity.based}.
            
...
@seealso split.get.bins.activity.based
```

What is your opinion on that?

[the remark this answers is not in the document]
bockthom: Sounds good! 
I've only spotted a doubled whitespace, everything else looks good to me.

util-split.R
MaLoefUDS: The method, which was the base of the new `split.data.by.time.or.bins` method, did no checks on the format of the bins either, even though the multiple notations of bins used here was not introduced by me. Other parameters are not checked consequently for their format either, and lastely, the method is only called by the two new wrapper methods, so if we want to introduce the format checks, I think we should do it there :).

bockthom: Ok, thanks for your reply. I am of two minds. Some time ago, we started to do input validation in some of the methods of the Conf and Data classes. I am not sure if we also should do it for other functions, such as the splitting functions, or if we should just let it fail if the format is wrong.

@hechtlC What is your opinion on that?

MaLoefUDS:
There is now only one point open (excluding the NEWS.md). @hechtlC What is your opinion on adding input validation for the `bins` variable in the splitting function (as mentioned in an earlier comment)?

util-split.R
```diff
@@ -94,11 +96,11 @@ split.data.by.bins = function(project.data, activity.amount, bins, split.basis =
 #'
 #' @param project.data the *Data object from which the data is retrieved
 #' @param splitting.length either \code{time.period} from \code{split.data.time.based}
-#'                         or \code{activity.amount} from\code{split.data.by.bins}
+#'                         or \code{activity.amount} from \code{split.data.by.bins}
 #' @param bins either formatted as the \code{bins} parameter of \code{split.data.time.based}
 #'             or as the \code{bins} parameter of \code{split.data.by.bins}
 #' @param split.by.time logical indicating whether splitting is done time-based or activity-bins-based
-#' @param number.windows see \code{number.windows} from \code{split.data.time.by.bins.vector}
+#' @param number.windows see \code{number.windows} from \code{split.data.time.based}
```
bockthom, line 103: Really? Shouldn't this be just `split.data.by.bins`? Please check again.

util-split.R
```diff
@@ -485,7 +487,7 @@ split.data.activity.based = function(project.data, activity.type = c("commits",
         ## will be a sliding range (which started at the half of the last regular range) which
         ## contains only items also included in the last regular range, which makes the sliding
         ## range obsolete.
-        if (((items.unique.count - 1) %% (activity.amount)) >= (offset.start)) {
+        if ((items.unique.count %% activity.amount) > (offset.start)) {
```
bockthom, line 490: Could you please remove the unnecessary brackets around `offset.start`?

util-split.R
MaLoefUDS: No, this parameter comes from the `split.data.time.based` function

bockthom:
> There is now only one point open (excluding the NEWS.md). @hechtlC What is your opinion on adding input validation for the `bins` variable in the splitting function (as mentioned in an earlier comment)?

@MaLoefUDS There is still one other point open: 

"When we take the directed option from the network configuration, we also should have a test for the directed issue artifact network." 

As far as I can see, the tests in [tests/test-networks-artifact.R](https://github.com/se-sic/coronet/pull/244/files#diff-9fa163c01b5f6ca153b2b0b043c5b694deeabfcaee1acc56dbaf304752ef9c12) only test undirected networks.
Maybe it is possible to use the existing tests and just make parameterized tests out of it, using `packrat` - then you just need to pass directed as a parameter and use it everywhere you need it in the test

util-split.R
```diff
@@ -87,226 +97,29 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
-        cf.data = split.data.time.based.by.ranges(project.data, ranges)
-
-        ## update project configuration
-        project.conf.new$set.revisions(bins, bins.date, sliding.window = TRUE)
-        for (cf in cf.data) {
-            ## re-set project configuration due to object duplication
-            cf.conf = cf$set.project.conf(project.conf.new)
-        }
+    # validate type of the 'vector' component of the 'bins' parameter
+    if (class(bins[["vector"]]) != "numeric") {
```
bockthom, line 115: Hm, I am not sure if this strict check is necessary or even correct. In R, every variable can have multiple classes.
Maybe you can use `!inherits(bins[["vector"]], "numeric")` here?

util-split.R
```diff
@@ -87,226 +97,29 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
-        revisions.dates = bins.date
-    )
-
-    ## set bin attribute
-    attr(cf.data, "bins") = bins.date
-
-    ## return list of RangeData objects
-    return(cf.data)
+    split = split.data.by.time.or.bins(project.data, activity.amount, bins, split.by.time = FALSE,
+                                        sliding.window = sliding.window, split.basis = split.basis)
```
bockthom, line 121: The indentation of the second line is wrong here (needs to be shifted by one to the left).

bockthom (changes requested):
Almost done with the implementation ;) I only spotted a few potential issues with the input validation. I am not sure whether all the spotted issues are actual issues or not, please check.

util-split.R
MaLoefUDS: I can change it if you want, it does not break our tests to have it like this, but I agree your version is probably more correct (does not break any tests either)

util-split.R
```diff
@@ -64,6 +64,16 @@ requireNamespace("lubridate") # for date conversion
 split.data.time.based = function(project.data, time.period = "3 months", bins = NULL,
                                  number.windows = NULL, split.basis = c("commits", "mails", "issues"),
                                  sliding.window = FALSE, project.conf.new = NULL) {
+
+    # validate type of the 'bins' parameter
+    if (!is.null(bins) && !lubridate::is.POSIXct(bins)) {
+        dates = parallel::mclapply(bins, function(bin) lubridate::ymd_hms(bin, truncated = 3))
+        if (any(is.na(dates))) {
+            logging::logerror("The bins parameter, if present, needs to be a character representing a date")
```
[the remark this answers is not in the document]
MaLoefUDS: The `dates` variable is defined above as `dates = parallel::mclapply(bins[["bins"]], function(bin) lubridate::ymd_hms(bin, truncated = 3))`. lubridate::ymd_hms returns NA ist the string given as input does not match the time format.

So in the case where one entry of `dates` is NA, this means that we don't have a valid time stamp here.

util-split.R
```diff
@@ -64,6 +64,16 @@ requireNamespace("lubridate") # for date conversion
 split.data.time.based = function(project.data, time.period = "3 months", bins = NULL,
                                  number.windows = NULL, split.basis = c("commits", "mails", "issues"),
                                  sliding.window = FALSE, project.conf.new = NULL) {
+
+    # validate type of the 'bins' parameter
+    if (!is.null(bins) && !lubridate::is.POSIXct(bins)) {
+        dates = parallel::mclapply(bins, function(bin) lubridate::ymd_hms(bin, truncated = 3))
```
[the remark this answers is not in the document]
MaLoefUDS: If we are *only* talking about the line below the conditional, I can change it to `dates = parallel::mclapply(bins, function(bin) get.date.from.string(bin))` without a problem :)

util-split.R
bockthom: Yeah, please change it. I'd feel better if we are less restrictive, maybe there are some cases in which multiple classes occur. In addition, a less restrictive check should also be more robust to future changes in R and (hopefully) reduces the change effort for us then.

util-split.R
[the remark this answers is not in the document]
bockthom: That's clear. The question I had in mind is the following, though (sorry for being unprecise in my first comment): What happens in all the splitting functions etc. when one of the elements in `NA`. Would be interesting to know if our splitting methods just ignore the `NA` or whether this leads to errors or unwanted behavior (ranges that start and end differently than expected)

util-split.R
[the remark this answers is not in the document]
bockthom: This should be enough, I guess. If I am not mistaken, we use the `isPOSIXct` check at different places in coronet (we could also think about defining our own wrapper for it, but I guess it's not worth yet). 
But the actual date conversion should only be done at one single place - and that's what's `get.date.from.string` is doing.

util-split.R
```diff
@@ -87,226 +97,29 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
     }
 
-    if (is.null(project.conf.new)) {
-        ## Clone the project configuration, so that splitting repeatedly does not interfere
-        ## with the same configuration.
-        project.conf.new = project.data$get.project.conf()$clone()
+    # validate type of the 'bins' component of the 'bins' parameter
+    dates = parallel::mclapply(bins[["bins"]], function(bin) lubridate::ymd_hms(bin, truncated = 3))
+    if (any(is.na(dates))) {
+        logging::logerror("The 'bins' component of the bins parameter needs to be a character representing a date")
```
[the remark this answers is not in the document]
bockthom: Wait. What happens if there is no "bins" element or no "vector" element (below). In such cases, the "mclapply" could throw errors. Maybe adjust the check above to check for the existence of "bins" and "vector" first...

util-split.R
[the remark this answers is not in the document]
MaLoefUDS: 1) I do not even validate that the bins do not contain NAs (maybe I should, because...)
2) Having an NA inside as a bin entry does break the code when calling `findInterval` in `split.data.by.timeor.bins`(`stop("'vec' must be sorted non-decreasingly and not contain NAs")`)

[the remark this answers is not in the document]
bockthom: Thanks for checking what happens in this case. Then we should definitively check for `NA`s here :+1:

util-split.R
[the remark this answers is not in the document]
MaLoefUDS: In this case the dates object will be an empty `list()`. As this will not trigger the conditional `any(is.na(dates))` I will add a check for it. As for the case where `vector` is missing, this will mean that `bins[["vector"]]` returns `NULL` which triggers the conditional `!inherits(bins[["vector"]], "numeric")`. There is no check needed in this case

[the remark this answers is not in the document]
bockthom: Agreed!

util-split.R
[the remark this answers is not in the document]
MaLoefUDS: Nevermind, in the case where a bin is `NA` the `dates` object will contain an `NA` which will trigger a `stop`, no need for a check

util-split.R
```diff
@@ -65,9 +65,9 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
                                  number.windows = NULL, split.basis = c("commits", "mails", "issues"),
                                  sliding.window = FALSE, project.conf.new = NULL) {
 
-    # validate type of the 'bins' parameter
+    # validate existence and type of the 'bins' parameter
     if (!is.null(bins) && !lubridate::is.POSIXct(bins)) {
-        dates = parallel::mclapply(bins, function(bin) lubridate::ymd_hms(bin, truncated = 3))
+        dates = parallel::mclapply(bins, function(bin) get.date.from.string(bin))
```
bockthom, line 70: Could be simplified to: `dates = parallel::mclapply(bins, get.date.from.string)` (also below in the second occurrence of this line)

util-split.R
```diff
@@ -104,21 +104,26 @@ split.data.by.bins = function(project.data, activity.amount, bins, split.basis =
         stop("Stopped due to incorrect parameter types")
     }
 
-    # validate type of the 'bins' component of the 'bins' parameter
-    dates = parallel::mclapply(bins[["bins"]], function(bin) lubridate::ymd_hms(bin, truncated = 3))
+    # validate existence and type of the 'bins' component of the 'bins' parameter
+    if (!"bins" %in% names(bins)) {
```
bockthom, line 108: Please add parentheses for better readability: `(!("bins" %in% names(bins)))`

bockthom (commented):
The current implementation looks good to me. 
I have two suggestions for better code readability.
This is my last comment on the implementation in this PR, looks really good to me now.

Let's see when I find some time to have a look at some of the tests...

util-split.R
```diff
@@ -105,12 +105,12 @@ split.data.by.bins = function(project.data, activity.amount, bins, split.basis =
...
 
     # validate existence and type of the 'bins' component of the 'bins' parameter
-    if (!"bins" %in% names(bins)) {
+    if (!("bins" %in% names(bins))) {
         logging::logerror("The 'bins' parameter needs to include a component 'bins'")
         stop("Stopped due to incorrect parameter types")
     }
 
-    dates = parallel::mclapply(bins[["bins"]], function(bin) get.date.from.string(bin))
+    dates = parallel::mclapply(bins, get.date.from.string)
```
bockthom, line 113: Shouldn't that be `bins[["bins"]]` here instead of just `bins`, as it was before?

Interestingly, this did apparently not lead to a test failure in our CI pipeline, and I am wondering why. If I try to call  `as.POSIXct(list(vector=c(1,2,3), bins=list("2020-02-02","2020-02-03", "2020-02-04"))`, which should be the format of how `bins` looks like, I get an error. So, I would expect a similar error in our tests, as this function call should always produce an error if `bins` consists of elements `bins` and `vector` as expected... 

So, now I am confused. Can you please find out why this did not lead to an error in our tests?

MaLoefUDS: Oh this was a copy paste error 💀 but still interesting outcome, I will investigate until tomorrow ^^

MaLoefUDS: Okay I got it done quicker than I thought:

1) What it does is, it sets all values in the `bins[["vector"]]` to `NA` and all the values in the``bins[["bins"]]`` to the dates they were before (as intended). Then it goes over all elements in bins (which are `vector` and `bins`) and says, none of those are `NA`, this is why it does not go into the error case.

2) It does also not do false positives because none of our tests test against this input validation yet (so no tests have `NA` in the bins parameter) and therefore the error case should not be triggered by any test and is factually not being triggered because of 1).

bockthom: Nevermind, I forgot to add the `lapply` in an exemplary call. If I add the `lapply`, I don't get any errors:

`lapply(list(x=c(1,2,3), y=c("2020-02-02","2020-02-03", "2020-02-04")), as.POSIXct)`

Thus, there also will be no error when we execute our tests. So, it should be fine to just fix the `bins[["bins"]]` part here.

bockthom: >  none of our tests test against this input validation yet

That's a valid argument and actually a good idea to add some negative tests.
Could you please add some negative tests? What I have in mind is something like this:

Add one test case that tests invalid inputs for the two functions for which we have added input validation and call either of them with the format that would be expected by the respective other function. And we can also try to pass totally invalid `bins` that are not integers to them. This should be enough negative testing.

I am sorry to ask for something new again :see_no_evil:

MaLoefUDS: When testing whether the more complex format of `bins` from `split.data.by.bins` produce issues in the input validation of `split.data.time.based` I encountered this issue again from https://github.com/se-sic/coronet/pull/244#discussion_r1400689507: 

(For context, the comment is talking about the conversion by `get.date.from.string` of the more complex type of `bins`, which is applied in the input validation of `split.data.time.based`)

> 1. What it does is, it sets all values in the `bins[["vector"]]` to `NA` and all the values in the`bins[["bins"]]` to the dates they were before (as intended). Then it goes over all elements in bins (which are `vector` and `bins`) and says, none of those are `NA`, this is why it does not go into the error case.

In this case this will still produce an error in the execution of the splitting logic later on, but I think it would be better to check for this case in the input validation and early exit. 

Is it fine if I amend it to [Improve input validation](https://github.com/se-sic/coronet/pull/244/commits/bb5aeb242b69c0e5b242b3a347e7f3b7462adb43) or does it need a separate commit?

bockthom: First of all, it is good that we already get an error in this case - better than nothing :smile: 
But I agree with you: If we can check it earlier and, thus, we receive a more appropriate error message, it would be better.

And yes, of course, you can amend the last commit. As long as this PR is not merged yet and as long as it does make sense to amend the last commit because the changes are somehow related, I am fine with amending. But keep in mind that you can also adjust the commit message while amending...

bockthom: I am a little bit confused by your commit message.
"Disallowing non-unlisted named lists" means what? :confused: 

Why not saying something like "disallow lists that contain elements that are not representing a date, as they do not comply with the expected format of bins for 'split.data.by.time.or.bins'."?


---


Btw, I just noticed that the error message is a little bit imprecise: 

```R
dates = parallel::mclapply(unlist(bins), get.date.from.string)
if (any(is.na(dates))) {
    logging::logerror("The bins parameter, if present, needs to be a character representing a date")
```
The bins parameter usually is not a character representing a date, but usally a _character vector_ representing _multiple dates_, right?

tests/test-split-data-activity-based.R
```diff
@@ -1794,3 +1794,46 @@ patrick::with_parameters_test_that("Split a data object activity-based (number.w
...
 ))
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Input validation for splitting data by activity bins --------------------
+
+##
+## Test that split.data.by.bins does not accept an invalid bins parameter
+##
+
+test_that("Split a data object time-based with invalid bins parameter.", {
```
bockthom, line 1805: Shouldn't the test have a different name than the other test? I miss a "by bins" after time-based here...

MaLoefUDS: I can add that, surely

util-split.R
MaLoefUDS: The problem I encountered was that there are named lists and named lists. One of the format
```R
$a
[1] 1 2 3

$b
[1] 2 3 4
```
and one of the format
```R
a1 a2 a3 b1 b2 b3 
 1  2  3  2  3  4 
```

Both are named lists, however `list1[1]` yields
```R
$a
[1] 1 2 3
```
(which is an issue),
and `list2[1]` yields
```R
a1 
 1
```
(which is not an issue).

Sorry for my confusing comment, I will change it 👍

MaLoefUDS: Yes, about the error message im with you!

MaLoefUDS:
(I just found out that I used paste0 is not what I want here so I needed to amend once more 💀)

tests/test-split-data-time-based.R
```diff
@@ -961,6 +961,43 @@ patrick::with_parameters_test_that("Split a data object time-based (bins = ... ,
     "pasta, synchronicity: TRUE" = list(test.pasta = TRUE, test.synchronicity = TRUE)
 ))
 
+##
+## Test that split.data.time.based does not accept an invalid bins parameter
+##
+
+test_that("Split a data object time-based by bins with invalid bins parameter.", {
```
bockthom, line 968: Please correct me if I am wrong, but did you add "by bins" to wrong test?

tests/test-split-data-activity-based.R
bockthom, line 1805: I am not sure if this is the right position for the test here. As the test handles bins, it belongs rather to time based splitting than to activity-based splitting. So, I would suggest to move this test to test-split-data-time-based.R.

util-split.R
```diff
@@ -104,21 +105,27 @@ split.data.by.bins = function(project.data, activity.amount, bins, split.basis =
...
+    if (!("bins" %in% names(bins))) {
+        logging::logerror("The 'bins' parameter needs to include a component 'bins'")
+        stop("Stopped due to incorrect parameter types")
+    }
+
+    dates = parallel::mclapply(bins[["bins"]], get.date.from.string)
     if (any(is.na(dates))) {
-        logging::logerror("The 'bins' component of the bins parameter needs to be a character representing a date")
+        logging::logerror(paste("The 'bins' component of the 'bins' parameter, needs to be vector",
+                                "whose elements represent dates"))
```
bockthom, line 117: needs to be **a** vector

util-split.R
```diff
@@ -65,11 +65,12 @@ split.data.time.based = function(project.data, time.period = "3 months", bins =
...
                                  sliding.window = FALSE, project.conf.new = NULL) {
 
-    # validate type of the 'bins' parameter
+    # validate existence and type of the 'bins' parameter
     if (!is.null(bins) && !lubridate::is.POSIXct(bins)) {
-        dates = parallel::mclapply(bins, function(bin) lubridate::ymd_hms(bin, truncated = 3))
+        dates = parallel::mclapply(unlist(bins), get.date.from.string)
         if (any(is.na(dates))) {
-            logging::logerror("The bins parameter, if present, needs to be a character representing a date")
+            logging::logerror(paste("The bins parameter, if present, needs to be vector",
```
bockthom, line 72: needs to be **a** vector

bockthom (changes requested):

tests/test-split-data-time-based.R
MaLoefUDS: Hmm to me it looks like it’s the right test, I only added two, one for time based and one for activity based splitting, and this is the time based one

tests/test-split-data-activity-based.R
MaLoefUDS: Yeah I was not sure about the position either 😅 should I just move it to the other test I added?

tests/test-split-data-time-based.R
bockthom: I am confused now. What I meant was the following:
The test that starts in line 1805 of test-split-data-activity-based.R tests `split.data.by.bins`, but the test name says `Split a data object time-based`. So, there, in line 1805, it should be `Split a data object by bins`, if I am not mistaken.

But here, starting in line 968 in test-split-data-time-based.R, `split.data.time.based` is tested (**not** `split.data.by.bins`). So, I think that the "by bins" that you have add here in line 968 is wrong, as it should be added to line 1805 in the other test file instead.

tests/test-split-data-activity-based.R
bockthom: Yeah, I also have talked about the test position together with @hechtlC and he also voted for moving it to test-split-data-time-based.R.

So I would suggest to move it directly below the other test that you already have added to test-split-data-time-based.R.

tests/test-split-data-time-based.R
MaLoefUDS: Ohhh yeah you're right, I got the wrong test.

Actually the other test does not perform splitting by time but just splitting by given bins (which come from activity based splitting [which is why I moved it into the activity test file]) so the test caption is wrong altogether and I will update both when I get to it :)

bockthom: [even if splitting by bins is called by activity-based splitting, the splitting by bins per se is time based as the bins denote time stamps]

MaLoefUDS: Actually not. The relevant information after which is being split in this case in the bins vector (which is activity based). The dates in the bins are used as labels and attributes and whatnot but not for splitting itself.

hechtlC: I would prefer to have the test in `test-split-misc.R`.

tests/test-split-misc.R
```diff
@@ -124,6 +124,50 @@ test_that("Split network and data on low level (split.dataframe.by.bins, split.n
 
 })
 
+##
```
bockthom, line 127: @MaLoefUDS  Could you please add your copyright header at the beginning of this file `test-split-misc.R` ?

bockthom (changes requested):

bockthom:
Thanks @MaLoefUDS for addressing my last comment. Now, everything I've spotted seems to be fixed now.
I will do my best to have a look at the changed tests until our meeting on Wednesday, because I would like to merge this PR before Christmas. 

So, given that I don't spot anything problematic in the changed tests (but we'll see), the only thing missing is the update of the `NEWS.md`. Could you please already collect the different changes from this PR and sort them into the different categories (fixed, added, changed/improved), so that we can discuss them on Wednesday? (Not every single change needs to be listed separately, many of them can be combined into one single item, I guess.) Also don't forget about this one here: https://github.com/se-sic/coronet/pull/244#discussion_r1331663005

tests/test-networks-artifact.R
```diff
@@ -101,3 +101,97 @@ test_that("Network construction of the undirected artifact-cochange network", {
...
+    network.built = network.builder$get.artifact.network()
+
+    ## 1) vertices
+    vertices = data.frame(name = c("<issue-jira-ZEPPELIN-328>",
+                                   "<issue-github-2>",
+                                   "<issue-github-1>",
+                                   "<issue-github-3>",
+                                   "<issue-github-4>",
+                                   "<issue-jira-ZEPPELIN-332>",
+                                   "<issue-github-6>"),
```
bockthom, line 179: The order looks a little bit weird. As in the test directly above, I would start with 328, followed by 332...

bockthom (changes requested):
I found some time to have a look at the tests on a sample basis. 
In general, the tests (at which I had a look at) look good. I spotted a few minor inconsistencies and a few changes that we should double-check and document.

In addition, I spotted one inconsistency in the test data :see_no_evil: 

Moreover, after everything is fixed in the test data, we need to copy 
`test_feature/feature/issues-github.list` to `test_proximity/proximity/issues-github.list`, and the same also for `issues-jira.list`, as our basic assumption is that the issue data for feature and proximity are the same.

tests/test-split-data-activity-based.R
```diff
@@ -761,7 +770,7 @@ patrick::with_parameters_test_that("Split a data object activity-based (activity
         split.type = "activity-based",
         split.length = 18,
         split.basis = "commits",
-        split.sliding.window = TRUE,
+        split.sliding.window = FALSE,
```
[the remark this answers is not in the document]
MaLoefUDS: To be honest, I cannot find the reason why it has not been like this before and it had to be changed. But I remember, that I fixed something and this was also a consequence from it (from the timestamps, it was most likely something about removing the last range debate we had). 

The code which sets the `sliding.window` to false can be found in util-split line 980 and was not introduced by me:
```R
if ((length(bins.ranges) <= 1) && sliding.window) {
    logging::logwarn("Sliding-window approach does not apply for one range or less.")
    sliding.window = FALSE
}
```

This test here checks the splitting info set for each individual split returning from `split.data.activity.based` (compared to the splitting info set for the "global config") and therefore I think its fair to say that `sliding.window` is false. But I would need to dive in deeper to find the reason for the change.

tests/codeface-data/results/testing/test_feature/feature/issues-jira.list
```diff
@@ -11,6 +11,8 @@
...
 "ZEPPELIN-328";"[ZEPPELIN-328] Interpreter page should clarify the % magic syntax for interpreter group.name";"[""issue"", ""bug""]";"closed";"[""fixed""]";"2013-04-21 23:52:09";"2013-05-25 20:02:08";"[""GUI"", ""Interpreters""]";"resolution_updated";"Björn";"bjoern@example.org";"2013-06-01 06:53:06";"fixed";"""unresolved"""
+"ZEPPELIN-328";"[ZEPPELIN-328] Interpreter page should clarify the % magic syntax for interpreter group.name";"[""issue"", ""bug""]";"closed";"[""fixed""]";"2013-04-21 23:52:09";"2013-05-25 20:02:08";"[""GUI"", ""Interpreters""]";"add_link";"Thomas";"thomas@example.org";"2013-06-01 06:54:00";"ZEPPELIN-332";"""issue"""
+"ZEPPELIN-332";"[ZEPPELIN-332] CNFE when running SQL query against Cassandra temp table";"[""issue"", ""bug""]";"open";"[""unresolved""]";"2016-07-12 16:01:30";"";"[""Interpreters""]";"referenced_by";"Thomas";"thomas@example.org";"2013-06-01 06:54:00";"ZEPPELIN-328";"""issue"""
```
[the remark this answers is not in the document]
MaLoefUDS: Oh god, this sounds like fun 😂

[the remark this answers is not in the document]
bockthom: I know. And I am sorry for that. But the change I suggested should be the least expensive one - just flipping "referenced_by" and "add_link" and changing two dates. I know, changing the dates will affect many of splitting test, but not all.

tests/test-split-data-activity-based.R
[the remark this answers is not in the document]
bockthom: Ok, but then I would like to add the information from the warning message also as a comment in the tests.
But still it is important to check that this behavior is idententical for the different splitting methods. So, activity-based, time-based, what ever splitting all need to set the splitting information to false in such a case - otherwise it would be inconsistent. Could you please check that?

tests/test-networks-artifact.R
MaLoefUDS: Unfortunately the order cannot be freely chosen, since both `igraph::identical_graphs` as well as `compare.networks` also test for the order of elements in the vertices column 🫠.

bockthom: Ok, but this is weird. Why would one order the vertices in this crazy order? It is neither ordered lexicographically, nor by occurrence in the data.
But if igraph wants to use exactly this order, then we don't have another choice, I guess...

tests/test-data.R
```diff
@@ -402,8 +403,8 @@ test_that("Filter bots from issue data", {
     filtered.issues = proj.data$get.issues()
 
     expect_true(all(filtered.issues[["author.name"]] != "Thomas"))
-    ## there are now 41 issue events remaining, since 6 issue events have been removed during filtering
-    expect_equal(nrow(filtered.issues), 41)
+    ## there are now 42 issue events remaining, since 11 issue events have been removed during filtering
```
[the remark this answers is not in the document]
bockthom: @MaLoefUDS Did you forget about this one? I could not spot this change in your today's commits...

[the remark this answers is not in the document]
MaLoefUDS: No I just wanted to commit the other changes before leaving work for the day, and for this change I wanted to wait for your answer on the comment with the weird vertices order. I will commit this fix soon

NEWS.md
```diff
@@ -9,12 +9,17 @@
...
 - Add CI support for GitHub Actions (PR #234, fa1fc4af65751402ae6b23298dd4ed821930c6d2)
+- Add issue-based artifact-networks (PR #244, 98a93ee721a293410623aafe46890cfba9d81e72)
+- (Re)add 'split.data.by.bins' function which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698)
 
 ### Changed/Improved
 
 - Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f)
 - Include removal of empty and deleted users in the setters of mails, commits, issues, and authors. For commits, also the `committer.name` column is now checked for deleted or empty users (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
 - Check for empty values (i.e., values of length < 1) when updating configuration attributes and throw an error if a value is empty (9f36c544637ab4f4173408152d223b9b5098ce5a)
+- Enhance testing data by adding 'add_link' and 'referenced_by' issue events which connect issues to form edges in issue-based artifact-networks (PR #244, 9f840c040d552e8639aa82c3dd537c189679b348)
```
bockthom, line 20: Please add another commit hash here in addition: ea4fe8d3c84f948af6147cf0137e80181ebb7a1e

NEWS.md
```diff
@@ -9,12 +9,17 @@
 - Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #231, d1d9a039f50480ec5b442dc7e8b518648d1f9d9d)
 - Add helper function `is.single.na` to check whether an element is of length 1 and is `NA` (ddff2b8bbca6405f5c7c1cf4e7e97374fb1426ca, ccfc2d12a68dfa412f05159e8e3b03118694e748)
 - Add CI support for GitHub Actions (PR #234, fa1fc4af65751402ae6b23298dd4ed821930c6d2)
+- Add issue-based artifact-networks (PR #244, 98a93ee721a293410623aafe46890cfba9d81e72)
```
bockthom, line 12: Please add this commit hash in addition: 771bcc8d961d419b53a1e891e9dc536371f1143b

NEWS.md
```diff
@@ -9,12 +9,17 @@
...
+- (Re)add 'split.data.by.bins' function which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698)
 
 ### Changed/Improved
 
 - Include structural verification to almost all functions that read dataframes from files or set a dataframe (setter-functions) (PR #231, b7a95881da72ccaa548c6cd5d94bd558a25caa6f)
 - Include removal of empty and deleted users in the setters of mails, commits, issues, and authors. For commits, also the `committer.name` column is now checked for deleted or empty users (PR #235, 08fbd3e11e33d060f42cbc6f729eaf60b48a6de7)
 - Check for empty values (i.e., values of length < 1) when updating configuration attributes and throw an error if a value is empty (9f36c544637ab4f4173408152d223b9b5098ce5a)
+- Enhance testing data by adding 'add_link' and 'referenced_by' issue events which connect issues to form edges in issue-based artifact-networks (PR #244, 9f840c040d552e8639aa82c3dd537c189679b348)
+- Add input validation for the 'bins' parameter in time-, and bin-based splitting (PR #244, ed0a5302ea8c8934d7200b95be7ac1446305af07)
```
bockthom, line 21: Please name all the affected functions instead of just saying "time- and bin-based splitting". This makes it easier to grep the NEWS.md for specific function names to find out if there has changed something :)

And also add this commit hash here in addition: 5e5ecbac44d07927b953ae9d4330a616f8224ba7

NEWS.md
```diff
@@ -9,12 +9,17 @@
 - Add function `verify.data.frame.columns` to check that a dataframe includes all required columns, optionally with a specified datatype (PR #231, d1d9a039f50480ec5b442dc7e8b518648d1f9d9d)
 - Add helper function `is.single.na` to check whether an element is of length 1 and is `NA` (ddff2b8bbca6405f5c7c1cf4e7e97374fb1426ca, ccfc2d12a68dfa412f05159e8e3b03118694e748)
 - Add CI support for GitHub Actions (PR #234, fa1fc4af65751402ae6b23298dd4ed821930c6d2)
+- Add issue-based artifact-networks (PR #244, 98a93ee721a293410623aafe46890cfba9d81e72)
+- (Re)add 'split.data.by.bins' function which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698)
```
bockthom, line 13: Instead of "(Re)add" I would prefer to write "Add a new ... function (not to be confused with a previously existing function that had the same name and was renamed in this context)"

As this function was renamed after the referenced commits, we should also add the commit hash of the commit that does the renaming: ed5feb214a123b605c9513262f187cfd72b9e1f4

NEWS.md
```diff
@@ -28,6 +33,10 @@
 - Prevent R warnings `In xtfrm.data.frame(x) : cannot xtfrm data frames` (PR #237, c24aee7d8f0b6ff4b641c8922e6ee1dce6f5999c)
 - Fix wrong bracket in pasted logging message (PR #241, 50c68cb60114b49c32dc5be15014745cb8d42ded)
 - Replace deprecated R function calls (PR #237, ed433821c04711a96501887b315d1b0ea8681f5a)
+- Reformat 'event.info.1' column of issue data according to the <issue-%source-%id> format, if the content of the 'event.info.1' field references another issue (PR #244, 62ff9d0f31adbefb3381936237dc4ab984e33acb)
+- Replace time-based splitting as a helper inside activity-based splitting, because of its imprecision, by pure bin-based splitting (see 'split.data.by.bins'). Previously activity-based splitting partitioned data into bins, obtained the timestamps of the elements at the bin borders and did time based splitting. However, elements crossing bin borders can have equal timestamps which causes issues (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698)
```
bockthom, line 37: That sounds very complex. Can we put the spotlight on the actual fix instead of discussing implementation details?
That is, it does not matter whether time-based splitting was used as a helper inside activity-based splitting. Instead, we should start with the problem that has been fixed here. Speaking of time-based splitting sounds confusing, as this change should not affect time-based splitting at all. So, for me, it is more important to state what has been fixed and why than explaining how this has been fixed, if it is just an implementation detail that is not relevant to the user. So, the "how" part could be placed at the end of the item, if it makes sense, but I would not put it at the beginning...

NEWS.md
```diff
@@ -28,6 +33,10 @@
...
+- Remove last range when using a sliding-window approach and the last ranges elements are fully contained in the second last range (PR #244, 48ef4fa685adf6e5d85281e5b90a8ed8f6aeb197)
```
bockthom, line 38: the last ranges elements :arrow_right: the last range's elements

NEWS.md
```diff
@@ -28,6 +33,10 @@
...
+- Rename vertex attribute from 'IssueEvent' to 'Issue' in multi-networks to be consistent with bipartite-networks (PR #244, 26d7b7e9fd6d33d1c0a8a08f19c5c2e30346a3d9)
```
bockthom, line 39: Not sure about the wording, what about "Rename vertex attribute 'IssueEvent' to 'Issue' in multi-networks, to be..."

bockthom (changes requested):
Thanks for updating the NEWS.md. There is one major issue: You have added the changes to the v4.3 version, which has already been released. This is not a good idea, we will not change a past release. So, please move all your changes to a new section and just call it `## unversioned`. We will add a version number when the next release will be scheduled. Also make sure that we have two empty lines when the version section ends, to be consistent with the previous versions.

I agree with the commits you have referenced, but sometimes I would like to add another commit in addition, if a change is not fully implemented in one commit (and here I only speak of real implementation commits that affect the behavior, not only tiny changes). I added suggestions for these commits below.

bockthom (commented):
And one more comment: Please use backticks  around function names and strings such as `referenced_by` or `IssueEvent`  instead of ', such that it is properly highlighted by markdown.

hechtlC (approved):

bockthom (approved):
Finally, everything seems to be fine with this PR.
Many thanks @MaLoefUDS for your commitment and endurance in this PR. This was a really tough one, also due to plenty of necessary changes to the test data and splitting tests. But we also spotted bugs along this way and improved the splitting functionality  substantially, which is a good result for a long PR that was supposed to "just" introduce issue-based artifact networks :smile:

PR #250, opened 2024-02-08
Enable simplification of multi-relation edges & make multi-network construction work with igraph version 2.0.1.1

### LINKED ISSUE #246: Error message logging in `metrics.smallwordness` is broken
### Description

Logging the error message in `metrics.smallworldness` does not work, as `logging::error` does not exist. Instead, it should be `logging::logerror`.

https://github.com/se-sic/coronet/blob/a0c3a5262a8837679e2c04343e10b2e41dcab4df/util-networks-metrics.R#L170

However, this does not really matter, as the execution would be stopped in the next line via `stop` anyway. Thus, this bug only prevents printing the error message.

### Versions

[v4.3](https://github.com/se-sic/coronet/releases/tag/v4.3) [v4.2](https://github.com/se-sic/coronet/releases/tag/v4.2) [v4.1](https://github.com/se-sic/coronet/releases/tag/v4.1) [v4.0](https://github.com/se-sic/coronet/releases/tag/v4.0)

bockthom:
### Prerequisites

### Description
This PR addresses multiple issues:

- It fixes #246 (i.e., fixes a broken logerror statement)
- If adjusts multi-network construction such that it works with the recently released igraph version 2.0.1.1, in which it is not possible any more to add an empty list of vertices. A simple check for an empty list fixes this problem.
- It updates the README regarding the effectiveness of network conf parameter "artifact.directed", which now is effective as of PR #244.
- It enables multi-relation edges to be simplified. That is, the simplify methods get an additional parameter and also the network configuration gets an additional parameter to decided whether edges between a pair of vertices should be contracted to a single edge when the edges are of different relations. The default behavior stays as before, that is, edges of different relations are not contracted during simplification. However, notice that the implementation of this feature is not complete, as also the edge attribute handling needs to be adjusted accordingly, and also proper tests are missing. I leave this open to a future PR -  I have summarized the steps that are still missing in issue #251.

#### Added
- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa)

#### Changed/Improved
- Explicitly add R version 4.3 to the CI test pipeline (9f346d5bc3cfc553f01e5e80f0bbe51e1dc2b53e)

#### Fixed
- Fix broken error logging in `metrics.smallworldness` (03e06881f06abf30d44b69d7988873f20b95232d)
- Fix multi-network construction to work with igraph version 2.0.1.1, which does not allow to add an empty list of vertices (PR #250, 5547896faa279f6adaae4b2b77c7ab9623ddf256)

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -12,6 +12,7 @@
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
 ## Copyright 2023 by Maximilian Löffler <s8maloef@cs.uni-saarland.de>
+## Copyright 2024 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## All Rights Reserved.
 
 name: Build Status
@@ -29,17 +30,17 @@ permissions:
 jobs:
   build:
     name: Build
-    
+
     # change to 'runs-on: self-hosted' to run on self-hosted runners (https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job)
     runs-on: ubuntu-latest
-    
+
     strategy:
       fail-fast: false
       matrix:
-        r-version: ['3.6', '4.0', '4.1', '4.2', 'latest']
+        r-version: ['3.6', '4.0', '4.1', '4.2', '4.3', 'latest']
 
     steps:
-      - name: Checkout Repo 
+      - name: Checkout Repo
         uses: actions/checkout@v3
 
       - name: Update system
@@ -56,7 +57,7 @@ jobs:
         uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.r-version }}
-      
+
       - name: Install dependencies
         run: Rscript install.R
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,10 +6,11 @@
 
 ### Added
 
-- Add issue-based artifact-networks (PR #244, 98a93ee721a293410623aafe46890cfba9d81e72, 771bcc8d961d419b53a1e891e9dc536371f1143b)
+- Add issue-based artifact-networks (PR #244, 98a93ee721a293410623aafe46890cfba9d81e72, 771bcc8d961d419b53a1e891e9dc536371f1143b, 368e79264adf5a5358c04518c94ad2e1c13e212b)
 - Add a new `split.data.by.bins` function (not to be confused with a previously existing function that had the same name and was renamed in this context), which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698, ed5feb214a123b605c9513262f187cfd72b9e1f4)
 - Add new `assert.sparse.matrices.equal` function to compare two sparse matrices for equality for testing purposes (PR #248, 9784cdf12d1497ee122e2ae73b768b8c334210d4, d9f1a8d90e00a634d7caeb5e7f8f262776496838)
 - Add tests for file `util-networks.misc.R` for issue #242 (PR #248, f3202a6f96723d11c170346556d036cf087521c8, 030574b9d0f3435db4032d0e195a3d407fb7244b, 380b02234275127297fcd508772c69db21c216de, 8b803c50d60fc593e4e527a08fd4c2068d801a48, 7335c3dd4d0302b024a66d18701d9800ed3fe806, 6b600df04bec1fe70c272604f274ec5309840e65)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa)
 
 ### Changed/Improved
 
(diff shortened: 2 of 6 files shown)


szhorvat:
> * the recently released igraph version 2.0.1.1, in which it is not possible any more to add an empty list of vertices

Can you please elaborate on this, and give an example of what changed? This may not be an intentional change.

bockthom:
> Can you please elaborate on this, and give an example of what changed? This may not be an intentional change.

Thanks for having a look at our pull request @szhorvat :smile: Sure, I will elaborate on this:

Up until now, we used the "+" operator and  `igraph::vertices(vertices.to.add, type = type.of.the.vertices.to.add)` to add some vertices to a given graph later on. In this situation, `vertices.to.add` is a vector that could potentially be empty in cases where we don't want to add any vertex, as `vertices.to.add` is the outcome of some other computations. In particular, if we don't want to add any vertex, it happens that `vertices.to.add = character(0)`, which now leads to the following error in the following example:
```R
graph + igraph::vertices(vertices.to.add, type = type.of.the.vertices.to.add)
```
``Error: Error in `add_vertices(e1, la, attr = e2)`: At rinterface_extra.c:78 : Expecting a scalar integer but received a vector of length 2. Invalid value``

As I assumed that `character(0)` may not be allowed any more in such cases (you might have some good reasons for that), I just fixed this on our side by adding an additional check for `length(vertices.to.add) > 0` before executing the statement. If the change in behavior was not intended, I hope my description is helpful for you.

bockthom:
@szhorvat I just encountered another problem in our failing tests which is also related to `+ igraph::vertices`:

```R
author.net =
    igraph::make_empty_graph(n = 0, directed = FALSE) +
    igraph::vertices("A", "B", "C", type = "author", kind = "author") +
    igraph::edges("A", "A", type = "intra", relation = "mail")
```
In previous versions of igraph, this did not fail. However, with the current igraph version, this fails:
`Error in add_vertices(e1, la, attr = e2) : 
  At rinterface_extra.c:78 : Expecting a scalar integer but received a vector of length 2. Invalid value`

What I figured out now is the following: Previously, the attributes have automatically been expanded. That is, if I add 3 vertices but the type attribute contains only one element, then the type-attribute vector was automatically expanded and added as an attribute to each added vertex. Now, the code is failing if the attribute vector is not equal to the number of added vertices. (This might also be the problem in the example that I stated some hours ago, where the attribute vector had length one but zero vertices should be added).
I am not sure if this is an intended change in igraph, but the previous version was much more comfortable in this regard - if you want to add numerous vertices that should all  have the same value for an attribute, then it was fine to just add an attribute vector of length one and having this one value automatically applied to all vertices... Are you going to fix this again or do we need to fix this on our side by expanding the vector length properly?

szhorvat:
So it seems that the problem is that if the number of attribute values provided does not match the number of vertices or edges added, then there's an error, _even if a single attribute value was given_.

This works:

```
g<-make_empty_graph(1) + vertices('a', 'b', foo=c(2,3))
```

This does not work, and rightfully so:

```
g<-make_empty_graph(1) + vertices('a', 'b', foo=c(2,3,4))
```

But one might expect this to assign na attribute value of `5` to both vertices, yet it does not work:

```
g<-make_empty_graph(1) + vertices('a', 'b', foo=5)
```

Am I understanding you correctly?

Pinging @krlmlr to ask if this is indeed expected.

bockthom:
> So it seems that the problem is that if the number of attribute values provided does not match the number of vertices or edges added, then there's an error, _even if a single attribute value was given_.

Correct. 
> 
> This works:
> 
> ```
> g<-make_empty_graph(1) + vertices('a', 'b', foo=c(2,3))
> ```

Sure, this should work.

> 
> This does not work, and rightfully so:
> 
> ```
> g<-make_empty_graph(1) + vertices('a', 'b', foo=c(2,3,4))
> ```

Agreed, this should not work.

> 
> But one might expect this to assign an attribute value of `5` to both vertices, yet it does not work:
> 
> ```
> g<-make_empty_graph(1) + vertices('a', 'b', foo=5)
> ```
> 
> Am I understanding you correctly?

Yes, I totally agree with your understanding.
> 
> Pinging @krlmlr to ask if this is indeed expected.

I am not sure if the behavior with regard to multiple vertices and a single-element attribute vector was expected in past versions of igraph - but it was a nice convenience feature to add multiple vertices at once that should all have the same vertex attributes - then passing just one value for each attribute is much more readable and faster to write than extending all the vectors for all attributes to the number of vertices to add. We used this feature of igraph in our test suite for more than 6 years already. But if you think that this should be changed, it is easy to fix on our side. Nevertheless, I'd also be glad if you can restore the previous behavior in igraph :wink:

krlmlr:
With igraph 1.6.0:

``` r
options(conflicts.policy = list(warn = FALSE))
library(igraph)

g <-
  make_empty_graph(1) +
  vertices("a", "b", foo = 5)

V(g)$foo
#> [1] NA  5  5
```

<sup>Created on 2024-02-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

We clearly had no tests for this, and I agree that this is idiomatic to R.

hechtlC (approved):
Looks good thank you @bockthom 
I will merge this now

Issue #251, opened 2024-02-08
Fix edge-attribute handling when simpliyfing networks


bockthom:
In PR #250, we introduce a new possibility for simplifying multi-relation edges. However, there are no tests yet for this new feature, and also the edge-attribute handling still needs to be adjusted accordingly. Hence, in this issue, I describe the missing steps that should be approached after PR #250 is merged:

Let me start with an example:
Assume that we have an author network containing two vertices A and B, which has multiple relations, namely "mail" and "cochange". Assume that there are three cochange edges and four mail edges between A and B.
Then the default simplification reduced the three cochange edges to one cochange edge, and the four mail edges to one mail edge. The new "simplify.multiple.relations" parameter instead simplifies the three cochange and four mail at edges to one single edges that represents all mail and cochange edges together.

Problem: Edge attributes need to be handled properly for the new case. The edge-attribute handling for simplification is specified here:
https://github.com/se-sic/coronet/blob/a0c3a5262a8837679e2c04343e10b2e41dcab4df/util-networks.R#L55-L70

Currently, we only take the "first" relation - which was not a problem, was only edges of the same relation were simplified to one edge. Now, when we simplify edges of different relations to the same edge, "first" is wrong, as we need a unique vector of all relations instead of just taking the first one. Unfortunately, the edge-attribute handling of igraph does not provide a dedicated function for that. However, igraph allows to provide a custom function that is applied when handling edge attributes during simplification. Hence, such a custom function is necessary for relation here.

However, changing the edge-attribute handling will also break some other parts of coronet, which assume that the edge attribute relation  are only a single element - but now it also can be a vector. For example, this might break the following lines, in which the data sources are determined from the relations - this might break if there are vectors of relations instead of single elements: https://github.com/se-sic/coronet/blob/a0c3a5262a8837679e2c04343e10b2e41dcab4df/util-networks.R#L1673-L1689

It is important, that the new functionality is backwards compatible, this is, if the default simplification procedure is used, then everything should work as before, and only if there are multiple relations combined into a single edge, we need to find a way that all the functions are also able to deal with a vector of relations instead of a single edge.

In addition, PR #250 also lacks of tests. So proper tests should also be added :wink:

MaLoefUDS:
[...]

PR #255, opened 2024-03-07
Work on the simplification of multi-relation edges in networks

MaLoefUDS:
### Prerequisites

### Description

Adjust edge attribute handling slightly to make it compatible with the `simplify.multiple.relations` configuration option introduced in PR #250. In addition, provide more tests for network simplification.

### Changelog

Added:

* Network simplification tests using both approaches

Fixed:

* Fix an issue in `get.data.sources.from.relations` when confronted with networks that include edges that stem from multiple relations (e.g., have been simplified using `simplify.multiple.relations`)

Changed:

* Adjust the default handling of the `relation` edge-attribute when simplifying through `igraph`

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -11,7 +11,7 @@
 ## with this program; if not, write to the Free Software Foundation, Inc.,
 ## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 ##
-## Copyright 2023 by Maximilian Löffler <s8maloef@cs.uni-saarland.de>
+## Copyright 2023-2024 by Maximilian Löffler <s8maloef@cs.uni-saarland.de>
 ## Copyright 2024 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## All Rights Reserved.
 
@@ -41,7 +41,7 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Update system
         run: |
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,9 +10,10 @@
 - Add a new `split.data.by.bins` function (not to be confused with a previously existing function that had the same name and was renamed in this context), which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698, ed5feb214a123b605c9513262f187cfd72b9e1f4)
 - Add new `assert.sparse.matrices.equal` function to compare two sparse matrices for equality for testing purposes (PR #248, 9784cdf12d1497ee122e2ae73b768b8c334210d4, d9f1a8d90e00a634d7caeb5e7f8f262776496838)
 - Add tests for file `util-networks-misc.R` for issue #242 (PR #248, f3202a6f96723d11c170346556d036cf087521c8, 030574b9d0f3435db4032d0e195a3d407fb7244b, 380b02234275127297fcd508772c69db21c216de, 8b803c50d60fc593e4e527a08fd4c2068d801a48, 7335c3dd4d0302b024a66d18701d9800ed3fe806, 6b600df04bec1fe70c272604f274ec5309840e65)
-- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff)
-- Add tests for network simplification (PR #254, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, PR #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff, 58d77b01ecc6a237104a4e72ee5fb9025efeaaf2)
+- Add tests for network simplification (PR #255, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
 - Add `get.bin.dates.from.ranges` function to convert date ranges into bins format (PR #249, a1842e9be46596321ee86860fd87d17a3c88f50f, 858b1812ebfc3194cc6a03c99f3ee7d161d1ca15)
+- Add network simplification to showcase file (PR #255, dc32d44f9db7dfc3cc795ef5d6b86609d6c1936f)
 
 ### Changed/Improved
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,8 @@
 - Add a new `split.data.by.bins` function (not to be confused with a previously existing function that had the same name and was renamed in this context), which splits data based on given activity-based bins (PR #244, ece569ceaf557bb38cd0cfad437b69b30fe8a698, ed5feb214a123b605c9513262f187cfd72b9e1f4)
 - Add new `assert.sparse.matrices.equal` function to compare two sparse matrices for equality for testing purposes (PR #248, 9784cdf12d1497ee122e2ae73b768b8c334210d4, d9f1a8d90e00a634d7caeb5e7f8f262776496838)
 - Add tests for file `util-networks-misc.R` for issue #242 (PR #248, f3202a6f96723d11c170346556d036cf087521c8, 030574b9d0f3435db4032d0e195a3d407fb7244b, 380b02234275127297fcd508772c69db21c216de, 8b803c50d60fc593e4e527a08fd4c2068d801a48, 7335c3dd4d0302b024a66d18701d9800ed3fe806, 6b600df04bec1fe70c272604f274ec5309840e65)
-- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff)
+- Add tests for network simplification (PR #254, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
 - Add `get.bin.dates.from.ranges` function to convert date ranges into bins format (PR #249, a1842e9be46596321ee86860fd87d17a3c88f50f, 858b1812ebfc3194cc6a03c99f3ee7d161d1ca15)
 
 ### Changed/Improved
--- a/showcase.R
+++ b/showcase.R
@@ -23,6 +23,7 @@
 ## Copyright 2021 by Johannes Hostert <s8johost@stud.uni-saarland.de>
 ## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
 ## Copyright 2022 by Jonathan Baumann <joba00002@stud.uni-saarland.de>
+## Copyright 2024 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -360,6 +361,24 @@ g.motifs = motifs.count(network = g,
                                       motif.collaborating.and.communicating = MOTIFS.TRIANGLE.POSITIVE),
                         remove.duplicates = TRUE, raw.data = FALSE)
 
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Network simplification --------------------------------------------------
+
+## construct sample network
+g = y$get.multi.network()
+g = igraph::delete_edges(g, c(5, 6))
+g = igraph::delete_vertices(g, c(2, 4, 5, 6, 7, 8))
+g = g + igraph::edges(c("Björn", "Olaf", "Björn", "Olaf"), type = TYPE.EDGES.INTRA, weight = 1,
+                      relation = "cochange", artifact.type = "Feature")
+
+## merge edges between vertice pairs that stem from the same data source
+g.simplified = simplify.network(g)
+plot.network(g.simplified)
+
+## merge all edges between vertice pairs
+g.simplified = simplify.network(g, simplify.multiple.relations = TRUE)
+plot.network(g.simplified)
 
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Plots -------------------------------------------------------------------
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -84,9 +84,6 @@ test_that("Simplify basic multi-relational network", {
     ## Simplify networks with vertices connected by multi-relational edges
     ##
 
-    ## create network configuration
-    net.conf = NetworkConf$new()
-
     ## create artifact network with vertices connected by "cochange" and "mail edges"
     network =
         igraph::make_empty_graph(n = 0, directed = FALSE) +
@@ -119,7 +116,7 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
     proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
     proj.conf$update.value("commits.filter.base.artifact", FALSE)
     net.conf = NetworkConf$new()
-    net.conf$update.values(updated.values = list(author.relation = c("cochange", "mail")))
+    net.conf$update.values(updated.values = list(author.relation = c("cochange", "mail"), simplify = TRUE))
 
     ## construct objects
     proj.data = ProjectData$new(project.conf = proj.conf)
@@ -171,8 +168,8 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
     network.expected = igraph::graph.data.frame(data, vertices = authors,
                                                 directed = net.conf$get.value("author.directed"))
 
-    ## build network and simplify it
-    network.built = simplify.network(network.builder$get.author.network())
+    ## build simplified network
+    network.built = network.builder$get.author.network()
 
     assert.networks.equal(network.built, network.expected)
 
@@ -180,7 +177,7 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
     ## ---------------------- simplify.multiple.relations == TRUE --------------------------- ##
 
     data = data.frame(comb.1. = c("Björn", "Olaf", "Olaf", "Karl"),
-                    comb.2. = c("Olaf", "Karl", "Thomas", "Thomas"))
+                      comb.2. = c("Olaf", "Karl", "Thomas", "Thomas"))
 
     data$date = list(get.date.from.string(c("2016-07-12 15:58:59", "2016-07-12 16:00:45",   # cochange
                                             "2016-07-12 15:58:40", "2016-07-12 15:58:50")), # mail
@@ -218,8 +215,9 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
     network.expected = igraph::graph.data.frame(data, vertices = authors,
                                                 directed = net.conf$get.value("author.directed"))
 
-    ## build network and simplify it
-    network.built = simplify.network(network.builder$get.author.network(), simplify.multiple.relations = TRUE)
+    ## build simplified network
+    network.builder$update.network.conf(updated.values = list(simplify.multiple.relations = TRUE))
+    network.built = network.builder$get.author.network()
 
     assert.networks.equal(network.built, network.expected)
 
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -13,6 +13,7 @@
 ##
 ## Copyright 2018-2019 by Claus Hunsen <hunsen@fim.uni-passau.de>
 ## Copyright 2021 by Niklas Schneider <s8nlschn@stud.uni-saarland.de>
+## Copyright 2024 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
@@ -918,4 +919,25 @@ test_that("Get the data sources from a network with one relation", {
     network = network.builder$get.author.network()
 
     expect_identical(expected.data.sources, get.data.sources.from.relations(network), info = "data sources: mails")
+})
+
+test_that("Get the data sources from a network with multiple relations on a single edge", {
+    expected.data.sources = c("commits", "mails")
+
+    ## configurations
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("commits.filter.base.artifact", FALSE)
+    ## construct data object
+    proj.data = ProjectData$new(project.conf = proj.conf)
+
+    ## construct network builder
+    net.conf = NetworkConf$new()
+    network.builder = NetworkBuilder$new(project.data = proj.data, network.conf = net.conf)
+    network.builder$update.network.conf(updated.values = list(author.relation = c("mail", "cochange")))
+
+    ## build network
+    network = network.builder$get.author.network()
+    network = simplify.network(network, simplify.multiple.relations = TRUE)
+
+    expect_identical(expected.data.sources, get.data.sources.from.relations(network), info = "data sources: commits, mails")
 })
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -78,6 +78,153 @@ test_that("Simplify network with more than one relation", {
 
 })
 
+test_that("Simplify basic multi-relational network", {
+
+    ##
+    ## Simplify networks with vertices connected by multi-relational edges
+    ##
+
+    ## create network configuration
+    net.conf = NetworkConf$new()
+
+    ## create artifact network with vertices connected by "cochange" and "mail edges"
+    network =
+        igraph::make_empty_graph(n = 0, directed = FALSE) +
+        igraph::vertices("A", "B", type = TYPE.ARTIFACT, kind = "feature")
+    for (i in 1:3) {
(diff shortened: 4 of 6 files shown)


tests/test-networks.R
```diff
@@ -78,6 +78,153 @@ test_that("Simplify network with more than one relation", {
...
 })
 
+test_that("Simplify basic multi-relational network", {
+
+    ##
+    ## Simplify networks with vertices connected by multi-relational edges
+    ##
+
+    ## create network configuration
+    net.conf = NetworkConf$new()
```
bockthom, line 88: Why do you need a network configuration if you don't use a network builder? I guess, the network conf can be removed here.

tests/test-networks.R
```diff
@@ -78,6 +78,153 @@ test_that("Simplify network with more than one relation", {
...
+    ## build network and simplify it
+    network.built = simplify.network(network.builder$get.author.network())
+
+    assert.networks.equal(network.built, network.expected)
+
+
+    ## ---------------------- simplify.multiple.relations == TRUE --------------------------- ##
+
+    data = data.frame(comb.1. = c("Björn", "Olaf", "Olaf", "Karl"),
+                    comb.2. = c("Olaf", "Karl", "Thomas", "Thomas"))
```
bockthom, line 183: Indentation is broken.

bockthom (changes requested):
Thanks for this PR @MaLoefUDS. I have not performed a detailed review yet, but I had a quick look at the tests (without checking them in detail, I just skimmed them).

I have some questions targeting the second test ("Simplify author-network with relation = c('cochange', 'mail') using both algorithms"):

Is there a reason why you call the `simplify.network` function manually and pass the `simplify.multiple.relations` parameter manually to the `simplify.network` function? Isn't it possible to configure this via the recently introduced network-conf parameter `simplify.multiple.relations` (see PR #250) in combination with using `simplify = TRUE` in the network conf?
Maybe I overlook something or this does not belong here, but I would like to see the interplay of network conf and automatic simplification in the network builder also tested properly. If I am not mistaken, this should be easy to adjust in the test. As the automatic simplification calls the `simplify.network` function internally, it should not be necessary to test this function separately, as there is already the basic test above.

And a second idea: `simplify.networks` (the plural version) also uses the the new `simplify.multiple.relations` parameter. Do we already have basic tests for the plural variant? If so, would it be easily adjustable to also test the new parameter in the plural version, to make sure that the parameter is correctly passed to the single version inside?

(In addition, I spotted already two minor issues, see the comments below.)

MaLoefUDS:
Yeah, the only reason I did not use the network conf to set the simplification parameters were a lack of knowledge on my side 😅. 

I just tried it out and I discovered something interesting. Creating a network, then simplifying it does not yield the same result as creating a network with the `simplify` parameter set. In particular, everything that is `NA` when post-simplifying (i.e., attributes of edges that only exist for edges of a specific relation) is `NULL` when simplifying via the config parameter. I could not observe any other changes in output.

I am not sure if that is an actually intended behavior but I guess its not and it needs further investigation. Also, we should decide on which is favored, `NA` or `NULL`.

bockthom:
> I just tried it out and I discovered something interesting. Creating a network, then simplifying it does not yield the same result as creating a network with the `simplify` parameter set. In particular, everything that is `NA` when post-simplifying (i.e., attributes of edges that only exist for edges of a specific relation) is `NULL` when simplifying via the config parameter. I could not observe any other changes in output.

This is really interesting and I'd guess that this behavior is not intended. In general, I've assumed that both ways of simplifying the network should end up in the same network, since the network-construction function of the NetworkBuilder actually calls the simplify function. So, maybe there is something additional going on in the NetworkBuilder after the simplification function is called? That's the only potential difference that comes to my mind. 

> 
> I am not sure if that is an actually intended behavior but I guess its not and it needs further investigation. Also, we should decide on which is favored, `NA` or `NULL`.

Indeed, this needs further investigation. Could you please provide concrete examples of which edge attributes are affected and how their usual values look like? Also checking if there are some additional statements in the NetworkBuilder that could make a difference could be helpful.

Actually, I don't know yet whether `NA` or `NULL` is the expected value. Do we have similar situations in which these attributes are either `NA` or `NULL` (when not using the new functionality) such that we can choose the value that would be most consistent with similar attribute values?

MaLoefUDS:
I have investigated the issue and I think I found a bug in `plyr::rbind.fill` 💀, but lets start a bit earlier.

To get everyone on board, the pipeline of network creation networks with multiple relations is as follows:
1)  Create individual networks for each relation
2) Simplify individual networks
3) Merge networks together into one

So when we create a network with `simplify = TRUE` once we reach step 3, the edges of the individual networks will be simplified, i.e., some attributes will now be of type `list` instead of `character(0)` or something else. For example, two cochange edges between two vertices will be simplified and the `hash` attribute of the simplified hash will be a concatination of the hashes of the individual edges. And it is here were `plyr::rbind.fill` in `merge.network.data` "fails". 

According to its documentation `plyr::rbind.fill` is supposed to fill "... missing columns with NA" [[doc](https://www.rdocumentation.org/packages/plyr/versions/1.8.9/topics/rbind.fill)]. However, it does fill the entries with `NULL` if and only if column entries contain values of type `list`. When we do not specify `simplify = TRUE` and therefore all columns only contain singular elements, missing values will be indeed filled with `NA`. 

This is the end of my bug report and now I will use this comment to answer some of your other questions I have not answered yet:

> And a second idea: simplify.networks (the plural version) also uses the the new simplify.multiple.relations parameter. Do we already have basic tests for the plural variant? If so, would it be easily adjustable to also test the new parameter in the plural version, to make sure that the parameter is correctly passed to the single version inside?

We have exactly 0 tests that test this function, however, adding tests similar to the ones I already added should not be too difficult.

> Indeed, this needs further investigation. Could you please provide concrete examples of which edge attributes are affected and how their usual values look like?

Every mutually exclusive edge attribute between two relations is affected. For example, in mail edges, we have the `message.id` or the `thread` edge attribute, while in cochange networks we have `hash`. When merging mail and cochange edges into one, the attributes values for edges of relation-type one that do not exist in relation-type two will be filled with `NA` or `NULL` respectively.

> Actually, I don't know yet whether NA or NULL is the expected value. Do we have similar situations in which these attributes are either NA or NULL (when not using the new functionality) such that we can choose the value that would be most consistent with similar attribute values?

After revisiting the R documentations of `plyr::rbind.fill`, I think having `NA` is the more reasonable, however, im open for discussions.

MaLoefUDS:
Another problem just came to my mind, but I have not verified it yet. According to the network creation pipeline in my last comment, I do not think `simplify.multiple.relations` as a parameter in the network configuration can even do what is expected from it, as simplification only happens on the individual networks in step 2).

bockthom:
> Another problem just came to my mind, but I have not verified it yet. According to the network creation pipeline in my last comment, I do not think `simplify.multiple.relations` as a parameter in the network configuration can even do what is expected from it, as simplification only happens on the individual networks in step 2).

Indeed, thanks for spotting this! Unfortunately, I just changed the call to a simplification function that is called inside the network creation but before the different networks are merged, my bad.

I guess the problem could be fixed when adding the following statements at two specific places:
```R
if (network.conf$get.value("simplify")) {
        net = simplify.network(net,
                               simplify.multiple.relations = network.conf$get.value("simplify.multiple.relations"))
}
```

(1) After `merge.networks` in `get.author.network` line 758: 
https://github.com/se-sic/coronet/blob/56ff0b33a6fa5e6b704e9cc54c5e7a5b75c3594c/util-networks.R#L758

(2) After `merge.networks` in `get.artifact.network` line 820:
https://github.com/se-sic/coronet/blob/56ff0b33a6fa5e6b704e9cc54c5e7a5b75c3594c/util-networks.R#L820

This way, the networks should be simplified after the different networks are combined.
Adding the simplification there should not change anything in case of single relation networks. Maybe we should extend the condition in the above `if` statement such that we only perform the additional simplification step in case of multiple relations, to only run the additional simplification in cases where it really is necessary, for performance reasons.

My previous change in line 1312, should not affect anything, so I assume we can keep this as I have implemented it? It should not make any difference, if I am not mistaken.

bockthom:
> I have investigated the issue and I think I found a bug in `plyr::rbind.fill` 💀

Puh, that's really interesting. Thanks for the thorough investigations. In general, I agree with you. However, we need to put some more efforts in how these `NULL` values emerge. Maybe the handling of lists only allows `NULL` values there. Thus, we need to find out if there is something we can change to avoid the `NULL` values and get the `NA` values instead. Otherwise, we need to think about whether we can replace the `NULL` values by `NA` values somewhere afterwards. 
And finally, we need to find out whether this is really a bug in `plyr::rbind.fill` or whether this is just an artifact of some special situation. Actually, I don't think that this is a bug in `plyr::rbind.fill`, as I would assume that someone would already have reported this and this has already been fixed if this really is a bug there. 

So, further investigations are necessary before we can decide on how to deal with this problem on our side and whether there really is a bug that we should report to `plyr`. Let's discuss this in our next meeting in more detail.

MaLoefUDS:
Let me concretize my findings: `plyr::rbind.fill` does not fill every column with `NULL`s once there is one column entry that is a list, just all entries belonging to that column where the list object is in. Here is an example:

Consider an arbitrary `data.frame` A and one `data.frame` B that contains columns b1 and b2 that are not included in A. After calling `plyr::rbind.fill` on A and B, we obtain a merged `data.frame` AB. `AB$b1` will be filled with `NA` if not a single value in `B$b1` is a list and it will be filled with `NULL` once there is just one list in `B$b1`. However, if `B$b1` contains a list and `B$b2` does not, `AB$b2` will still be regularly filled with `NA`.

I have also checked the R documentation[[doc](https://www.rdocumentation.org/packages/base/versions/3.6.2)] regarding `NULL` vs. `NA` because I did not really understand the difference. My conclusion is that while both represent a somehow missing or undefined value, `NA` in addition is a logical element, the third, `TRUE`, `FALSE`, and `NA`, while `NULL` is just the total abstinence of a value.

bockthom:
> Consider an arbitrary `data.frame` A and one `data.frame` B that contains columns b1 and b2 that are not included in A. After calling `plyr::rbind.fill` on A and B, we obtain a merged `data.frame` AB. `AB$b1` will be filled with `NA` if not a single value in `B$b1` is a list and it will be filled with `NULL` once there is just one list in `B$b1`. However, if `B$b1` contains a list and `B$b2` does not, `AB$b2` will still be regularly filled with `NA`.

To be honest, I don't get the last sentence here. What's the relation between `B$b1` and `B$b2`?  The behavior regarding `AB$b2` should be independent of `B$b1` anyway. So, I don't get your point.

> I have also checked the R documentation[[doc](https://www.rdocumentation.org/packages/base/versions/3.6.2)] regarding `NULL` vs. `NA` because I did not really understand the difference. My conclusion is that while both represent a somehow missing or undefined value, `NA` in addition is a logical element, the third, `TRUE`, `FALSE`, and `NA`, while `NULL` is just the total abstinence of a value.

True. Or, more colloquially speaking:  `NA` represents an element that is present but for which we don't know its actual value (either because we don't know the value or because we know that its value is different than the remaining pre-existing values). Although `NA` is most often used as a logical, it can also be used in numeric or character contexts to state that a value is not known or cannot be determined. `NULL`, instead, means that an element does not exist at all and is not present. For example, if you set a column of a data frame to `NA`, all the values of these column are set to `NA` in every row. However, if you set column of a data frame to `NULL`, you remove the column and it is not existent anymore afterwards.

.github/workflows/pull_request.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
bockthom, line 44: Great, the pipeline still succeeds but the warnings raised by GitHub actions have disappeared. I'm glad :smile: 

Regarding the commit message: Could you please alter it in such a way that it becomes clear that this commit relates to the CI pipeline / the GitHub Actions workflow? Thanks!

bockthom: And also don't forget about the copyright header in this file...

tests/test-networks.R
```diff
@@ -218,8 +215,9 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
     network.expected = igraph::graph.data.frame(data, vertices = authors,
                                                 directed = net.conf$get.value("author.directed"))
 
-    ## build network and simplify it
-    network.built = simplify.network(network.builder$get.author.network(), simplify.multiple.relations = TRUE)
+    ## build simplified network
+    net.conf$update.values(updated.values = list(simplify.multiple.relations = TRUE))
```
bockthom, line 219: Could you please use `network.builder$update.network.conf(...)` here instead of directly accessing the network conf? (The result is the same, but this makes sure that the reference to the network conf has not changed in the network builder.)

util-networks.R
```diff
@@ -822,6 +828,13 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
             ## set vertex and edge attributes for identifaction
             igraph::V(net)$type = TYPE.ARTIFACT
 
+
```
bockthom, line 831: No need for doubling the blank line. One blank line should be enough here :wink:

bockthom (commented):
Just a few comments on your last two commits, see below.
For the last commit: Could you please adjust the commit message - the commit only cleans up the tests, but the commit message suggests that the actual implementation (not the tests) was changed.

bockthom:
To get an overview of the status of this PR:  The only part that is still missing in this PR are tests for the "plural version" `simplify.networks`, right? Or was there anything else we did talk about for including it in this PR?

MaLoefUDS:
Yes, the tests are on my bucketlist for today. I don't remember anything that we did talk about but not yet implemented it in this PR ^^.

MaLoefUDS:
I was just wondering, since `simplify.networks` (plural) is just a wrapper for `simplify.network`, that we can keep tests simple for the plural version and rather focus on the main function. I think it would be more senseful to test the other parameters of `simplify.network` (e.g., `remove.multiple`) instead of having elaborate `simplify.networks` tests.

bockthom:
I agree that there is no need for "elaborate" `simplify.networks` tests, since it is just a wrapper function, as you already mentioned. But we should make sure that the parameters are correctly passed internally and the output networks are simplified as expected, this is what the new tests should account for.

tests/test-networks.R
```diff
@@ -223,6 +222,49 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
 
 })
 
+test_that("Simplify multiple basic multi-relational networks", {
+
+    ##
+    ## Simplify networks with vertices connected by multi-relational edges
+    ##
+
+    ## create artifact network with vertices connected by "cochange" and "mail edges"
```
bockthom, line 231: the last quotation mark should be after "mail" and not after "edges" :wink:

tests/test-networks.R
```diff
@@ -223,6 +222,49 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
...
+        igraph::edges("A", "B", type = TYPE.EDGES.INTRA, relation = "cochange")
+    network.B.expected = igraph::make_empty_graph(n = 0, directed = FALSE) +
+        igraph::vertices("C", "D", type = TYPE.AUTHOR, kind = TYPE.AUTHOR) +
+        igraph::edges("C", "D", type = TYPE.EDGES.INTRA, relation = "mail") +
+        igraph::edges("C", "D", type = TYPE.EDGES.INTRA, relation = "cochange")
+
+    ## simplify networks without simplifying multiple relations into single edges
+    networks.simplified = simplify.networks(list(network.A, network.B), simplify.multiple.relations = FALSE)
+    assert.networks.equal(networks.simplified[[1]], network.A.expected)
+    assert.networks.equal(networks.simplified[[2]], network.B.expected)
```
bockthom, line 257: Two ideas: 
- Shouldn't we make sure that the resulting list has the same size as the given list? Currently we only test for two elements but ignore potential other elements...
- I am not sure what happens with named lists. Are the names preserved or not? If they are preserved, we should also check their correctness. (If they are not preserved, just ignore my comment.)

tests/test-networks.R
```diff
@@ -223,6 +222,49 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
...
+
+    ## simplify network with simplifying multiple relations into single edges
+    networks.simplified = simplify.networks(list(network.A, network.B), simplify.multiple.relations = TRUE)
```
bockthom, line 260: The comment is wrong, it should be "networks" instead of "network"

bockthom (commented):
Just a few comments below–I only performed a "syntactical" review, not yet a "semantic" one, that is, I did not yet think about the test scenarios yet.

tests/test-networks.R
```diff
@@ -78,6 +78,193 @@ test_that("Simplify network with more than one relation", {
...
+
+    ## simplify network without simplifying multiple relations into single edges
+    network.simplified = simplify.network(network, simplify.multiple.relations = FALSE)
+    assert.networks.equal(network.simplified, network.expected)
+
+    ## simplify network with simplifying multiple relations into single edges
+    network.simplified = simplify.network(network, simplify.multiple.relations = TRUE)
+    expect_identical(igraph::ecount(simplify.network(network.simplified)), 1)
+    expect_identical(igraph::E(network.simplified)$type[[1]], "Unipartite")
+    expect_identical(igraph::E(network.simplified)$relation[[1]], c("cochange", "mail"))
```
bockthom, line 109: I am wondering, why are we not testing any of the other edge attributes in the basic test? E.g., using `date`? 
I'd suggest to add a date to all edges in all basic simplification tests (also the basic ones that have existed already).

And also the `weight` should be tested for simplification functions, in general (not only the special cases).
Would it be possible to check for the weight in all of the tests? If multiple edges are contracted to one edge, the weight attribute should automatically represent the number of previous edges, or the sum of previous weights if there was already a weight > 1 before. 
With that, we should have handled all different kinds of attribute handling.

tests/test-networks.R
bockthom, line 1: Also, please don't forget updating the copyright header in this file.

util-networks.R
```diff
@@ -1786,7 +1798,7 @@ delete.authors.without.specific.edges = function(network, specific.edge.types =
 #'         empty relation, i.e. \code{character(0)}
 get.data.sources.from.relations = function(network) {
     ## get all relations in the network
-    data.sources = unique(igraph::E(network)$relation)
+    data.sources = unique(unlist(igraph::E(network)$relation))
```
bockthom, line 1801: Is `get.data.sources.from.relations` used somewhere in tests?
Would it be possible to test this with the new simplification method, that is, having multiple relations at one edge? Only if this is doable without much effort.

bockthom (changes requested):
I'm done with my review now. Looks good to me.
I've a few suggestions on how to improve the tests with regard to other attributes etc. If you think my suggestions are beneficial improvements to the tests and easy to realize, I would appreciate it if you could extend the tests a little bit.
If you have good arguments why my suggestions are unnecessary, I am also open for them :smile:

tests/test-networks.R
MaLoefUDS: Hmm, in my opinion all edge attributes but `relation` use default handling values like `first`, `sum` etc., therefore adding them in our tests would just boil down to testing if `igraph` works. I see that working `igraph` is not a given at any time (💀)  but im not sure whether we need to test their behavior.

util-networks.R
MaLoefUDS: There are 2 tests that test this functions, that already aim to cover possible side cases. Adding a new test case should not be of any problem.

MaLoefUDS:
Im unsure whether the `showcase.R` file really needs an update regarding the new simplification option. Upon inspection, simplification is currently not an explicitly showcased feature. We obtain a sample network, simplify it, then plot it and continue. We also only do that one single simplification, we do not play with parameters later on or something. Therefore, I don't think we need an update but please let me know if you think differently 😄.

bockthom:
> Im unsure whether the `showcase.R` file really needs an update regarding the new simplification option. Upon inspection, simplification is currently not an explicitly showcased feature. 

just because it's not an explicitly showcased feature now doesn't mean it has to stay that way :wink: 

> We obtain a sample network, simplify it, then plot it and continue. We also only do that one single simplification, we do not play with parameters later on or something. Therefore, I don't think we need an update but please let me know if you think differently 😄.

We don't really need an update, but I still think it can be beneficial to demonstrate the new feature there.
Maybe the plotting section is not the right place, but maybe you can add a new section directly above the plotting section in which you demonstrate the two different simplification methods on a single multi-relation network? Just call "plot.network(...)" on each of them and it should be clearly visible to everyone what the difference between the two simplification options is.
If you want to avoid code duplication, you can also reuse the created network in the plotting section afterwards.

NEWS.md
```diff
@@ -10,7 +10,8 @@
...
 - Add tests for file `util-networks-misc.R` for issue #242 (PR #248, f3202a6f96723d11c170346556d036cf087521c8, 030574b9d0f3435db4032d0e195a3d407fb7244b, 380b02234275127297fcd508772c69db21c216de, 8b803c50d60fc593e4e527a08fd4c2068d801a48, 7335c3dd4d0302b024a66d18701d9800ed3fe806, 6b600df04bec1fe70c272604f274ec5309840e65)
-- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff)
+- Add tests for network simplification (PR #254, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
```
bockthom, line 14: Wrong PR number: should be #255 here.

NEWS.md
```diff
@@ -10,7 +10,8 @@
...
 - Add new `assert.sparse.matrices.equal` function to compare two sparse matrices for equality for testing purposes (PR #248, 9784cdf12d1497ee122e2ae73b768b8c334210d4, d9f1a8d90e00a634d7caeb5e7f8f262776496838)
 - Add tests for file `util-networks-misc.R` for issue #242 (PR #248, f3202a6f96723d11c170346556d036cf087521c8, 030574b9d0f3435db4032d0e195a3d407fb7244b, 380b02234275127297fcd508772c69db21c216de, 8b803c50d60fc593e4e527a08fd4c2068d801a48, 7335c3dd4d0302b024a66d18701d9800ed3fe806, 6b600df04bec1fe70c272604f274ec5309840e65)
-- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff)
```
bockthom, line 13: Could you please add a `PR` directly before `# 255` to differentiate between issues and PRs. Without `PR` it means "issue". In addition, you could reference issue `#251`.

bockthom (changes requested):
Only two small comments regarding the NEWS.md:

MaLoefUDS:
Regarding the `showcase.R`: `plot.network` cannot handle lists in relation attributes as produced by our new simplification. I will need to see why ...

util-plot.R
```diff
@@ -142,7 +143,7 @@ plot.get.plot.for.network = function(network, labels = TRUE) {
     if (igraph::ecount(network) > 0) {
         p = p +
             ggraph::geom_edge_fan(
-                mapping = ggplot2::aes(colour = relation, linetype = edge.type, width = 0.3 + 0.5 * log(weight)),
+                mapping = ggplot2::aes(colour = paste(relation, " "), linetype = edge.type, width = 0.3 + 0.5 * log(weight)),
```
bockthom, line 146: I am not sure what this does exactly, but I am glad if it works. I will try to view the resulting plots in a couple of minutes to see how it looks like and will comment on this again.

Independent from that, could you please add the parameter name `sep = " "` for better understandability? Thanks!

As this is related to the new simplification method, the commit hash of this commit should be added to the NEWS.md, item "Add the possibility to simplify edges of multiple-relation networks...", I guess.

showcase.R
```diff
@@ -361,6 +362,24 @@ g.motifs = motifs.count(network = g,
                         remove.duplicates = TRUE, raw.data = FALSE)
 
 
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Network simplification --------------------------------------------------
+
+## construct sample network
+g = y$get.multi.network()
+g = igraph::delete_edges(g, c(5,6))
+g = igraph::delete_vertices(g, c(2,4,5,6,7,8))
```
bockthom, line 371: Please add spaces after the all the commas.

NEWS.md
```diff
@@ -10,9 +10,10 @@
...
-- Add tests for network simplification (PR #254, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
+- Add the possibility to simplify edges of multiple-relation networks into a single edge at all instead of a single edge per relation (PR #250, PR #255, 2105ea89b5227e7c9fa78fea9de1977f2d9e8faa, a34b5bd50351b9ccf3cc45fc323cfa2e84d65ea0, 34516415fed599eba0cc7d3cc4a9acd6b26db252, 78f43514962d7651e6b7a1e80ee22ce012f32535, d310fdc38690f0d701cd32c92112c33f7fdde0ff)
+- Add tests for network simplification (PR #255, 338b06941eec1c9cfdb121e78ce0d9db6b75da19, 8a6f47bc115c10fbbe4eee21985d97aee5c9dc91, e01908c94eccc4dda5f2b3c0746b0eab0172dc07, 7b6848fb86f69db088ce6ef2bea8315ac94d48f9)
 - Add `get.bin.dates.from.ranges` function to convert date ranges into bins format (PR #249, a1842e9be46596321ee86860fd87d17a3c88f50f, 858b1812ebfc3194cc6a03c99f3ee7d161d1ca15)
+- Add network simplification to showcase file (PR #250, abcc26dcfea3bd69be5be0934ed184c5416bc081)
```
bockthom, line 16: Shouldn't this be PR #255 instead of 250?

bockthom (changes requested):

bockthom:
> Regarding the `showcase.R`: `plot.network` cannot handle lists in edge attributes as produced by our new simplification. I will need to see why ...

Great! First of all, thanks for spotting this bug!  Nobody would ever have thought of fixing the plotting functionality to deal with edges that represent multiple relations.  I've said it many times already: Playing around with the `showcase.R` really is beneficial for identifying potential problems with new features. That proved to be true again here :smile:

> 
> Update (2/2):
> 
> I tried concating the relations and it seems to work now, I will finish up a commit so you can have a look.

Thanks for solving the problem. I really like your fix. I've tried plotting networks with multiple different combinations of relations on single edges, and all the different combinations get different edge colors, as intended.
When just looking at the code, I was wondering what you fix really does – but it produces the correct solution. Just R things... Good work @MaLoefUDS!

bockthom (approved):

tests/test-networks.R
```diff
@@ -78,6 +79,198 @@ test_that("Simplify network with more than one relation", {
...
+        igraph::edges("A", "B", type = TYPE.EDGES.INTRA, relation = "mail") +
+        igraph::edges("A", "B", type = TYPE.EDGES.INTRA, relation = "cochange")
+
+    ## simplify network without simplifying multiple relations into single edges
+    network.simplified = simplify.network(network, simplify.multiple.relations = FALSE)
+    assert.networks.equal(network.simplified, network.expected)
+
+    ## simplify network with simplifying multiple relations into single edges
+    network.simplified = simplify.network(network, simplify.multiple.relations = TRUE)
+    expect_identical(igraph::ecount(simplify.network(network.simplified)), 1)
```
bockthom, line 108: Although this is already merged code, I have a question @MaLoefUDS: 
Was it intended to simplify the `network.simplified` here in line 108 again?
The same holds for line 268 below.

MaLoefUDS: Yeahhh both of those occurrences stem from copy-pasting and require a change. What is your preferred way to include a fix? Should I open a new PR or may I include it in the still open PR #257?

bockthom: I'd prefer to include it in PR #257. There are reasons why I left the PR #257 open :wink:

PR #263, opened 2024-05-21
Add Commit Networks

Leo-Send:
### Prerequisites

### Description
Add a new type of network to coronet: Commit Network
The commit network uses commits as vertices. The edges are based on commit interactions or cochange.

### Added
- Add commit network as a new type of network. It uses commits as vertices and connects them either via cochange or commit interactions. This includes adding new config parameters and a function for adding vertex attributes to a commit network(PR #263, ab73271781e8e9a0715f784936df4b371d64c338, ab73271781e8e9a0715f784936df4b371d64c338, cd9a930fcb54ff465c2a5a7c43cfe82ac15c134d)

### Changed7Improved
- Refactor function `construct.edge.list.from.key.value.list` to be more readable(PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)

### Fixed
- Fix networks based upon commit interaction data to also have the attribute `artifact.type`(PR #263, 849123a8b7d898fbb1343745ecffc1f6000c9367)
- Fix endless recursion that could occur when commit interaction data was configured and commit data is empty (PR #263, 3fb7437b68950303916b62984fa449732c70353e)


Leo-Send:
Currently the part in the showcase is still missing, I will look into where and what to add there.

Leo-Send:
Currently the networks contain very little information, because most information typically is 
- in the commit data
- shown on the edges of the network

Because in this case the commits are the vertices of the network, that information is not portayed anywhere - meaning the date, author information and commit-id have to be looked up seperately from the network using the commit hash.

Should I include more vertex attributes that also contain this information or some other functionality to more easily retrieve this data?

Leo-Send:
In the network construction for the commit cochange network, respect.temporal.order is currently always set to 'TRUE'. This matches the construction of the artifact cochange network, that's why I did the same. Should that be the case?

bockthom:
@Leo-Send Could you please rebase your branch onto the current `dev` branch? Otherwise the CI pipeline will not work (for two reasons: one the one hand, there are conflicting changes in `util-networks-covariates.R`; on the other hand, we have made some changes to the CI pipeline anyway)

util-networks.R
```diff
@@ -679,6 +685,90 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+
+            logging::logdebug("get.commit.network.commit.interaction: starting.")
+
+            ## do not compute anything more than once
+            if (!is.null(private$commit.network.commit.interaction)) {
+                logging::logdebug("get.commit.network.commit.interaction: finished. (already existing)")
+                return(private$commit.network.commit.interaction)
+            }
+
+            ## get the authors that appear in the commit-interaction data as the vertices of the network
```
hechtlC, line 701: The hashes, correct? Because you are not getting authors here.

util-networks.R
```diff
@@ -679,6 +685,90 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+            vertices = unique(c(private$proj.data$get.commit.interactions()[["base.hash"]],
+                                private$proj.data$get.commit.interactions()[["commit.hash"]]))
+            vertices = data.frame(name = vertices)
+
+            ## get the commit-interaction data as the edge data of the network
+            edges = private$proj.data$get.commit.interactions()
+            ## set the commits as the 'to' and 'from' of the network and order the dataframe
+            edges = edges[, c("base.hash", "commit.hash", "func", "interacting.author",
+                              "file", "base.author", "base.func", "base.file")]
+            edges = cbind(edges, data.frame(artifact.type = c("CommitInteraction")))
```
hechtlC, line 711: Why do you use cbind here and not just append the column `artifact.type` to the edges data frame?

util-networks.R
```diff
@@ -1214,146 +1353,213 @@ construct.edge.list.from.key.value.list = function(list, network.conf, directed
...
+        item.edge.attrs = item[ , edge.attributes[cols.which], drop = FALSE]
 
-            ## break if there is no author
-            if (length(vertices) < 1) {
-                return(NULL)
-            }
+        ## construct edges
+        combinations = c()
+        if (network.type == "commit") {
+            combinations = expand.grid(item.vertex[["commit"]], vertices.processed.set[["commit"]], stringsAsFactors = FALSE)
```
hechtlC, line 1461: This line is too long

util-networks.R
```diff
@@ -1214,146 +1353,213 @@ construct.edge.list.from.key.value.list = function(list, network.conf, directed
...
+            combinations = expand.grid(item.vertex, vertices.processed.set, stringsAsFactors = FALSE)
+        }
 
-            ## if there is only one author, just create the vertex, but no edges
-            if (length(vertices) == 1) {
-                edges = data.frame()
-                attr(edges, "vertices.processed") = vertices # store set of processed vertices
-                return(edges)
-            }
+        if (nrow(combinations) > 0 & nrow(item.edge.attrs) == 1) {
```
hechtlC, line 1466: Please use &&

Leo-Send:
I have Identified the issue with the showcase: since we manually add the column 'artifact.type' to dataframes when constructing commit interaction based networks, it breaks when the existing data is empty. The empty data was not an issue for the showcase before; there were just warnings but since we do not do anything with the networks anyway, everything worked. Now, I see two ways to fix this problem:
1) Adding commit interaction data in the data path that the showcase uses (tests/codeface-data/results/testing/test_feature/feature)
2) Checking in the network construction if the data is empty before adding a column

Which would you prefer @hechtlC ?

bockthom:
I don't know what @hechtlC prefers, but here a few thoughts from my side on your two suggestions:

>     2. Checking in the network construction if the data is empty before adding a column

I don't fully understand what this column is for. Is it just used for network construction or is it also available to the end user? If it is available for the end user, I would like to make sure that the expected columns are there, independent of empty data or not. If it is just used internally for network construction and does not have any effect on the resulting network, then a check for empty data before adding a column might make sense (but I am not sure).

>     1. Adding commit interaction data in the data path that the showcase uses (tests/codeface-data/results/testing/test_feature/feature)

Regarding adding commit interaction data: I am not sure. Is it just an issue of the non-existent file, or of the non-existent data (as it would also be in an empty file)? Missing data should not lead to failing network construction. So, even when adding more test data would solve the problem, I don't think that this should be considered as a fix for the problem.

Leo-Send:
For some reason, adding the line 
>proj.conf$update.value("commit.interactions", TRUE)

breaks the splitting, meaning that the content of 'cf.data' after
> mybins = c("2012-07-10 15:58:00", "2012-07-15 16:02:00", "2012-07-20 16:04:00", "2012-07-25 16:06:30")
cf.data = split.data.time.based(x.data, bins = mybins)

is total gibberish. This seems to be a problem in the splitting - I believe commit interaction data should be excluded from any splitting as it is not annotated with the relevant data for splitting anyway.

bockthom:
> For some reason, adding the line
> 
> > proj.conf$update.value("commit.interactions", TRUE)
> 
> breaks the splitting, meaning that the content of 'cf.data' after
> 
> > mybins = c("2012-07-10 15:58:00", "2012-07-15 16:02:00", "2012-07-20 16:04:00", "2012-07-25 16:06:30")
> > cf.data = split.data.time.based(x.data, bins = mybins)
> 
> is total gibberish. This seems to be a problem in the splitting - I believe commit interaction data should be excluded from any splitting as it is not annotated with the relevant data for splitting anyway.

I agree with @Leo-Send: commit-interaction data should not be split. Therefore, commit-interaction data should be considered as additional data sources that are ignored by splitting and added to each range data object as a whole, such as pasta data or author data. According to the following line, commit-interaction data are correctly categorized as additional data sources:
https://github.com/se-sic/coronet/blob/74ebe0bfeb3145dace92fca3e505a5f3ffd85385/util-data.R#L1895-L1902

So, there might be any other space where commit-interaction data are considered to be in the wrong category... Maybe there is a typo somewhere (using the plural version instead of the singular, or something similar leads to be not recognized correctly...)

Is this something that has worked correctly before your rebase today? Then we might find the problem in the previous PR(s) that have been merged on `dev` in the meanwhile. Otherwise, the problem could be located anywhere else in the data or splitting module...

If you could trace down the problem until where splitting breaks, we could have a closer look at that particular part of the code. Our expert for splitting is @MaLoefUDS, maybe he also has an idea on what's going on there?

Leo-Send:
I am even more confused now - checking older commits I found that adding this line to the showcase has broken the splitting since commit interactions were added. But I added the commit network using commit interactions to the showcase a month ago - and I am very sure that I tried it and the CI also worked (apart from the old R version that is now removed). Since I rebased and force-pushed, I do not now how to access the CI run reports from old commits, SO I cannot confirm it currently.

bockthom:
> I will look into how to exclude commit interaction data from splitting.

As already said yesterday: Commit-interaction data are already in the correct category that is excluded from splitting, as the following snippet shows:
https://github.com/se-sic/coronet/blob/74ebe0bfeb3145dace92fca3e505a5f3ffd85385/util-data.R#L1895-L1902

So, either there is some typo somewhere in the used spelling variants of "commit interaction(s)", or there is something else going on that is totally weird. Again, if you can provide the exact line at which it breaks, we can have a look why this line is executed at all.

Leo-Send:
Okay so i found some weird stuff going on. As I am not an expert on how splitting works, please correct me if I am misunderstanding something.

For this bug to occur in the showcase, you need to copy the 'commit.ineractions.yaml' from the testing/test_proximity/proximity folder to the testing/test_feature/feature and subsequent 002-v2-v3 folder.

If you want to replicate my 'testing without commit interactions in showcase', comment out line 78 and change the value in line 69 to "cochange".

1. When testing without commit interactions in showcase and setting 'browser()' statements in the function 'split.data.by.time.or.bins', I found that the 'Re-arranging data' part around line 1025 sets the content of 'data.split' to NULL. It does this regardless of the commit interactions. Is this intended?

2. The only difference between the construction of the RangeData objects in the two runs with and without commit interaction data is the content of 'additional.data.sources'. I am currently looking into the possibility that there is an infinite recursion somewhere.

bockthom (changes requested):
Thanks for this PR @Leo-Send. I have reviewed your changes (except for the tests and showcase) and found a couple of minor inconsistencies or cases in which I am not sure about whether the current implementation is consistent or not. Please see my detailed comments below. Please don't see every comment as a need to change something - in some cases I just would like you to check similar cases in the existing implementation before deciding on whether we should change something or  not.

In addition, could you please rebase the last two commits? 

- The second sentence of the commit message of "[Add 'artifact.type' to commit interaction data](https://github.com/se-sic/coronet/pull/263/commits/4e56de09c879b2171c293fed49752ab7fdaca06c)" contains the word "until" twice; and I also don't understand the sentence. Could you please rephrase that?
- [Fix crash in showcase and add update tests](https://github.com/se-sic/coronet/pull/263/commits/f051730c51c54d0892702f6fd7b61c68f7a2385a): I guess this commit will not be present in the final PR any more, but I would like to mention it though: First, the "add update tests" part sounds broken. Second, the commit message does neither describe what's going on there, nor what the problem is that is fixed by this commit. "Crash in showcase" could be everything, but the problem is not the showcase, but something else in coronet that causes the showcase to crash ;) But I guess this will change anyway after we have figured out the actual problem there.

bockthom:
> There is indeed an infinite recursion, although I do not understand why. The setting (or getting) of commit interactions calls 'update.commit.interactions' which in turn calls 'get.commits' which calls 'set.commits' which again calls 'update.commit.interactions'. This SHOULD be prevented by caching (as it does normally... In my tests for commit interactions there is no infinite recursion and I also call 'get.commit.interactions' on a new projectData object). The only difference I can spot is that of RangeData vs ProjectData, although I am unsure if that can have anything to do with it.

In general, it should not make a difference whether you use a RangeData or a ProjectData object. But maybe there is some inconsistency regarding caching additional data sources. I suggest to check how the different additional data sources are  handled? Maybe there is a different behavior because commit-interaction data are linked to the commit data, which is not the case for author data, for example. Maybe the connection between commit data and commit-interaction data leads to this infinite loop? Maybe you could also check what's different to how PaStA data are handled (but, if I am not mistaken, the relationship between PaStA and commit data is the other way round, which might not trigger the problem...)

Unfortunately, I have no further ideas.

MaLoefUDS:
Hey, sorry for the late response. I have some nice insight on the issue regarding the splitting in `showcase.R` that "breaks" which @Leo-Send found here:

>When testing without commit interactions in showcase and setting 'browser()' statements in the function 'split.data.by.time.or.bins', I found that the 'Re-arranging data' part around line 1025 sets the content of 'data.split' to NULL. It does this regardless of the commit interactions. Is this intended?

First of all, this does have nothing to do with the selected data source or if commit interactions are present (as Leo already mentioned). Interestingly this broken application has been in `showcase.R` for a really long time. I guess nobody really cared because the only use of the split data is plotting it once, which does not break directly.

The reason for the break is that for some data sources, some of their elements are before the predefined bins by which we split. When we split by given bins, we use the `base::findInterval` method to separate all data elements into the corresponding bins they belong, i.e., it returns a vector of integers, of which each integer describes the index of the bin in which we should place an item (e.g., `1 1 1 2 2 3 3 3 3 3 6`). Then we split according to this vector and rename all splits according to the datestring of the bin.

However, the `findInterval` function returns `0` for datapoints that do not lie in any bin (i.e., before the first bin). Consequently, when trying to find the name for said split, we index into the vector of bin-labels with 0 resulting in `character(0)`. This in the end leads to the breakage observed. This might actually require a fix from our side. I believe, spitting should not break like this when some elements are not in any bin.

bockthom:
Good catch @MaLoefUDS! Thanks for looking into this! I agree with you that splitting should not break when there are elements that are not part of a bin. I guess that this can happen in two different ways: Prior to the first bin, and after the last bin. Do we already handle the case "after the last bin" correctly? If so, we could try to handle "prior to the first bin" similarly. If we don't handle either of them up until now, I suggest to open a new issue for that and discuss potential solutions there, to not mess up this PR more than we currently already do...

@Leo-Send: Independent from the splitting bug that @MaLoefUDS has identified, the infinite recursion seems to be a different problem that also needs to be fixed... Is there a way to fix this problem independent of the splitting bug, or do the two problems depend on each other? That is, does the infinite recursion occur depending on the splitting bug, or does it occur independently?

Leo-Send:
I found the reason for the infinite recursion and it does not seem to be directly related to what @MaLoefUDS found, although I believe it both originates from the problem that we do not verify the bins when splitting.

This infinite recursion is supposed to be stopped by checks if the data source is already cached - the method get.commits first sets the field for storing commits and then calls update.commit.interactions, thus does not get called in update.commit.interactions again, as commits are cached. This does not, however, work if the commit data is empty. This specific call to splitting that is in the showcase results in one RangeData to have empty commit data, thus leading to this recursion.

I see two ways to handle this:
1. disallow bins which would result in empty commit data. I believe this should be done if it is generally unwanted to ever have empty commit data.
2. remove the potential for this circular recursion entirely, by checking who the caller of the function was. This would of course only handle this problem in commit interactions and not any other problem that might arise when commit data is empty.

MaLoefUDS:
Regarding fixing this problem of broken splitting which may or may not be the sole cause of Leo's problems, I have two possible fixes in mind.

But first of: Currently we do not have any problems with data elements that lie beyond the last bin. The bins we provide are always just the first date bound. The next following bin is then the last date bound, while also being the first date of its own new bin. In the current implementation, the last bin is then holds all elements from this point on until infinity. Obviously, we cannot simply include all elements before the first bin into the first bin as well, as this would lead to unintuitive behavior (i.e., what would it mean to provide only one bin when splitting?).

Here are my ideas for fixes:

1. Ignore all elements before the first bin and disregard them. We can simply filter the bin assignment vector and filter out all `0`s.
2. Create an artificial bin before the first actual bin with some made up name (?) that holds all elements before the first bin. This is also easy to implement as we can just rename the bin of `character(0)` to `whatever`.

I don't have any preference on this, since I am not sure which idiomatic values of coronet are at stake 😅

bockthom:
Now it is getting confusing here. We discuss two different problems and both of you, @Leo-Send and @MaLoefUDS, have suggested two potential solutions for one of the two problems...

<hr>

Let's start with the infinite recursion spotted by @Leo-Send:

> I see two ways to handle this:
> 
>     1. disallow bins which would result in empty commit data. I believe this should be done if it is generally unwanted to ever have empty commit data.
> 
>     2. remove the potential for this circular recursion entirely, by checking who the caller of the function was. This would of course only handle this problem in commit interactions and not any other problem that might arise when commit data is empty.

Option 1. is **not** a valid solution. There always can be time ranges without any commit and we still want to keep this range.

Option 2. sounds promising. When choosing option 2, we also need to check whether there are other cases which lead to a similar problem. Either we are able to find a common solution that handles all these cases, or we need to handle each of them separately.

Maybe you can think about that until our meeting, and then let's discuss the potential options in our meeting.

<hr>

Regarding the problems occurring during splitting @MaLoefUDS:

We need to figure out in which situations the elements before the first bin are actually needed. This is the key question to this problem, and I don't have an answer to this question. If there are no such situations, we could potentially remove them. Otherwise we need to find a way to keep them. But first we need to figure out whether there is any use case in which the elements before the first bin are relevant.

As this problem seems to be independent of @Leo-Send's problem (as the infinite recursions can occur on any empty commit data), let's continue the discussion on splitting in a separate issue.

bockthom:
I've just transferred the comments that discuss the splitting problem into a new issue #267. Please continue discussing the splitting problem there.

Let's stick to the changes of this PR and the discussion of the infinite recursion here.

util-data.R
```diff
@@ -2143,6 +2143,35 @@ ProjectData = R6::R6Class("ProjectData",
...
+
+        #' Group the commits of the given \code{data.source} by the given \code{group.column}.
+        #' For each group, the column \code{"hash"} is duplicated and prepended to each
+        #' group's data as first column (see below for details).
+        #'
+        #' Example: To obtain the commits that changed the same source-code artifact,
+        #' call \code{group.commits.by.data.column("commits", "artifact")}.
+        #'
+        #' @param data.source The specified data source. One of \code{"commits"},
+        #'                    \code{"mails"}, and \code{"issues"}. [default: "commits"]
```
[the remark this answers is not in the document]
Leo-Send: Good catch, I didn't notice this made little sense. Since the 'hash' column does not exist in neither mail nor issue data, you can only call the method using commit data. I will change it so that the first parameter of the method is removed entirely.

util-networks-covariates.R
```diff
@@ -140,6 +141,42 @@ add.vertex.attribute = function(net.to.range.list, attr.name, default.value, com
...
+#' Utility function to add a vertex attribute from commit data to a commit network.
+#' Attribute name should be a column name of the commit data dataframe.
+#' Default column names can be seen in 'COMMITS.LIST.COLUMNS' in 'util-read.R',
+#' though more might be possible.
+#'
+#' @param network the commit network
+#' @param project.data the project data from which to extract the values
+#' @param attr.name the name of the attribute
+#' @param default.value the dafault value of the attribute
+#'                      if it does not occur in the commit data
```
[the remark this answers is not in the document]
Leo-Send: I changed it to 'the default value of the attribute if the commit data contains no data for that vertex regarding the attribute'. Is that better?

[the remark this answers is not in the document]
Leo-Send: Actually that would be incorrect - the last part 'regarding the attribute' is wrong. The default value only occurs if a hash is not contained in the commit data at all.

util-networks.R
```diff
@@ -352,7 +358,7 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
                 network.conf = private$network.conf,
                 directed = FALSE,
                 respect.temporal.order = TRUE,
-                artifact.edges = TRUE
+                network.type = "artifact"
```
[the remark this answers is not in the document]
Leo-Send: Before, this method got a boolean flag to distinguish if it was constructing an edge list for an artifact network or not, because it's behaviour then needs to change slightly. Now, the method also changes when it constructs a commit network, therefore the parameter changed from a boolean to a string.

util-networks.R
```diff
@@ -245,6 +248,9 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
             colnames(edges)[1] = "to"
             colnames(edges)[2] = "from"
             colnames(edges)[4] = "hash"
+            if (nrow(edges) > 0) {
+                edges[["artifact.type"]] = "CommitInteraction"
+            }
```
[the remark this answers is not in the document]
Leo-Send: In the other networks, the artifact.type attribute is added during 'construct.edge.list.from.key.value.list'. Since networks based on commit interactions do not use this method and commit interaction data does not contain this information, it is added manually here.

I can add a constant for 'CommitInteraction' if wanted, and also do the same for 'File' and 'Function' if wanted.

util-networks-covariates.R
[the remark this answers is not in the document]
bockthom: Why not using the last part of your last sentence for the parameter description?
Something like "the default value that is used if a hash is not contained in the commit data at all"

util-networks.R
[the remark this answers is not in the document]
bockthom: Ah, thanks for the clarification. 
Could you please enhance the parameter description of `construct.edge.list.from.key.value.list` to make the possible values more obvious (I know, they are contained in the parameter definition; but it would also be useful the enhance the parameter description) as follows:

```R
#' @param network.type the type of network for which the key value data is to be used as edges
#'                     (one out of "author", "artifact", or "commit") [default: "author"]
```

showcase.R
```diff
@@ -232,6 +239,8 @@ sample.pull.requests = add.vertex.attribute.author.issue.count(my.networks, x.da
 ## add vertex attributes for the project-level network
 x.net.as.list = list("1970-01-01 00:00:00-2030-01-01 00:00:00" = x$get.author.network())
 sample.entire = add.vertex.attribute.author.commit.count(x.net.as.list, x.data, aggregation.level = "complete")
+## add vertex attributes to commit network
+add.vertex.attribute.commit.network(x$get.commit.network(), x.data, "author.name", "NO_AUTHOR")
```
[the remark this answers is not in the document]
Leo-Send: We do not have a constant like that.
I added a comment, the whole line now reads 'add vertex attributes to commit network. Default value is used if vertex is not in commit data' and also added the parameter names to the function call.

util-networks.R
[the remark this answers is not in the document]
Leo-Send: I agree in general, however the use of this constant would be a bit weird here. We would need to do something like ARTIFACT.CODEFACE[["commit.interaction"]], since there is no corresponding config parameter that we can put into ARTIFACT.CODEFACE.
I would thus suggest creating a new constant that just reads "CommitInteraction".

showcase.R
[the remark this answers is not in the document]
bockthom: Maybe use "Default value "NO_AUTHOR" is used...". The rest sounds good to me.

util-networks.R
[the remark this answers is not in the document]
bockthom: Hm, I see your point.
I would like to have "CommitInteraction" as part of `ARTIFACT.CODEFACE` or directly below this constant as a second constant, to make the connection between the two obvious. For me, it would be even ok to use `ARTIFACT.CODEFACE[["commit.interaction"]]`, even if this looks weird, to just to have all the related constants in one variable - but I also see the disadvantages. Let's @hechtlC decide between the two options.

util-read.R
```diff
@@ -952,6 +953,7 @@ read.commit.interactions = function(data.path = NULL) {
         ## Author data will be merged from commit data in \code{update.commit.interactions}.
         interactions["base.author"] = NA_character_
         interactions["interacting.author"] = NA_character_
+        interactions["artifact.type"] = "CommitInteraction"
```
bockthom, line 956: Please replace "CommitInteraction" by the new constant also here.

README.md
```diff
@@ -630,7 +630,7 @@ Updates to the parameters can be done by calling `NetworkConf$update.variables(.
 - `author.relation`
     * The relation(s) among authors, encoded as edges in an author network
     * **Note**: The  author--artifact relation in bipartite and multi networks is configured by `artifact.relation`!
-    * possible values: [*`"mail"`*, `"cochange"`, `"issue"`]
+    * possible values: [*`"mail"`*, `"cochange"`, `"issue"`, `commit.interaction`]
```
bockthom, line 633: Here are the double quotation marks inside the back-ticks missing.

README.md
```diff
@@ -649,7 +649,7 @@ Updates to the parameters can be done by calling `NetworkConf$update.variables(.
 - `artifact.relation`
     * The relation(s) among artifacts, encoded as edges in an artifact network
     * **Note**: Additionally, this relation configures also the author--artifact relation in bipartite and multi networks!
-    * possible values: [*`"cochange"`*, `"callgraph"`, `"mail"`, `"issue"`]
+    * possible values: [*`"cochange"`*, `"callgraph"`, `"mail"`, `"issue"`, `commit.interaction`]
```
bockthom, line 652: Here are the double quotation marks inside the back-ticks missing.

README.md
```diff
@@ -234,6 +234,11 @@ There are four types of networks that can be built using this library: author ne
      * The vertices in an artifact network denote any kind of artifact, e.g., source-code artifact (such as features or files) or communication artifact (such as mail threads or issues). All artifact-type vertices are uniquely identifiable by their name. There are only unipartite edges among artifacts in this type of network.
      * The relations (i.e., the edges' meaning and source) can be configured using the [`NetworkConf`](#networkconf) attribute `artifact.relation`. The relation also describes which kinds of artifacts are represented as vertices in the network. (For example, if "mail" is selected as `artifact.relation`, only mail-thread vertices are included in the network.)
 
+- Commit networks
+     * The vertices in a commit network denote any commits in the data. All vertices
+     are uniquely identifyable by the hash of the commit. There are only unipartite edges among commits in this type of network.
```
bockthom, line 239: Please remove the line break here. (It is not visible, but just for consistency reasons.)

README.md
```diff
@@ -234,6 +234,11 @@ There are four types of networks that can be built using this library: author ne
...
+     * The relations (i.e., the edges meaning and source) can be configured using the [`networkConf`](#networkconf) attribute `commit.relation`. The relation also describes the type of data used for network construction (`cochange` uses commit data, `commit.interaction` uses commit interaction data).
```
bockthom, line 240: the edges meaning :arrow_right: the edges' meaning

README.md
```diff
@@ -269,6 +275,7 @@ Relations determine which information is used to construct edges among the verti
 - `commit.interaction`
     * For author networks (configured via `author.relation` in the [`NetworkConf`](#networkconf)), authors who contribute to interacting commits are connected with an edge.
     * For artifact networks (configured via `artifact.relation` in the [`NetworkConf`](#networkconf)), artifacts are connected when there is an interaction between two commits that occur in the artifacts.
+    * For commit networks (configured via `commit.relation` in the [`NetworkConf`](#networkconf)), commits are connected when they interact in the commit interaction data.
```
bockthom, line 278: commit interaction data :arrow_right: commit-interaction data

README.md
```diff
@@ -234,6 +234,11 @@ There are four types of networks that can be built using this library: author ne
      * The vertices in an artifact network denote any kind of artifact, e.g., source-code artifact (such as features or files) or communication artifact (such as mail threads or issues). All artifact-type vertices are uniquely identifiable by their name. There are only unipartite edges among artifacts in this type of network.
      * The relations (i.e., the edges' meaning and source) can be configured using the [`NetworkConf`](#networkconf) attribute `artifact.relation`. The relation also describes which kinds of artifacts are represented as vertices in the network. (For example, if "mail" is selected as `artifact.relation`, only mail-thread vertices are included in the network.)
 
+- Commit networks
```
bockthom, line 237: The new network-conf attributes `commit.relation` and `commit.directed` are missing in the section NetworkConf further below in the README. Please add them there.

NEWS.md
```diff
@@ -12,17 +12,21 @@
...
 - Replace deprecated `igraph` functions by their preferred alternatives (PR #264, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037)
 - Deprecate support for R version 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
+- Refactor function `construct.edge.list.from.key.value.list` to be more readable(PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
 
 ### Fixed
 
 - Fix the creation of edgelists for issue-based artifact-networks by correctly iterating over the issue data (PR #264, 321d85043112971c04998249c14a0677a32c9004)
+- Fix networks based upon commit interaction data to also have the attribute `artifact.type`(PR #263, 849123a8b7d898fbb1343745ecffc1f6000c9367)
```
bockthom, line 28: This is not a fix, as the commit-interaction functionality has not yet been released. So, please remove this line and just add the commit hash of this fix to the very first item "Add commit-interaction data..."

NEWS.md
```diff
@@ -12,17 +12,21 @@
...
 
 ### Changed/Improved
 
 - Change the default value for the `issues.from.source` configuration parameter. Instead of reading JIRA and GitHub issues together, which was the previous default, the new default value causes only GitHub issue data to be read. To restore the previous default behavior and read data from both issue sources, this now needs to be manually configured when needed. (PR #264, 5ff83c364f6bfc1e6ff95e9c5f1087e031c48a5d, 8c8080cb9caf115f19d9f145ad6e6c108b131a67, 8bcbc81db521877908d2e5c2989082ed672f2a3b)
 - Replace deprecated `igraph` functions by their preferred alternatives (PR #264, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037)
 - Deprecate support for R version 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
+- Refactor function `construct.edge.list.from.key.value.list` to be more readable(PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
```
bockthom, line 23: Missing space before opening parenthesis

NEWS.md
```diff
@@ -12,17 +12,21 @@
...
 
 ### Fixed
 
 - Fix the creation of edgelists for issue-based artifact-networks by correctly iterating over the issue data (PR #264, 321d85043112971c04998249c14a0677a32c9004)
+- Fix networks based upon commit interaction data to also have the attribute `artifact.type`(PR #263, 849123a8b7d898fbb1343745ecffc1f6000c9367)
+- Fix endless recursion that could occur when commit interaction data was configured and commit data is empty (PR #263, 3fb7437b68950303916b62984fa449732c70353e)
```
bockthom, line 29: Same as above: If I am not mistaken, this is not a fix, as the commit-interaction functionality has not yet been released. So, please remove this line and just add the commit hash of this fix to the very first item "Add commit-interaction data..."

And it would be nice if you would add the current commit in addition (in which you have moved the check for the caller to another location, as the first commit on that was not finally correct). However, this involves that you don't add the fixes of this review to the previous commit, otherwise it will not be possible to add a commit hash for the second fix.

NEWS.md
```diff
@@ -12,17 +12,21 @@
 - Add line-based code coverage reports into CI pipeline. Coverage reports are generated by `coverage.R` (PR #262, 10cac49d005e87c3964cc61711e7f5acef749626, b3b9f4ac7a9911bd00293c68fac88e0f9033bdfb, c815d18dc6266d620a7a145493417b87ac08679e, e8093525fdaf46e54f2f7fcc6358ca7892e795e5, 32d04823e2007c63d2a43ce59bea3057327c19a7)
 - Add the possibility to split data time-based by multiple data sources (PR #261, 1088395f46b84028c8d7c463ca86b5dc38500c26, e1f79fc9e40cd6f41c946be42db364b2101cfe10, 0bb187fec0fd801d7634bf8d5180525770f6ab0b, 371a97ac6ebf3de4fe9360dea79d62e2ed3ef585)
 - Add tests for uncovered functionality in `util-misc.R` and `util-networks.R` (PR #264, ff30f3238b1bf2539280d0d055a5d925c197c271, af80551d0615a49b86e45ff596bd75941ee88f91)
+- Add commit network as a new type of network. It uses commits as vertices and connects them either via cochange or commit interactions. This includes adding new config parameters and a function for adding vertex attributes to a commit network(PR #263, ab73271781e8e9a0715f784936df4b371d64c338, ab73271781e8e9a0715f784936df4b371d64c338, cd9a930fcb54ff465c2a5a7c43cfe82ac15c134d)
```
bockthom, line 15: Please add a space before the opening parenthesis.

And please also add the name of the function `add.vertex.attribute.commit.network` after "and a function ..."

bockthom (changes requested):
I had a look at README, NEWS, and the most recent changes @Leo-Send.
Please see my comments below - there is one line in which we should make use of the new constant in the code; and there are some missing items in the README.
Everything else are just styling issues (missing spaces etc.).

bockthom (approved):

hechtlC (approved):

Issue #260, opened 2024-04-24
Update deprecated function calls from other packages


bockthom:
### Description

Some of the functions that we call from other packages are marked as deprecated.
Especially since `igraph` release 2.0.0, `igraph` has marked many functions as deprecated. In most cases, the functions have just been renamed.  To provide one example, `igraph::get.edgelist(...)` has been renamed to `igraph::as.edgelist(...)`.
As I assume that, at some point, these functions will become defunct, we should think about renaming them.

We can use this issue to collect deprecated function calls here, if necessary.

MaLoefUDS:
I have conducted an analysis of the warning messages I retrieved (R: 4.1.2 on my Ubuntu VM). In total I classified 66 warnings into 7 classes:

1. "can't open file" (**fixable**: mixed)
2. "data length is not a multiple of split variable" (**fixable**: yes!) This warning actually stems from a "bug" in the implementation of the `add.link` <-> `referenced.by` edge list construction. There, we have a dataframe containing all `add.link` events called `add.links` and we want to `mclapply` over it. I used the `split(add.links, ...)` function for that and that function usually works, that is, if you supply a second parameter that is a vector from 1 to the length of the first parameter. However, to get this vector I used `seq_along(add.links)`, which does return a vector from 1 to 17 in every case since we have 17 columns in an issue event. Therefore, we get the warning that the length of the second parameter is not in line with the length of the first parameter. I experimented a bit and was able to this by replacing the second parameter with `seq_len(nrow(add.links))` or `seq.int(nrow(add.links))`, alternatively.
3. "All formats failed to parse. No formats found." (**fixable**: unsure) There is an issue with `project.data$get.data.timestamps(outermost = TRUE)` in `util-split.R:427`. Sometimes this call does return a vector where `start` is `Inf` and `end` is `-Inf`. A consecutive `get.date.from.string` call then throws this issue, since `Inf` and `-Inf` are not in the correct format for conversion.
4. "no non-missing arguments to min; returning Inf", (**fixable**: unsure) This is related to point 3. This warnings is being thrown inside the `get.date.timestamps` call. There reason for this is that the `date.timestamps` are empty
This is because the test these issues come from is a test that operates on an empty network.
5. "All formats failed to parse. No formats found." 2.0. (**fixable**: unsure) This issue comes from splitting by invalid bins time based (the test splits explicitly wants to split by invalid bins) (these bins are not in correct date format)
[...]

PR #264, opened 2024-05-28
Fix igraph deprecation warnings as well as a bug in issue network construction

MaLoefUDS:
### Prerequisites

### Description

* Fix igraph deprecation warnings
* Fix a bug in issue network construction that is detailed in #260.


bockthom:
I've identified the reason for why the pipeline fails for R 3.6 now: Package `testthat` depends on package `evaluate`. A new version of package `evaluate` has been released a few days ago, and the most recent version of package `evaluate` depends on R 4.0. Thus, we would have to hard-code the previous version of package `evaluate` to be able to install `testthat` for R 3.6. As we try to avoid hard-coding package versions in coronet, this is the moment where should think about ceasing the support for R 3.6. 

Independent of that, before we fix that issue, can we somehow enable the generation of the coverage report in our CI pipeline? Maybe it would be a good idea to generate the coverage report if the tests for the version, for which we generate the report, have finished successfully, even if the test pipeline fails for other versions? Just to avoid that the test coverage is not generated just because of specific issues in a single R version. What do you think @MaLoefUDS?

MaLoefUDS:
That should be straightforward. I will look into it tomorrow 👍

MaLoefUDS:
To make the `upload-coverage-report` job run, we need the `build` to finish successfully. The `build` job only counts as being passed if all its steps passed (i.e., for all R versions). I found multiple fixes, that implement your requested behavior. For example we could just add `continue-on-error` onto the run tests and run showcase steps. Then in any case when an older R version fails, we simply continue and for the latest R version, we still check in the generate coverage report step that the run tests step succeeded and nothing about the behavior would change. However, for this fix (and also for all other fixes I discovered) all `build` jobs will always display the pass symbol and never the red cross. There are indications about the internal failure but they are not as immanent. 

Bildschirmfoto 2024-06-22 um 15 54 36

So what is your preference? Currently, the only options I see is that we do not upload coverage if at least 1 R version did not run successfully (which may even make sense?) or that we still upload but then we only have the green check on the build status indicators.

bockthom:
I don't like to have green check marks for failing steps. That'd be weird.
Then we should stay with the current version of the pipeline, I guess.
That is, we really need to fix every failing steps for old R versions immediately before merging PRs – otherwise the code coverage reports would be missing until the failing steps are fixed.
In our case, this means that we need to exclude 3.6 from the build pipeline, remove the corner-case handling for 3.6 in `install.R`, and update the README.md regarding supported R versions...

MaLoefUDS:
Is that something you want me to do inside of this PR? 🤓

bockthom:
We have two options: Either we open a new PR that fixes the CI pipeline, and rebase this PR after the other one is merged.
Or we directly address the CI issue here in this PR, which is completely legitimate since this PR deals with fixing deprecation issues :wink:  What ever you'd prefer; I would be interested in your opinion on this.

Regarding how and when to update the `README.md` we should discuss potential solutions in our next meeting, this could be tricky as we differentiate between "supported" and "recommended" R versions. But fixing the CI pipeline and the deprecation workarounds in the package installation script is something that should be straightforward.

MaLoefUDS:
I agree with you: since we are addressing deprecated features in here, I think it is legitimate to fix the CI here as well 👍

MaLoefUDS:
Regarding the workarounds in the install script: What of the two workarounds are **only** present for R 3.6? As I see it, both of the workarounds may come to play also for newer but strangely configured or incorrectly installed R versions.

Context:

https://github.com/se-sic/coronet/blob/24005e44aaf602789168f69dbcd5cc0a077c0194/install.R#L67-L80

bockthom:
> Regarding the workarounds in the install script: What of the two workarounds are **only** present for R 3.6?

Actually non of the ones in your quoted excerpt. The workaround is only present on the `dev` branch right now:

https://github.com/se-sic/coronet/blob/a87ff243f0729da11d7e7a40a9395c53f64e911d/install.R#L81-L83

The other two version checks in your quoted excerpt are independent of R 3.6. The first one is because we want to support/recommend a minimum version number of `igraph`, and the other one regarding `Matrix` is because `Matrix` is part of the automatically installed packages of `r-recommended` (or even `r-base`) and needs to be re-installed to get a more recent package version that what comes with the R installation. This should be kept.
The only R3.6-specific change is the hard-coded `Matrix` version.

MaLoefUDS:
Oh okay, I see 🤓

.github/workflows/pull_request.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        r-version: ['3.6', '4.0', '4.1', '4.2', '4.3', 'latest']
+        r-version: ['4.0', '4.1', '4.2', '4.3', 'latest']
```
bockthom, line 40: Could you please add '4.4' instead? This means that we will run '4.4' twice until '4.5' will be published. But we don't need to keep track of when '4.5' will be eventually published and automatically run '4.4' and the latest version then.

MaLoefUDS: Of course!

bockthom: And could you please fix your e-mail address up there in the copyright header? "cs" should be replaced by "stud" in your e-mail address, I guess...

bockthom:
Interesting: The upload of the coverage report worked again without changing anything, right?

util-networks.R
```diff
@@ -1764,7 +1764,7 @@ simplify.networks = function(networks, remove.multiple = TRUE, remove.loops = TR
 #'
 #' @return the network without isolates
 delete.isolates = function(network) {
-    network.no.isolates = igraph::delete.vertices(
+    network.no.isolates = igraph::delete_vertices(
```
bockthom, line 1767: Interesting that we don't have a test for `delete.isolates`. We should change that :wink:

util-networks.R
```diff
@@ -1868,7 +1868,7 @@ delete.authors.without.specific.edges = function(network, specific.edge.types =
     ## compute all authors without specific edges as vertex IDs
     vertex.ids.author.no.specific = setdiff(vertex.ids.author, vertex.ids.specific)
     ## remove all authors without specific edges from network
-    network = igraph::delete.vertices(network, vertex.ids.author.no.specific)
+    network = igraph::delete_vertices(network, vertex.ids.author.no.specific)
```
bockthom, line 1871: It looks like that we also don't have a test for `delete.authors.without.specific.edges`?
I'd also be happy to have a simple test for this function.

bockthom (changes requested):
Thanks for this PR @MaLoefUDS. It looks really great, thanks for your efforts on changing all the reading and splitting tests due to the new default value of the issue source.

I spotted two minor issues:
- the new default value is not updated in the `README.md` yet (see details below)
- Codecov points us to some uncovered functionality. We should take advantage of this and add a few tests for some of the uncovered functions revealed by Codecov. But definitively not for all of them. I don't care about everything in `util-bulk.R`, and we don't have a proper testing setup for `util-plot.R` yet. So we can ignore Codecov's complaints regarding these files. But Codecov also revealed a small number of functions in well-tested files such as `util-networks.R` that seem to be untested still, and I would like to add some small tests for them. Please see my comments below for the particular functions I am referring to.

util-networks.R
```diff
@@ -1726,7 +1726,7 @@ simplify.network = function(network, remove.multiple = TRUE, remove.loops = TRUE
 
     ## re-apply all network attributes
     for (att in names(network.attributes)) {
-        network = igraph::set.graph.attribute(network, att, network.attributes[[att]])
+        network = igraph::set_graph_attr(network, att, network.attributes[[att]])
```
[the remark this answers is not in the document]
MaLoefUDS: From my "research", graph attributes can be anything really mapping a key to a value. In coronet, we really use only 2 different graph attributes. 1) `layout` (some layout information relevant for plotting), 2) `range` (range-split information). Interestingly, the `get.sample.network()` method does also add a graph attribute `sample.network` (value: `TRUE`); I guess we just never simplify a sample network in the tests. 

Adding a graph attribute to a network, should usually not have an impact on tests, therefore, I don't really prefer one attribute over the other. We can just add any in my opinion :)

tests/test-misc.R
```diff
@@ -19,6 +19,33 @@
...
+        igraph::vertices("A", "B", "C", "D") +
+        igraph::edges(edges, relation = "mail", date = timestamps)
+
+
+    ## get edgelist augmented with timestamps
+    edgelist = get.edgelist.with.timestamps(network)
+
+    ## check correctness
+    expect_equal(names(edgelist), c("from", "to", "date"))
+    for (i in 1:nrow(edgelist)) {
```
bockthom, line 41: Can we avoid `1:nrow` here, please? Although this is a test and `edgelist` is expected to have, at least, one row, I would like to avoid using `1:nrow`, in general.

bockthom:
> ## [Codecov](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) Report
> 
> Attention: Patch coverage is `72.46377%` with `19 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 79.89%. Comparing base [(`a87ff24`)](https://app.codecov.io/gh/se-sic/coronet/commit/a87ff243f0729da11d7e7a40a9395c53f64e911d?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) to head [(`fb3f547`)](https://app.codecov.io/gh/se-sic/coronet/commit/fb3f5474259d4a88f4ff545691cca9d1ccde90e3?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic).
> 
> > ❗ **Current head [fb3f547](https://github.com/se-sic/coronet/commit/fb3f5474259d4a88f4ff545691cca9d1ccde90e3) differs from pull request most recent head [0146a3b](https://github.com/se-sic/coronet/commit/0146a3b180586bc4457a1aba5542967034fa2f39)**
> > Please [upload](https://docs.codecov.com/docs/codecov-uploader) reports for the commit [0146a3b](https://github.com/se-sic/coronet/commit/0146a3b180586bc4457a1aba5542967034fa2f39) to get more accurate results.
> 
> [Files](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) 	Patch % 	Lines
> [util-plot.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-plot.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1wbG90LlI=) 	0.00% 	[11 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> [util-bulk.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-bulk.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1idWxrLlI=) 	0.00% 	[4 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> [util-networks.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-networks.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1uZXR3b3Jrcy5S) 	93.02% 	[3 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> [util-misc.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-misc.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1taXNjLlI=) 	0.00% 	[1 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> Additional details and impacted files
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic).

@MaLoefUDS Do you have any idea why the Coverage report in this issue is not updated?
On the Codecov website, everything looks fine and everything is up to date. But here in the PR, the corresponding comment has been updated right after your commits today in the morning, but it still shows outdated information. Do you have any idea why this is the case?

tests/test-misc.R
MaLoefUDS: Yeah, thats not a problem 👌

util-networks.R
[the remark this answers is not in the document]
bockthom: Ok, then please choose any graph attribute that sounds reasonable (at least, to some extent).
Do you have already any test in mind in which you would like to add a graph attribute?

MaLoefUDS:
I do not know why the comment does not update. [The docs](https://docs.codecov.com/docs/pull-request-comments) state that the default behavior is to update the comment and otherwise send a new message. We did not update the default and there is no new comment, so we can assume that the codecov update happened and failed for some reason. We can force codecov to send new comments on every change to the PR, maybe that would help, but im unsure. Apart from that I have no idea 🤷‍♂️

util-networks.R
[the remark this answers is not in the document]
MaLoefUDS: There are only 4 tests that use `simplify.network`(s). The first test simplifies a network with multiple relations (`simplify.multiple.relations` = `FALSE`), the second is basic, the third has again a network with multiple relations (`simplify.multiple.relations` being `TRUE` and `FALSE`). The last test tests `get.data.sources.from.relations` with simplification (`simplify.multiple.relations` = `TRUE`). 

The last test is already pretty tailored and not so much towards the simplifying functionality, so I would not pick this one to test. The second test is the simplest. The third is the most complex. From your conditions, this would only leave only the first test.
Personally, I would rather take the most complex test though, just because it is the only test that also uses `simplify.networks`.

bockthom:
Right now the comment from Codecov is up to date again and refers to the most recent commit (see quote below). So I guess something got stuck there (for whatever reason).  So let's assume it works well.

> ## [Codecov](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) Report
> 
> Attention: Patch coverage is `76.81159%` with `16 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 80.22%. Comparing base [(`a87ff24`)](https://app.codecov.io/gh/se-sic/coronet/commit/a87ff243f0729da11d7e7a40a9395c53f64e911d?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) to head [(`07a06cd`)](https://app.codecov.io/gh/se-sic/coronet/commit/07a06cdd0c6c159549ece727c35df87465477dd8?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic).
> 
> [Files](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic) 	Patch % 	Lines
> [util-plot.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-plot.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1wbG90LlI=) 	0.00% 	[11 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> [util-bulk.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-bulk.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1idWxrLlI=) 	0.00% 	[4 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> [util-networks.R](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&filepath=util-networks.R&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic#diff-dXRpbC1uZXR3b3Jrcy5S) 	97.67% 	[1 Missing ⚠️ ](https://app.codecov.io/gh/se-sic/coronet/pull/264?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic)
> Additional details and impacted files
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/se-sic/coronet/pull/264?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=se-sic).

util-networks.R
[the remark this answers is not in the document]
bockthom: > From your conditions, this would only leave only the first test. Personally, I would rather take the most complex test though, just because it is the only test that also uses `simplify.networks`.

Thanks for having a look at these tests. I am fine with both options. So, I'll let you decide.

bockthom:
> We can force codecov to send new comments on every change to the PR

I don't think that @hechtlC would agree with that :wink: We already receive lots of e-mails, we don't need to receive multiple e-mails for one push event... If codecov is able to properly update the first comment (as it seems to be, at the moment), updating the comment is preferred over sending new comments, I guess.

tests/test-misc.R
```diff
@@ -19,6 +19,34 @@
...
+    edgelist = get.edgelist.with.timestamps(network)
+
+    ## check correctness
+    expect_equal(names(edgelist), c("from", "to", "date"))
+    lapply(1:4, function(i) {
+        actual = edgelist[i, ]
+        expect_equal(actual[["from"]], edges[[i]][[1]])
+        expect_equal(actual[["to"]], edges[[i]][[2]])
+        expect_equal(actual[["date"]], timestamps[i])
+    })
```
bockthom, line 46: In this new test implementation, we also need to check whether `nrow(edgelist)` is equal to `4`. Otherwise the test would not fail if there would be more rows in the actual result than expected.

MaLoefUDS: I thought so too but wanted to avoid nrow 😅

bockthom: There is no need to avoid `nrow(...)`, `nrow(...)` is good.
In contrast, `1:nrow(...)` is bad and should be avoided, because this can lead to errors or unexpected behavior if `nrow(...)` is 0.

MaLoefUDS: Oh I see now, yeah I will change that quickly

bockthom (commented):
Everything looks good to me now. So, the only thing that is missing is to add the changes to the `NEWS.md`.

hechtlC (dismissed):
Looks good to me! Thanks @MaLoefUDS .

NEWS.md
```diff
@@ -11,12 +11,17 @@
...
 - Add the possibility to split data time-based by multiple data sources (PR #261, 1088395f46b84028c8d7c463ca86b5dc38500c26, e1f79fc9e40cd6f41c946be42db364b2101cfe10, 0bb187fec0fd801d7634bf8d5180525770f6ab0b, 371a97ac6ebf3de4fe9360dea79d62e2ed3ef585)
+- Add tests for uncovered functionality in `util-misc.R` and `util-networks.R` (PR #264, ff30f3238b1bf2539280d0d055a5d925c197c271, af80551d0615a49b86e45ff596bd75941ee88f91)
 
 ### Changed/Improved
 
+- Deprecate support for R 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
+- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037)
+- Change the default value for the `issues.from.source` configuration parameter from `JIRA` and `github` to `github`. This change is motivated by the fact that, in practice, `github` is the predominant source for issue data (PR #264, 5ff83c364f6bfc1e6ff95e9c5f1087e031c48a5d, 8c8080cb9caf115f19d9f145ad6e6c108b131a67, 8bcbc81db521877908d2e5c2989082ed672f2a3b)
```
bockthom, line 20: First, a general comment: When you use the parameter value as source code, it should be `jira` instead of `JIRA`. If you use it outside of a code environment, JIRA is the spelling variant we use.

And now to the content of this entry: Can we make the actual change a little bit more explicit?
The part "from `JIRA` and `github` to `github`." is a little bit hard to capture. I would suggest to add a "`github` only" at the end. And I am unsure how to stress "`JIRA` and `github`". One option would be to provide the actual vector `c(jira, github)`, but this may be too explicit, and we never have used vectors in the `NEWS.md` before.
Maybe we can rephrase the whole sentence to add a little bit more information and stress the actual difference. What about the following:
"Change the default value for the `issues.from.source` configuration parameter. Instead of reading JIRA and GitHub issues together, which was the previous default, the new default value causes only GitHub issue data to be read. To restore the previous default behavior and read data from both issue sources, this now needs to be manually configured when needed."
Feel free to rephrase and improve parts of my suggestion.

And in the second sentence, `github` should be replaced by GitHub, as you there talk about the data source itself and not about the configuration parameter.

---

And just a comment for @hechtlC: We should think about whether we would like to mark this change as breaking change or not; there are reasons for and against it. But that's nothing for this PR; we will deal with that in the next versioning PR prior to the next release.

MaLoefUDS: No complaints about the description you gave, I will just use that instead :)

NEWS.md
```diff
@@ -11,12 +11,17 @@
...
+- Add tests for uncovered functionality in `util-misc.R` and `util-networks.R` (PR #264, ff30f3238b1bf2539280d0d055a5d925c197c271, af80551d0615a49b86e45ff596bd75941ee88f91)
 
 ### Changed/Improved
 
+- Deprecate support for R 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
+- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037)
+- Change the default value for the `issues.from.source` configuration parameter. Instead of reading JIRA and GitHub issues together, which was the previous default, the new default value causes only GitHub issue data to be read. To restore the previous default behavior and read data from both issue sources, this now needs to be manually configured when needed. (PR #264, 5ff83c364f6bfc1e6ff95e9c5f1087e031c48a5d, 8c8080cb9caf115f19d9f145ad6e6c108b131a67, 8bcbc81db521877908d2e5c2989082ed672f2a3b)
```
bockthom, line 20: Ok, one last comment: Can we adjust the order please? The change of the default value should be the first one, and the deprecation of 3.6 should be the last one.

To stay consistent with previous entries in the `NEWS.md` I suggest the following changes:
- Replace "R 3.6" by "R version 3.6" (at least, this is the variant we have used further below in this document, if I am not mistaken.)
- To stay consistent with previous entries. We should add an additional, separate item thereafter: " Explicitly add R version 4.4 to the CI test pipeline (commit hash without PR number)" - we did this for 4.3 in the last release, so we should also state it for 4.4 here.

Everything else looks fine to me. So this PR is ready to merge after the above mentioned minor issues are addressed.

bockthom (approved):

PR #268, opened 2024-08-27
Miscellaneous improvements regarding network splitting and simplification

MaLoefUDS:
### Prerequisites

- [ ] I have checked whether I need to adjust the showcase file `showcase.R` with respect to my changes.


util-split.R
```diff
@@ -1016,6 +1016,8 @@ split.data.by.time.or.bins = function(project.data, splitting.length, bins, spli
                         findInterval(df[["date"]], bins.date, all.inside = FALSE)
             ## split data according to df.bins
             df.split = split(df, df.bins)
+            ## remove data before the first bin
+            df.split = df.split[names(df.split) != "0"]
```
bockthom, line 1020: Two questions: "0" as a character looks weird – although I know that it is correct. But could we add a comment at the end of the line that explains why we are looking for the "0" as character?

And do we have tests for this edge case? Not sure whether and how we can test this, but maybe by triggering a similar situation as the one in which we have found this bug?

util-networks.R
```diff
@@ -1993,6 +1993,43 @@ delete.isolates = function(network) {
     return(network.no.isolates)
 }
 
+#' Remove duplicate edges from the given network.
```
bockthom, line 1996: Maybe we should add an additional sentence describing what "duplicate edges" mean.
It may also be useful to add a sentence in the readme, when we introduce examples to explain temporal order and directedness, that duplicate edges in a particular case can be removed by calling this function. Otherwise, nobody will use this function because nobody is aware of it.

Let's also check if there is a comment on duplicate edges in one of the edge creation functions. If so, we could refer to this function in the comment. If not, then it is not really necessary.

tests/test-networks.R
```diff
@@ -317,6 +317,47 @@ test_that("Remove isolated authors given a specific edge type", {
 
 })
 
+test_that("Remove duplicate edges", {
+
+    ## configurations
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    net.conf = NetworkConf$new()
+    net.conf$update.value("author.respect.temporal.order", TRUE)
```
bockthom, line 325: For better comprehensibility of this test, it may be useful to also state which value of `author.directed` is used. Because this should make a difference on whether edges are actually removed or not, if I am not mistaken.

bockthom (commented):
Thanks! This PR looks good to me. I just have a few comments regarding documentation and tests.

util-split.R
MaLoefUDS: About the comment, yes. But your second point really is not testable. This issue does not really propagate to anything visible from the outside, it’s just internal. The reason for this is that ranges with names that are not corresponding to bin labels are dropped anyways just a few lines further down.

README.md
```diff
@@ -318,7 +318,7 @@ Based on the above raw data, we get the following author networks with relation
   </tr>
 </table>
 
-When constructing author networks with respecting the temporal order, there is one edge for each answer in a mail thread from the answer's author to the senders of every previous e-mail in this mail thread. Note that this can lead to duplicated edges if an author has sent several previous e-mails to the mail thread (see the duplicated edges `A –(3)– B` in the above example). This also leads to loop edges if an author of an answer has already sent an e-mail to this thread before (see the edge `A –(2)– A`).
+When constructing author networks with respecting the temporal order, there is one edge for each answer in a mail thread from the answer's author to the senders of every previous e-mail in this mail thread. Note that this can lead to duplicated edges if an author has sent several previous e-mails to the mail thread (see the duplicated edges `A –(3)– B` in the above example). These may be disambiguated again using the `remove.duplicate.edges` function. Furthermore, respecting the temporal order also leads to loop edges if an author of an answer has already sent an e-mail to this thread before (see the edge `A –(2)– A`).
```
bockthom, line 321: I don't like the word "disambiguated" here as does not clearly indicate what actually happens.
What about "conflated" ?

Everything else looks fine to me.

bockthom (changes requested):

bockthom:
Just one final question before I can approve this PR: The splitting fix regarding the elements before the first bin does not have any user-visible effects? If this really is the case, I am fine with not mentioning it. If it can have user-visible effects, we should mention it in the NEWS.md.

MaLoefUDS:
Yup, it has no user-visible effect. That is why I found it hard to justify mentioning it in the NEWS and decided against it 😄

bockthom (approved):

Issue #269, opened 2024-09-19
Wrong data type of NA timestamps depending on order of data sources


bockthom:
### Description

`project.data$get.data.cut.to.same.date(data.sources = c("issues", "mails", "commits"))` sometimes fails if some of the data sources are empty, ending up in the following error message:

```R
ERROR::The bins parameter, if present, needs to be a vector whose elements represent dates
```
This error occurs if the issue data is the empty data source, but not if the mail data is the empty data source, which is an unexpected behavior.

After some debugging, I noticed that the error is caused by wrong data types when converting strings to POSIXct dates: Usually, if the date is not a POSIXct object, it is converted to POSIXct when splitting the data:
https://github.com/se-sic/coronet/blob/24005e44aaf602789168f69dbcd5cc0a077c0194/util-split.R#L69-L70
However, this fails if the bins (i.e., the timestamps used for cutting) are not a string but a numeric (such as UNIX timestamps). Therefore, we need to look at where the timestamps for cutting come from: They are extracted in the function `extract.timestamps`: https://github.com/se-sic/coronet/blob/24005e44aaf602789168f69dbcd5cc0a077c0194/util-data.R#L724-L758

If no data is available for the specified data source, the corresponding timestamps are set to `NA`. For whatever reason, this converts the data frame used for cutting into a numeric data frame of UNIX timestamps if the first row contains `NA` values. Otherwise, if there are already POSIXct objects in there, the inserted `NA` values are interpreted to be POSIXct objects.

### Suggested Fix

Instead of setting the timestamps to `NA` in case of missing data sources, set the timestamps to `as.POSIXct(NA)`. This way, we ensure that the data type is POSIXct even if the `NA` values are the very first ones that are inserted into the timestamps data frame.

### Versions

This affects, at least, coronet version 4.4 in combination with R version 4.4.1. Also earlier coronet versions or R versions may be affected by this bug.

bockthom:
I have already implemented a fix for that, see the linked patch:
[0001-Ensure-correct-data-type-of-NA-timestamps.patch](https://github.com/user-attachments/files/17062942/0001-Ensure-correct-data-type-of-NA-timestamps.patch.txt)


However, we also should properly test that, to make sure that the patch properly works in various situations and does not break anything.

MaLoefUDS:
[...]

PR #270, opened 2024-10-16
Fix a bug in 'extract.timestamps' when dealing with an empty data source

maxloeffler:
### Prerequisites

### Description

Include a fix by @bockthom for issue #269 and a new unit test for the previously faulty behavior.

### Changelog

Fixed:
* Fix a bug in `extract.timestamps` that occurs when the first `data.source` is empty and leads to a return value of type `numeric` which should be `POSIXct`.

### DIFF
--- a/tests/test-data-cut.R
+++ b/tests/test-data-cut.R
@@ -18,6 +18,7 @@
 ## Copyright 2018 by Thomas Bock <bockthom@fim.uni-passau.de>
 ## Copyright 2020 by Thomas Bock <bockthom@cs.uni-saarland.de>
 ## Copyright 2018 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
+## Copyright 2024 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
 ## All Rights Reserved.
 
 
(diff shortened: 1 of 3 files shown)


tests/test-data-cut.R
```diff
@@ -82,3 +83,58 @@ test_that("Cut commit and mail data to same date range.", {
...
+                                        committer.email = c("bjoern@example.org", "bjoern@example.org"),
+                                        hash = c("72c8dd25d3dd6d18f46e2b26a5f5b1e2e8dc28d0", "5a5ec9675e98187e1e92561e1888aa6f04faa338"),
+                                        changed.files = as.integer(c(1, 1)),
+                                        added.lines = as.integer(c(1, 1)),
+                                        deleted.lines = as.integer(c(1, 0)),
+                                        diff.size = as.integer(c(2, 1)),
+                                        file = c("test.c", "test.c"),
+                                        artifact = c("A", "A"),
+                                        artifact.type = c("Feature", "Feature"),
+                                        artifact.diff.size = as.integer(c(1, 1)))
```
bockthom, line 115: Could you please align the indentation with the beginning of the first parameter?

tests/test-data-cut.R
```diff
@@ -82,3 +83,58 @@ test_that("Cut commit and mail data to same date range.", {
...
+    ## configurations
+
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.conf$update.value("issues.locked", TRUE)
+    data.sources = c("issues", "mails", "commits")
+
+    ## construct objects
+
+    x.data = ProjectData$new(proj.conf)
+    x.data$set.issues(NULL)
```
bockthom, line 98: Could you please add a comment that states that, in this test, actually the first processed data source is empty, to highlight that the first processed empty determines the data type of the data frame? This is an important aspect that has to be considered, for example, when this test is to be adjusted in the future.

tests/test-data-cut.R
```diff
@@ -82,3 +83,58 @@ test_that("Cut commit and mail data to same date range.", {
...
+    rownames(commit.data) = 1:nrow(commit.data)
+
+    mail.data = x.data$get.data.cut.to.same.date(data.sources = data.sources)$get.mails()
+    rownames(mail.data) = 1:nrow(mail.data)
+
+    issue.data = x.data$get.data.cut.to.same.date(data.sources = data.sources)$get.issues()
+
+    expect_identical(commit.data, commit.data.expected, info = "Cut Raw commit data.")
+    expect_identical(mail.data, mail.data.expected, info = "Cut mail data.")
+    expect_identical(issue.data, issue.data.expected, info = "Cut issue data.")
```
bockthom, line 138: Maybe highlight that this empty: "Cut issue data (empty)."

util-data.R
```diff
@@ -113,7 +113,7 @@ DATASOURCE.TO.ARTIFACT.COLUMN = list(
 
 
 ## the maximum time difference between subsequent mails of a patchstack
-PATCHSTACK.MAIL.DECAY.THRESHOLD = "30 seconds"
+PATCHSTACK.MAIL.DECAY.THRESHOLD = lubridate::as.duration("30 seconds")
```
bockthom, line 116: Please update the copyright header in this file.

bockthom (commented):
Thanks for this PR @MaLoefUDS.  I had a quick look at it, looks good to me. I just spotted a few minor issues, such as enhancing comments with additional information.

And it also seems that you need to rebase onto the current `dev` branch.

tests/test-data-cut.R
MaLoefUDS: Im struggeling a bit with formulating this comment. My problem is that now, after your fix, the first data source does not influence the data type of the dataframe anymore, as this was never intended. Currently, I'm stuck with something like

```
    ## ensure that the first data source contains empty data, as it used to determine
    ## the data type of the data frame that stores the project timestamps (which is unintended)
```

bockthom: That sounds already good to me. I just would like to highlight even more that this is important for the test. 
And of course it is intended that the first entry determines the data type of the data frame, this is what is going on in R. This is why it is important that the first entry has the correct data type.

So, what about the following:

```R
## in order to test whether the data types of timestamps originating from empty data are correct, 
## ensure that the first data source contains empty data, as it is used to determine
## the data type of the data frame that stores the project timestamps
```

MaLoefUDS: > And of course it is intended that the first entry determines the data type of the data frame, this is what is going on in R.

>  `as it is used to determine the data type of the data frame that stores the project timestamps`

I have to disagree with you on this one. While on a language level this is the intended behavior, it is not for the output of the `extract.timestamps` method and the latter propagation of this value in other coronet components. This is especially highlighted by the fact that after your fix the data type of the project timestamps data frame is totally uniform.

I find it challenging to find a fitting comment that states that the position of the empty data source mattered before your fix without explicitly mentioning your fix.

bockthom: Ok, your concerns are valid.

What about the following wording:

```R
... as R usually uses the first entry of a data frame to determine the data type of the data frame but this should be of no effect since the data type of the project timestamp should be properly set anyway.
```
What do you think about it?

I guess we need to split this up into multiple sentences, as the first part of this sentence is already quite long.

MaLoefUDS: Sounds good. Let me see what I can come up with

MaLoefUDS:
*The last force push was just because of rebasing

bockthom:
> *The last force push was just because of rebasing

That's what I have thought anyway :smile:

bockthom (dismissed):

bockthom (approved):

PR #274, opened 2024-12-05
Change default representation of edge attributes from vectors to lists & other miscellaneous adjustments

maxloeffler:
### Prerequisites

### Description

* Replace deprecated `igraph` functions by their more recent counterparts.
* Handle reading empty commit interaction data explicitly, to avoid crashing in that case.
* To make splitting networks by ranges more consistent with splitting time-based, also add range information to the return value.
* Convert certain edge attributes of networks to list type. This change is motivated by a more stricter type enforcement in `igraph` version 2.0 and up. When merging networks through `igraph::disjoint_union`, edge attributes that exist in both networks are now assumed to have identical types. But, when we simplify a network, its edge attributes convert to list type as they now need to store data that a vector in R cannot store anymore, i.e., one edge attribute now has to hold several *lists* of data of all the edges that got merged into one. Therefore, we need to convert edge attributes to list by default to be able to merge simplified and unsimplified networks.

#### Added

* Introduce `convert.edge.attributes.to.list` function.

#### Fixed

* Fix reading empty commit interaction data.

#### Changed / Improved

* Replace deprecated `igraph` functions.
* Add range information to the return value in `split.network.time.based.by.ranges`, to be consistent with `split.data.time.based`.
* Convert edge attributes to a list format to ensure compatibility with `igraph`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 ## unversioned
 
+### Announcement
+
+- Coronet is not compatible anymore with igraph versions below 2.1.0. This is due to the simultaneous deprecation of `subgraph.edges` and the introduction of the replacement for it, `subgraph_from_edges`, in igraph version 2.1.0.
+
 ### Added
 
 - Add commit-interaction data and add functions `read.commit.interactions` for reading, as well as `get.commit.interactions`, `set.commit.interactions` and utility functions for working with commit-interaction data (PR #252, d82857fbebd1111bb16588a4223bb24a8dcd07de, b4fd2a29c9b5fd561b1106c6febb54a32b0085ab, fd0aa05f824b93545ae8e05833b95b3bd9809286, bca35760eb0aac86c04923f2d534b2d8cece204e, PR #263, 849123a8b7d898fbb1343745ecffc1f6000c9367, 3fb7437b68950303916b62984fa449732c70353e, 170bc66eb779d7cf2ab504db7c3f4ec483103838) as well as tests for these features (PR #252, eeba7e29932bc973513c963fb9e716e9230d570f, 8bb39f4df39b49dfaff8f19feb6db5e5fbd81fac, 54b6f655248720436af116fe72521f9cb0348429, 7a5497aaf9114017d1b3b9b68b6cccd7ca8ac114, 7b8585f87675795822c07230192d6454de31dcc7, ef725407bf8818c8fff96ea6f343338b7162cbe0,)
@@ -18,8 +22,9 @@
 
 ### Changed/Improved
 
+- **Breaking Change**: Change the default representation of edge attributes from vectors to lists. This change is necessary for the interplay of coronet networks with certain igraph functionality since igraph version 2.1. (PR #274, 1c35d1fa2548deb297dbfa5e2b07fce31962c5b7, eda30b838369ec46376812298a3ea8159eec5789, 44c7b72e3234cb332bb2713fb408c124e67255d9, 7303eabef6a78198575fe5bdfc02813fde3d3974)
 - Change the default value for the `issues.from.source` configuration parameter. Instead of reading JIRA and GitHub issues together, which was the previous default, the new default value causes only GitHub issue data to be read. To restore the previous default behavior and read data from both issue sources, this now needs to be manually configured when needed. (PR #264, 5ff83c364f6bfc1e6ff95e9c5f1087e031c48a5d, 8c8080cb9caf115f19d9f145ad6e6c108b131a67, 8bcbc81db521877908d2e5c2989082ed672f2a3b)
-- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5)
+- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, PR #274, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5, e3617b8c6b21fb4242c1d392124813501069ca84, 4b0d5221dd56bb3c9ddf196f67719d4f503d9b61)
 - Deprecate support for R version 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
 - Refactor function `construct.edge.list.from.key.value.list` to be more readable (PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
@@ -28,6 +33,9 @@
 
 - Fix the creation of edgelists for issue-based artifact-networks by correctly iterating over the issue data (PR #264, 321d85043112971c04998249c14a0677a32c9004)
 - Fix a bug in `extract.timestamps` that occurs when the first `data.source` contains empty data and that leads to a return value of type numeric which should be POSIXct (PR #270, 10696e4cf4ae92371917ed8ccaec2b0183da145c, 646c01a42ad8decfbc9040030e790e51cb65cffd)
+- Fix `read.commit.interactions` by explicitly considering non-existent commit interaction data (PR #274, f591528a0f1f11b1a4390949ab770f3f74a766f9)
+- Add range information to network-splits when splitting a network using `split.network.time.based.by.ranges`. This effect also propagates into `split.networks.time.based` (PR #274, 87911ade231c44b93be194a1d6734f7de043a4af)
+- Adjust `metrics.scale.freeness` and `metrics.is.scale.free` functions be compatible with both older and newer igraph versions (PR #274, 4b0d5221dd56bb3c9ddf196f67719d4f503d9b61)
 
 ## 4.4
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 
 ### Announcement
 
-- Coronet is not compatible anymore with igraph versions below 2.1.0. This is due to the simultaneous deprecation of `subgraph.edges` and the introduction of the replacement for it, `subgraph_from_edges`, in igraph version 2.1.0.
+- Coronet is not compatible with `igraph` versions below 2.1.0 anymore. This is due to the simultaneous deprecation of `subgraph.edges` and the introduction of the replacement for it, `subgraph_from_edges`, in igraph version 2.1.0.
 
 ### Added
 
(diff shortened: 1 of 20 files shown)


util-split.R
```diff
@@ -834,6 +834,9 @@ split.network.time.based.by.ranges = function(network, ranges, remove.isolates =
         }
     )
 
+    # add range information
+    names(nets.split) = sapply(ranges.bounds, paste, collapse = "-")
```
bockthom, line 838: A quick question here: Could we just use `ranges` instead of  reconstructing ranges from the range bounds that we have previously extracted from `ranges` ? Or am I missing something here?

maxloeffler: That is not so easy to achieve as it may appear. The reason for this is that `ranges` does not always have the same format. Ranges constructed from `construct.ranges` are a named list in which the name of an entry is the range as `<time-A>-<time-B>` and the value associated with that name is then a vector as `chr[1:2] <time-A> <time-B>`. This means that to derive the valid range information from these ranges one would have to do `names(ranges)`. Other times, especially when called from `split.network.time.based`, `ranges` is just a vector in which all entries are of format `<time-A>-<time-B>` in which case you derive range information by just taking `ranges` as is.

After piping all of these through `get.range.bounds` earlier in the `split.network.time.based.by.ranges`, they follow the same format which allows for an easy reconstruction, so I thought it'll be an okayish solution to just do that. However, if you think otherwise, one could also check if `ranges` has names through `!(is.null(names(ranges)))` and depending on that select either the names or just the plain list.

bockthom: I see. Thanks for the clarification. In general, I would like to get rid of unnecessary `sapply` calls.
I have seen that you have edited the respective line already, but I have two questions regarding your new implementation:

```R
names(nets.split) = if (is.null(names(nets.split))) ranges else names(nets.split)
```
(1) Doesn't this mean that your arguments from the previous comment is invalidated by your most recent change? Because now you assume that `ranges` always is in the correct format...
(2) Is it intended to replace the names by themselves if they are present? Why not just setting names if no names are present:
 ```R
if (is.null(names(nets.split)) { 
    names(nets.split) = ranges
} 
```
Anyway, we need to make sure that there are always the correct ranges added. So, could you please double-check if `ranges` always has the correct form in this case, and also that when names are present they also are in the correct form?

File: util-networks.R [hunk 2]
```diff
@@ -1285,15 +1289,15 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
-            }
+            # u.actual.edge.attribute.date = do.call(base::c, igraph::edge_attr(u, "date"))
+            # if (!is.null(u.actual.edge.attribute.date)) {
+            #     if (is.list(u.actual.edge.attribute.date)) {
+            #         u.expected.edge.attribute.date = lapply(u.actual.edge.attribute.date, get.date.from.unix.timestamp)
+            #     } else {
+            #         u.expected.edge.attribute.date = get.date.from.unix.timestamp(u.actual.edge.attribute.date)
+            #     }
+            #     u = igraph::set_edge_attr(u, "date", value = as.list(u.actual.edge.attribute.date))
+            # }
```
bockthom, line 1308: Two questions regarding this change:
(1) Why did you do changes to the commented code? (`do.call(base::c` as well as `as.list(u.actual.edge.attribute.date)` are new). Is this necessary to make the commented code run with recent igraph versions, or did you just try something without reverting it?

(2) The comment above is outdated if we keep the code commented. "The following code block should be redundant as soon as igraph has fixed their bug" :right_arrow: "The following commented code block is redundant as of igraph version 2.1.1, in which igraph has fixed their bug." (or something similar)

maxloeffler: (1) The changes come from me experimenting with the fix before deciding that it must really be commented out all together to work with the new edge attributes. However, I did not yet delete it such that we can discuss here if we should try to make this whole changed edge attribute thing work with older `igraph` versions or if we can just drop support.

(2) That the comment is still there is just a symptom of (1).

util-split.R
maxloeffler: Both conditions are covered by tests. That was the initial reason for why I needed to do a case distinction here.

bockthom: But then I am waiting for an explanation why the tests did not fail with the "simpler" version proposed above that you consider to be faulty.

util-networks.R
bockthom: It depends on (a) whether you already have a version of this code that already works with older `igraph` versions and on (b) how much time it would take to make it work if it is not working yet.

If it is already working and only two lines need to be changed, we can discuss whether we support older `igraph` versions. If it would take a couple of hours to experiment with this code until it works with older `igraph` versions, I would suggest to cease the support of older `igraph` versions.

util-networks.R
```diff
@@ -1792,6 +1797,13 @@ add.edges.for.bipartite.relation = function(net, bipartite.relations, network.co
         extra.edge.attributes["type"] = TYPE.EDGES.INTER # add egde type
         extra.edge.attributes["relation"] = relation # add relation type
 
+        ## convert edge attributes to list similarely to 'convert.edge.attributes.to.list'
```
bockthom, line 1817: Could you please add a short reason why `convert.edge.attributes.to.list` cannot be used here? One line at max.
(and also fix the typo in "similarly" please)

util-split.R
```diff
@@ -524,7 +524,7 @@ split.network.time.based = function(network, time.period = "3 months", bins = NU
                                     number.windows = NULL,  sliding.window = FALSE,
                                     remove.isolates = TRUE) {
     ## extract date attributes from edges
-    dates = get.date.from.unix.timestamp(igraph::edge_attr(network, "date"))
+    dates = do.call(base::c, igraph::edge_attr(network, "date"))
```
bockthom, line 527: I am not sure if this change is correct. We still need to make sure that the date is in the correct format and not stored as a unix timestamp. I assume that the `do.call(base::c,` just makes the edge-attribute extraction work on lists, but it does not make sure that the timestamps are stored in the correct format.

(This also holds for the similar change in line 712 below).

tests/test-networks-equal-constructions.R
```diff
@@ -42,12 +42,12 @@ compare.edge.and.vertex.lists = function(split.networks.one, split.networks.two)
     for (i in seq_along(split.networks.one)) {
         edges.one = igraph::as_data_frame(split.networks.one[[i]], what = "edges")
         ordering = order(edges.one[["from"]], edges.one[["to"]],
-                         edges.one[["date"]])
+                         as.vector(edges.one[["date"]], "numeric"))
```
bockthom, line 45: Could you please explain this change to me? I neither understand why the `is.vector` is needed here (how does this work with lists?), nor why the date should be `numeric` here? Maybe you can provide a very short example here (just here in the PR discussion)

util-networks.R
bockthom, line 1: The commit that replaces the NULLs with NAs lacks a signed-off tag :wink:

tests/test-networks.R
```diff
@@ -992,6 +992,54 @@ test_that("Addition of edge attributes with data", {
 
 })
 
+
+patrick::with_parameters_test_that("Convert edge attributest to list", {
```
bockthom, line 996: Typo: Even if this is a "test", attributes are not called "attributest" :wink:

tests/test-networks.R
```diff
@@ -992,6 +992,54 @@ test_that("Addition of edge attributes with data", {
...
+        network = igraph::set_edge_attr(network, edge.attributes[i], value = attribute.defaults[[i]])
+    }
+
+    ## convert specified edge attributes to list
+    if (is.null(remain.as.is)) {
+        network.listified = convert.edge.attributes.to.list(network)
+
+        ## set 'remain.as.is' to default value in 'convert.edge.attributes.to.list'
+        ## i.e., intersection of edge attributes and names(EDGE.ATTR.HANDLING)
+        remain.as.is = names(EDGE.ATTR.HANDLING)
```
bockthom, line 1019: I also don't get this one. `Convert.edge.attributes.to.list` was already called beforehand.
And why is this an intersection? The actual edge attributes are not check for here.

bockthom (commented):
While looking over the rest of this PR, I only found a couple of minor issues, please find the individual comments below.

Regarding the uncovered lines reported by codecov: I suggest to ignore them. 2 out of 4 missing lines are part of an untested module anyway. The other two lines refer to very special cases that have been previously untested too.

util-split.R
maxloeffler: I looked it up and the answer is quite more complex than I thought. 

So to refresh, `ranges` comes in two formats:

```R
## Format 1:
 chr [1:6] "<time-A>-<time-B>" "<time-C>-<time-D>" ...

## Format 2:
List of 2
 $ <time-A>-<time-B>: chr [1:2] "<time-A>" "<time-B>"
 $ <time-C>-<time-D>: chr [1:2] "<time-C>" "<time-D>"
```

In case 1: `ranges.bounds` is a `list(POSIXct)` with the bounds for each range.
In case 2: `ranges.bounds` is identical to `ranges`.

Therefore, in the loop over the range bounds, the `bounds` are sometimes of type `chr` and sometimes of type `POSIXct`. The `split.network.time.based` method tries to assign names to the splits already but only succeeds if the ranges are `chr` (in case 2).

So in case 1, the `split.network.time.based` method will not set names to the network splits, however in case 2 it does. The faulty version, therefore, worked since when the splits had not been assigned names yet (case 1), that means that `ranges` can be used right away as the names for the splits, while when the splits were already assigned names (case 2), that means that `ranges` has names and `names(ranges)` can be used again (although redundantly) to assign names.

maxloeffler: That means that the names assignment can be simplified to 
```R
if (is.null(names(nets.split))) {
    names(nets.split) = ranges
}
```

bockthom: I see. So, which version of the fix would you suggest to use eventually?

bockthom: Agreed.

tests/test-networks.R
maxloeffler: Here, I set the `remain.as.is` value, so that I can later use it explicitly in the validation of edge attributes. 
The comment here is indeed outdated, previously I intersected `names(EDGE.ATTR.HANDLING)` and the attributes I use in this test, i.e., `edge.attributes` because it does not really matter whether I test edge attributes that I have not used at all or I don't. In the end I settled on not intersecting as it would not really portrait the default value of `convert.edge.attributes.to.list`, but forgot the remove the comment.

bockthom: Ok, then please adjust the comment, and I am fine with this.

tests/test-networks-equal-constructions.R
maxloeffler: Why `as.vector`: Well, lists cannot be used in the `order` function, only vectors. Previously, we always had vectors here but now we have lists.

Why `"numeric"`: `as.vector` takes a second argument (`mode`). If we do not specify it, it is `"any"` and when using `"any"` it does not perform any conversion here, as `is.vector(list(...))` holds. If we specify `"numeric"` though, it converts the list to a vector of numerics. As we only use this vector for ordering the entries, it does not impact anything else.

bockthom: Thanks for the explanation, sounds good.

util-split.R
maxloeffler: You're right. Previously, the conversion was necessary to pass some tests I believe, now it is not anymore, so I thought it was only for that purpose.

util-networks.R
maxloeffler: I don't have a working version for older `igraph` versions, but I will try to test a bit and see if it's feasible.

maxloeffler: After installing any `igraph` older version, I get errors because the `subgraph_from_edges` method (which I used to replace the deprecated `subgraph.edges` method) is not exported in these versions. Looking at the `rigraph` repo I found that the method has only been [exported recently](https://github.com/igraph/rigraph/commit/543cd9821d83b0a28fc0152eac8fb95eb54ee06d). I don't think would be good to still support older versions under this circumstance anyways.

util-networks.R
```diff
@@ -1806,6 +1814,15 @@ add.edges.for.bipartite.relation = function(net, bipartite.relations, network.co
 
         ## add the vertex sequences as edges to the network
         net = igraph::add_edges(net, unlist(vertex.sequence.for.edges), attr = extra.edge.attributes)
```
bockthom, line 1825: @maxloeffler Could you please report this `NULL` vs. `NA` problem to "rigraph" by replying to the issue that I have opened there a few weeks ago? Just a few sentences in which you relate to part (1), saying that the described problem not only applies to `disjoint_union`, but also to `add_edges`, should be enough, I guess.

(I don't mention the exact issue number here, as I don't want this entire PR be linked to the igraph issue)

util-networks.R
bockthom: Indeed, `subgraph_from_edges` has been introduced recently. So, I guess, we should not support older igraph versions any more.
As we want to have a major release of coronet very soon, I guess it is ok to cease the support of older igraph versions. What we could think of is whether we want to provide a bug-fix branch for the previous release where we cherry-pick certain bug fixes to support older igraph versions - but I am not sure whether this is necessary and I am also not sure what's @hechtlC's opinion on that, so let's discuss this tomorrow in our meeting.

bockthom (changes requested):
I've a few comments regarding the news. Please find them below.

NEWS.md
```diff
@@ -15,19 +15,22 @@
...
-- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5)
+- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, PR #274 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5, e3617b8c6b21fb4242c1d392124813501069ca84, 4b0d5221dd56bb3c9ddf196f67719d4f503d9b61)
 - Deprecate support for R version 3.6 (PR #264, c8e6f45111e487fadbe7f0a13c7595eb23f3af6e, fb3f5474259d4a88f4ff545691cca9d1ccde90e3)
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
 - Refactor function `construct.edge.list.from.key.value.list` to be more readable (PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
+- Change the default representation of edge attributes from vectors to lists. This change is necessary for the interplay of coronet networks with certain igraph functionality since igraph version 2.1. (PR #274, 1c35d1fa2548deb297dbfa5e2b07fce31962c5b7, eda30b838369ec46376812298a3ea8159eec5789, 44c7b72e3234cb332bb2713fb408c124e67255d9, 7303eabef6a78198575fe5bdfc02813fde3d3974)
```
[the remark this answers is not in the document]
maxloeffler: How I see it is that coronet in itself is consistent. So every action you can perform with it should work just as it did before. While networks created with older coronet versions may not necessarily be compatible with networks created with this version, I don't know how this should ever happen, if not through serializing an old network and now deserializing it. Apart from that, there *should* be no incompatibilities.

NEWS.md
```diff
@@ -15,19 +15,22 @@
...
+- Add range information to network-splits when splitting a network using `split.network.time.based.by.ranges` (PR #274, 87911ade231c44b93be194a1d6734f7de043a4af)
 
 ### Changed/Improved
 
 - Change the default value for the `issues.from.source` configuration parameter. Instead of reading JIRA and GitHub issues together, which was the previous default, the new default value causes only GitHub issue data to be read. To restore the previous default behavior and read data from both issue sources, this now needs to be manually configured when needed. (PR #264, 5ff83c364f6bfc1e6ff95e9c5f1087e031c48a5d, 8c8080cb9caf115f19d9f145ad6e6c108b131a67, 8bcbc81db521877908d2e5c2989082ed672f2a3b)
-- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5)
+- Replace deprecated `igraph` functions by their preferred alternatives (PR #264, PR #268, PR #274 0df9d5bf6bafbb5d440f4c47db4ec901cf11f037, 7ac840d287a862eff61b1a84e194a4cba399f9e5, e3617b8c6b21fb4242c1d392124813501069ca84, 4b0d5221dd56bb3c9ddf196f67719d4f503d9b61)
```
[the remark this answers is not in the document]
maxloeffler: Is that still the case if we deprecate support for `igraph` versions below 2.1? I should probably add that somewhere in the `NEWS.md` I just haven't found a good place for it.

maxloeffler:
Between the last and the second to last push I forgot to update the copyright headers of `util-networks.R` 😅.

NEWS.md
```diff
@@ -18,8 +22,9 @@
 
 ### Changed/Improved
 
+- **Breaking Change**: Change the default representation of edge attributes from vectors to lists. This change is necessary for the interplay of coronet networks with certain igraph functionality since igraph version 2.1. (PR #274, 1c35d1fa2548deb297dbfa5e2b07fce31962c5b7, eda30b838369ec46376812298a3ea8159eec5789, 44c7b72e3234cb332bb2713fb408c124e67255d9, 7303eabef6a78198575fe5bdfc02813fde3d3974)
```
bockthom: Same as above, please use back ticks for `igraph` also here. And my be we should also use version 2.1.0 here, to be consistent with the announcement.

NEWS.md
```diff
@@ -28,6 +33,9 @@
 
 - Fix the creation of edgelists for issue-based artifact-networks by correctly iterating over the issue data (PR #264, 321d85043112971c04998249c14a0677a32c9004)
 - Fix a bug in `extract.timestamps` that occurs when the first `data.source` contains empty data and that leads to a return value of type numeric which should be POSIXct (PR #270, 10696e4cf4ae92371917ed8ccaec2b0183da145c, 646c01a42ad8decfbc9040030e790e51cb65cffd)
+- Fix `read.commit.interactions` by explicitly considering non-existent commit interaction data (PR #274, f591528a0f1f11b1a4390949ab770f3f74a766f9)
+- Add range information to network-splits when splitting a network using `split.network.time.based.by.ranges`. This effect also propagates into `split.networks.time.based` (PR #274, 87911ade231c44b93be194a1d6734f7de043a4af)
+- Adjust `metrics.scale.freeness` and `metrics.is.scale.free` functions be compatible with both older and newer igraph versions (PR #274, 4b0d5221dd56bb3c9ddf196f67719d4f503d9b61)
```
bockthom, line 39: shouldn't it be "to be compatible with" instead of just "be compatible with" ?

NEWS.md
```diff
@@ -28,6 +33,9 @@
 
 - Fix the creation of edgelists for issue-based artifact-networks by correctly iterating over the issue data (PR #264, 321d85043112971c04998249c14a0677a32c9004)
 - Fix a bug in `extract.timestamps` that occurs when the first `data.source` contains empty data and that leads to a return value of type numeric which should be POSIXct (PR #270, 10696e4cf4ae92371917ed8ccaec2b0183da145c, 646c01a42ad8decfbc9040030e790e51cb65cffd)
+- Fix `read.commit.interactions` by explicitly considering non-existent commit interaction data (PR #274, f591528a0f1f11b1a4390949ab770f3f74a766f9)
```
bockthom, line 37: Here it seems to me that you have only fixed the first part (replaced empty be non-existent), but this only makes sense if you also replace the following words as indicated in my previous review: commit interaction data :arrow_right: commit interactions.

(the commit interaction data are present, but no commit interactions are present).

bockthom (commented):
I just found a few minor inconsistencies in the README, please find them below.
If also @hechtlC is fine with the new announcement, this PR is ready to be merged then.

bockthom (dismissed):

bockthom (approved):

PR #275, opened 2024-12-10
Add functions to compute a developer's last activity

bockthom:
### Prerequisites

### Description

While we already have functionality to compute the first activity per developer, there was no functionality yet to compute the last activity per developer. With this PR, functionality to compute the last activity per developer is added.

#### Added
- Add function to compute last activity per person and activity type
- Add function to add vertex attribute "last.activity"

### DIFF
--- a/util-networks-covariates.R
+++ b/util-networks-covariates.R
@@ -1582,43 +1582,78 @@ add.vertex.attribute.issue.is.pull.request = function(list.of.networks, project.
 ## Helper ------------------------------------------------------------------
 
 #' Helper function for first activity: computes first activity information per person and activity type.
+#’
+#' @param activity.types The activity types to compute information for [default: c("mails", "commits", "issues")]
+#' @param range.data The data to base the computation on
+#' @param default.value The default value to add if no information is available per author and activity type
+#'                      [default: NA]
 #'
-#' @param activity.types The activity types to compute information for. [default: c("mails", "commits", "issues")]
-#' @param range.data The data to base the computation on.
-#' @param default.value The default value to add if no information is available per author and activity type.
-#' [default: NA]
-#'
-#' @return A list containing per author a list of first activity values named with the corresponding activity type.
+#' @return A list containing per author a list of first activity dates named with the corresponding activity type.
 #'         Empty list if there are no activities in \code{range.data} at all or none corresponding to the configured
 #'         types in \code{activity.types}
-get.first.activity.data = function(range.data, activity.types = c("commits", "mails", "issues"),
-                                   default.value = NA) {
+#'
+#' @seealso get.aggregated.activity.data
+get.first.activity.data = function(range.data, activity.types = c("commits", "mails", "issues"), default.value = NA) {
+    return(get.aggregated.activity.data(range.data, activity.types, default.value, aggregation.function = min))
+}
+
+#' Helper function for last activity: computes last activity information per person and activity type.
+#
+#' @param activity.types The activity types to compute information for [default: c("mails", "commits", "issues")]
+#' @param range.data The data to base the computation on
+#' @param default.value The default value to add if no information is available per author and activity type
+#'                      [default: NA]
+#'
+#' @return A list containing per author a list of last activity dates named with the corresponding activity type.
+#'         Empty list if there are no activities in \code{range.data} at all or none corresponding to the configured
+#'         types in \code{activity.types}
+#'
+#' @seealso get.aggregated.activity.data
+get.last.activity.data = function(range.data, activity.types = c("commits", "mails", "issues"), default.value = NA) {
+    return(get.aggregated.activity.data(range.data, activity.types, default.value, aggregation.function = max))
+}
+
+#' Helper function to aggregate activity information (e.g., to compute first or last activity information per person
+#' and activity type).
+#'
+#' @param activity.types The activity types to compute information for [default: c("mails", "commits", "issues")]
+#' @param range.data The data to base the computation on
+#' @param default.value The default value to add if no information is available per author and activity type
+#'                      [default: NA]
+#' @param aggregation.function The function that should be used to aggregate the activity data (e.g., minimum for
+#'                             first activity, or maximum for last activity)
+#'
+#' @return A list containing per author a list of aggregated activity dates named with the corresponding activity type.
+#'         Empty list if there are no activities in \code{range.data} at all or none corresponding to the configured
+#'         types in \code{activity.types}
+get.aggregated.activity.data = function(range.data, activity.types = c("commits", "mails", "issues"),
+                                        default.value = NA, aggregation.function) {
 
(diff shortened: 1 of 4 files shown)


util-networks-covariates.R
```diff
@@ -1582,43 +1657,68 @@ add.vertex.attribute.issue.is.pull.request = function(list.of.networks, project.
...
+#' Helper function for last activity: computes last activity information per person and activity type.
+#
+#' @param ... See the parameters of function \code{get.aggregated.activity.data}
+#'
+#' @return A list containing per author a list of last activity dates named with the corresponding activity type.
+#'         Empty list if there are no activities in \code{range.data} at all or none corresponding to the configured
+#'         types in \code{activity.types}
+#'
+#' @seealso get.aggregated.activity.data
+get.last.activity.data = function(...) { get.aggregated.activity.data(..., aggregation.function = max) }
```
bockthom, line 1679: As I wanted to reuse code instead of duplicating it, I created a new helper function that is now called by `get.first.activity.data` and by `get.last.activity.data` with the appropriate `aggregation.function`. To avoid copying all the other parameters and their descriptions, I tried `...` here. While this prevents copying the entire parameter descriptions, this also has the drawback that the necessary parameters are not displayed anymore (e.g., also not by tooltips in IDEs), which makes it harder for coronet users to find out which parameters are necessary. The alternative to `...` would be, to list all parameters again.

For the `add.vertex.attribute.*.activity` functions I have tried the alternative option and copied all the parameters.

Which of the two variants do you prefer? In the final version of this PR, I will use the same way of implementation for both the data functions and the vertex-attribute functions. That is, I will change one of the two kinds of functions depending on your advise.

bockthom: As discussed with @maxloeffler, I have replaced the `...` by the actual parameters.
Now, this PR is ready to be reviewed @maxloeffler.

maxloeffler:
@bockthom I have had a closer look at the changes submitted in this PR and I could not find major issues with it! I only spotted some minor stylistic ones, I will list them below.

util-networks-covariates.R
```diff
@@ -688,30 +688,105 @@ add.vertex.attribute.author.email = function(list.of.networks, project.data, nam
...
 #' Add first activity attribute.
 #'
-#' @param list.of.networks The network list.
-#' @param project.data The project data.
+#' @param list.of.networks The network list
+#' @param project.data The project data
 #' @param activity.types The kinds of activity to use as basis: One or more of \code{mails}, \code{commits} and
-#'                       \code{issues}. [default: c("mails", "commits", "issues")]
+#'                       \code{issues} [default: c("mails", "commits", "issues")]
 #' @param name The attribute name to add. [default: "first.activity"]
```
maxloeffler, line 695: Please also remove the punctuation symbol (`.`)  in the description of the `name` parameter here an in the description of `add.vertex.attribute.author.last.activity`.

util-networks-covariates.R
```diff
@@ -768,9 +843,9 @@ add.vertex.attribute.author.first.activity = function(list.of.networks, project.
 #'
 #' @return A list of networks with the added attribute
 add.vertex.attribute.author.active.ranges = function(list.of.networks, project.data, name = "active.ranges",
-                                              activity.types = c("mails", "commits", "issues"),
-                                              default.value = list(),
-                                              combine.activity.types = FALSE) {
+                                                     activity.types = c("mails", "commits", "issues"),
+                                                     default.value = list(),
+                                                     combine.activity.types = FALSE) {
```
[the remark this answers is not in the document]
bockthom: Thanks for pointing this out! I just stumbled over one function where the indentation was broken, but I did not recognize that also many other functions suffered from a similar mis-indentation. I have fixed that now - could you please check if it is ok now or whether I have still overlooked anything?

util-networks-covariates.R
```diff
@@ -688,30 +688,105 @@ add.vertex.attribute.author.email = function(list.of.networks, project.data, nam
...
+                                                      activity.types = c("mails", "commits", "issues"),
+                                                      name = "aggregated.activity",
                                                aggregation.level = c("range", "cumulative", "all.ranges",
                                                                      "project.cumulative", "project.all.ranges",
                                                                      "complete"),
                                                default.value = NA,
-                                               combine.activity.types = FALSE) {
+                                               combine.activity.types = FALSE,
+                                               aggregation.function,
+                                               data.aggregation.function) {
```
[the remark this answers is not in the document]
bockthom: You are right, `get.{first/last}.activity.data` could become obsolete if we would directly pass the proper aggregation function. That is a very good idea! In general, I would also be in favor of your suggestion.

However, there is one reason, why I explicitly wanted to keep `get.{first/last}.activity.data`: While these two functions were intended to be helper functions, they are also pretty useful in other contexts. I have a couple of use cases where I use `get.{first/last}.activity.data` directly from outside coronet, as a nice utility function. For this purpose, these to functions are, of course, in the wrong module (I guess they would better fit into `util-data-misc.R`, for instance). However, lets keep them in `util-networks-covariates.R`, for historic reasons, and because the covariates functions use them as direct helper functions. And since these two functions will be used from outside, we can also use them within.

And there is one additional advantage of separating `aggregation.function` and `data.aggregation.function`: If someone wants to aggregate within data sources differently than across data sources (e.g., extract the earliest commit, mail, and issue, but extract the latest "earliest activity" of the three data sources, the current implementation allows for that). Not sure if this will ever be used by anyone - but we provide the possibility to do so :smile:

maxloeffler:
Thank you @bockthom for correcting the misalignments all over `util-networks-covariates.R`! I only spotted three tiny misalignments that slipped through (see below). Apart from those, I approve this PR.

maxloeffler:
The elements in the `aggregation.level` `c(...)` are not one space off:

https://github.com/se-sic/coronet/blob/4a593dc37a3ff48389c09708e9978a3dbdd7c83c/util-networks-covariates.R#L938-L943

Same goes for the elements in the `aggregation.level` `c(...)` here:

https://github.com/se-sic/coronet/blob/4a593dc37a3ff48389c09708e9978a3dbdd7c83c/util-networks-covariates.R#L635-L642

Lastly, these parameters are just overall off:

https://github.com/se-sic/coronet/blob/4a593dc37a3ff48389c09708e9978a3dbdd7c83c/util-networks-covariates.R#L1147-L1151

*Note*: As you did not edited these lines, it would not let me create a regular review comment and I had to resort to this.

bockthom:
I've fixed the three misalignment that you have spotted @maxloeffler (at least, I hope so). If  all the misalignments you have spotted are fixed now and if you agree with my descriptions in the NEWS.md, I'd be happy if you could also formally approve this PR :smile:

maxloeffler (approved):

bockthom:
As codecov is fine with the coverage of my tests, and as @maxloeffler has already approved the PR, I will merge right away.

PR #276, opened 2024-12-11
Additional Core/Peripheral Classification Methods

Leo-Send:
### Prerequisites

### Description

Add four new metric which can be used for the classification of authors into core and peripheral:
Betweenness, which measures the number of shortest paths between developers that go through a given developer vertex;
Closeness, which measures how close a developer is to all others by taking the inverse of the sum of all of it's shortest paths;
Pagerank, which is based on Google's Pagerank algorithm, which is closely related to Eigenvector Centrality;
Eccentricity, which measures the distance to the furthest developer vertex.

### Added

- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, e27acb5ad7b4c4532e787d793301d06f13e8bbba, 217880809a07f970b055bd3b84c05cb6e61ac6af)

### Changed/Improved

- Include core/peripheral classification in the `README.md` (PR #276, 6101e11f5c4ac1b5883e85cebd01a3cd7c76e056, c6744c00d3dc0d4e45a96c2d80ae78727e22cce2, 5fc2da5ece6604a6a87d8dd5f79237a82fb2b5ca)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,7 @@
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
 - Refactor function `construct.edge.list.from.key.value.list` to be more readable (PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
 - Update necessary `igraph` version to 2.1.0 in `README.md` (PR #274, 6c3bcd1a2366d0d3a176d9fde95b8356b0158da3)
+- Include core/peripheral classification in the `README.md` (PR #276, )
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 - Add `remove.duplicate.edges` function that takes a network as input and conflates identical edges (PR #268, d9a4be417b340812b744f59398ba6460ba527e1c, 0c2f47c4fea6f5f2f582c0259f8cf23af985058a, c6e90dd9cb462232563f753f414da14a24b392a3)
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
-- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
+- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, e27acb5ad7b4c4532e787d793301d06f13e8bbba, 217880809a07f970b055bd3b84c05cb6e61ac6af)
 
 ### Changed/Improved
 
@@ -31,7 +31,7 @@
 - Explicitly add R version 4.4 to the CI test pipeline (c8e6f45111e487fadbe7f0a13c7595eb23f3af6e)
 - Refactor function `construct.edge.list.from.key.value.list` to be more readable (PR #263, 05c3bc09cb1d396fd59c34a88030cdca58fd04dd)
 - Update necessary `igraph` version to 2.1.0 in `README.md` (PR #274, 6c3bcd1a2366d0d3a176d9fde95b8356b0158da3)
-- Include core/peripheral classification in the `README.md` (PR #276, )
+- Include core/peripheral classification in the `README.md` (PR #276, 6101e11f5c4ac1b5883e85cebd01a3cd7c76e056, c6744c00d3dc0d4e45a96c2d80ae78727e22cce2, 5fc2da5ece6604a6a87d8dd5f79237a82fb2b5ca)
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 - Add `remove.duplicate.edges` function that takes a network as input and conflates identical edges (PR #268, d9a4be417b340812b744f59398ba6460ba527e1c, 0c2f47c4fea6f5f2f582c0259f8cf23af985058a, c6e90dd9cb462232563f753f414da14a24b392a3)
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
+- Add four new metric which can be used for the classification of authors into core and peripheral: Betweenness, Closeness, Pagerank and Eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
 
 ### Changed/Improved
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 - Add `remove.duplicate.edges` function that takes a network as input and conflates identical edges (PR #268, d9a4be417b340812b744f59398ba6460ba527e1c, 0c2f47c4fea6f5f2f582c0259f8cf23af985058a, c6e90dd9cb462232563f753f414da14a24b392a3)
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
-- Add four new metric which can be used for the classification of authors into core and peripheral: Betweenness, Closeness, Pagerank and Eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
+- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
 
 ### Changed/Improved
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 - Add `remove.duplicate.edges` function that takes a network as input and conflates identical edges (PR #268, d9a4be417b340812b744f59398ba6460ba527e1c, 0c2f47c4fea6f5f2f582c0259f8cf23af985058a, c6e90dd9cb462232563f753f414da14a24b392a3)
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
-- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, e27acb5ad7b4c4532e787d793301d06f13e8bbba, 217880809a07f970b055bd3b84c05cb6e61ac6af)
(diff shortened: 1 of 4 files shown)


util-core-peripheral.R
```diff
@@ -669,6 +703,50 @@ get.author.class.network.hierarchy = function(network, result.limit = NULL,
     return(result)
 }
 
+get.author.class.network.betweenness = function(network, result.limit = NULL,
+                                              restrict.classification.to.authors = NULL) {
```
bockthom, line 731: Indentation wrong. (Also applies to some of the functions below.)

NEWS.md
```diff
@@ -15,6 +15,7 @@
 - Add commit network as a new type of network. It uses commits as vertices and connects them either via cochange or commit interactions. This includes adding new config parameters and the function `add.vertex.attribute.commit.network` for adding vertex attributes to a commit network (PR #263, ab73271781e8e9a0715f784936df4b371d64c338, ab73271781e8e9a0715f784936df4b371d64c338, cd9a930fcb54ff465c2a5a7c43cfe82ac15c134d)
 - Add `remove.duplicate.edges` function that takes a network as input and conflates identical edges (PR #268, d9a4be417b340812b744f59398ba6460ba527e1c, 0c2f47c4fea6f5f2f582c0259f8cf23af985058a, c6e90dd9cb462232563f753f414da14a24b392a3)
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
+- Add four new metric which can be used for the classification of authors into core and peripheral: Betweenness, Closeness, Pagerank and Eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
```
bockthom, line 18: metric :arrow_right: metrics
which :arrow_right: that
And there is no need to capitalize the metrics' names. But please put a comma before the final occurrence of "and".

bockthom (changes requested):
I had a quick look at the implementation (but not yet at the tests).
Please find my initial comments below.

util-core-peripheral.R
```diff
@@ -231,6 +241,30 @@ get.author.class.by.type = function(network = NULL,
...
+    } else if (type == "network.betweenness") {
+        betweenness.centrality.vec = igraph::betweenness(network, directed = TRUE)
+        ## Construct centrality dataframe
+        centrality.dataframe = data.frame(author.name = names(betweenness.centrality.vec),
+                                          centrality = as.vector(betweenness.centrality.vec))
+    } else if (type == "network.closeness") {
+        closeness.centrality.vec = igraph::closeness(network)
+        ## Construct centrality dataframe
+        centrality.dataframe = data.frame(author.name = names(closeness.centrality.vec),
+                                          centrality = as.vector(closeness.centrality.vec))
```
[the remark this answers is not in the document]
Leo-Send: Intuitively, both "all" and "out" make sense to me, as they refer to how distance is measured within the graph, with "all" considering edges as undirected and "out" as directed. "in" would consider the edges as directed in the opposite direction.
In our context, I would probably use "all" as I believe it would lead to less isolated vertices.

README.md
```diff
@@ -375,6 +376,49 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
...
+    * eigenvector centrality measures the importance of vertices within a graph by granting a higher score for outgoing edges proportional to the score of the target vertex
+- `network.hierarchy`
+    * calculates scores for authors based on the hierarchy found within an author network
+    * hierarchical scores are calculated by dividing the vertex degree by the clustering coefficient of each vertex
+- `network.betweenness`
+    * calculates scores for authors based on the betweenness of vertices in an author network
+    * betweenness measures the number of shortest paths that go through each vertex
+- `network.closeness`
+    * calculates scores for authors based on the closeness of vertices in an author network
+    * closeness measures how close vertices are to each other by taking the inverse if the sum of all their shortest paths
```
bockthom, line 398: if :arrow_right: of

and what is meant by "all their shortest paths"? Shortest paths to what? And what is the "inverse" in this case?

README.md
```diff
@@ -375,6 +376,49 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
...
+- `mail.thread.count`
+    * calculates scores based on the number of mail threads each author participated in
+- `issue.count`
+    * calculates scores based on the number of issues each author participated in
+- `issue.comment.count`
+    * calculates scores based on the number of comments each author made in issues
+- `issue.commented.in.count`
+    * calculates scores based on the number of issues each author commented in
+- `issue.created.count`
+    * calculates scores based on the number of issues each created
```
bockthom, line 420: each :arrow_right: each author

README.md
```diff
@@ -375,6 +376,49 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
 
 In some cases, it is not necessary to build a network to get the information you need. Therefore, please remember that we offer the possibility to get the raw data or mappings between, e.g., authors and the files they edited. The data inside an instance of `ProjectData` can be accessed independently. Examples can be found in the file `showcase.R`.
 
+#### Core/Peripheral Classification
+
+Core/Peripheral classification descibes the process of dividing the authors of a project into either `core` or `peripheral` developers based on the principle that the core developers contribute 80% of all work in a given project. In practice, this is done by assigning scores to developers to approximate their importance in a project and then dividing the authors into `core` or `peripheral` based on these scores such that the desired 80:20 split is achieved.
+
+In this section, we provide descriptions of the different algorithms we provide for classifying authors into core or peripheral authors.
```
bockthom, line 383: Could we split that into two subsections, namely count-based and network-based metrics? And I would put the count-based ones first as they are simpler and easier to understand.

README.md
```diff
@@ -375,6 +376,49 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
 
 In some cases, it is not necessary to build a network to get the information you need. Therefore, please remember that we offer the possibility to get the raw data or mappings between, e.g., authors and the files they edited. The data inside an instance of `ProjectData` can be accessed independently. Examples can be found in the file `showcase.R`.
 
+#### Core/Peripheral Classification
+
+Core/Peripheral classification descibes the process of dividing the authors of a project into either `core` or `peripheral` developers based on the principle that the core developers contribute 80% of all work in a given project. In practice, this is done by assigning scores to developers to approximate their importance in a project and then dividing the authors into `core` or `peripheral` based on these scores such that the desired 80:20 split is achieved.
```
bockthom, line 381: Hm, I am not sure whether we should put the 80/20 percent that prominently into our README. Coronet is flexible enough to also support other thresholds - either by using parameters or re-assigning the corresponding constant. This is why I would like to name the constant and omit the concrete numbers. It's the user's choice to set the value in accordance with their needs. If the constant is mentioned here, the user has the possibility to re-assign the needed value to this constant. However, we should say here what the default value is and why we 80/20 is the default threshold (just referring to literature in general, without citing a specific paper).

util-core-peripheral.R
```diff
@@ -260,6 +260,7 @@ get.author.class.by.type = function(network = NULL,
         eccentricity.vec = igraph::eccentricity(network)
         ## since core developers are expected to have a lower eccentricity,
         ## we need to invert all non-zero values
+        ## all entries with value zero are isolated vertices, so the value is already correct
```
bockthom, line 263: Could you please add a period at the end of the first sentence - otherwise it is difficult to guess where the first sentence ends.
And what does it mean that "the value is already correct"? Instead of "correct" I would prefer another wording, maybe something like "the expected value" or something similar?

README.md
```diff
@@ -375,6 +376,49 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
 
 In some cases, it is not necessary to build a network to get the information you need. Therefore, please remember that we offer the possibility to get the raw data or mappings between, e.g., authors and the files they edited. The data inside an instance of `ProjectData` can be accessed independently. Examples can be found in the file `showcase.R`.
 
+#### Core/Peripheral Classification
```
bockthom, line 379: Please use upper-case characters in headings only at the beginning of the heading - this would be inconsistent with the other headings otherwise. (Not only here, but also in the table of contents.)

Independent of that, it is debatable whether the "P" should be upper- or lower-case. Not sure what would lead to highest consistency.

bockthom (changes requested):
As I haven't had a look at the tests yet, I have now done so. I only have a few minor comments (but I did not check the actual centrality values; especially for eccentricity, I still don't have an impression whether the values do make sense or not :see_no_evil:)

In addition to my comments below, please don't forget about the changes to the README regarding "author networks" that we have agreed on in today's meeting.

README.md
```diff
@@ -404,27 +404,27 @@ In this section, we provide descriptions of the different algorithms we provide
 
 ##### Network-based metrics
 
-In this section, we provide descriptions of the different algorithms we provide for classifying authors into core or peripheral authors using metrics that are used on author networks.
+In this section, we provide descriptions of the different algorithms we provide for classifying authors into core or peripheral authors using metrics that are used on author networks. Note that the provided methods can be used for any network and not just author networks. The classification would then occur regarding the type of the vertices, i.e. an artifact network would result in a classification of the artifacts based on their importance in the network.
```
bockthom, line 407: Instead of `i.e. ...`, please use ` (e.g., an artifact network ...).`
And please don't use "importance", as this is already an interpretation of the metric. Just stick to "centrality" instead.

NEWS.md
```diff
@@ -20,6 +20,7 @@
...
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
+- Add four new metric which can be used for the classification of authors into core and peripheral: Betweenness, Closeness, Pagerank and Eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
```
bockthom, line 23: Could you please replace "new metric which" by "new network metrics that".
And you don't need to have upper-case letters for the metrics names.
And please add a comma before "and eccentricity". Thanks!

bockthom (commented):
Just two comments regarding wording:

util-core-peripheral.R
```diff
@@ -236,6 +236,10 @@ get.author.class.by.type = function(network = NULL,
         ## is 'Inf'. We do not get any complications here because there are no authors with
         ## degree == 0 and a CC > 0 (i.e., the hierarchy value would really be 0). Authors with
         ## a CC == NaN (degree < 2) will stay with their hierarchy value of NaN, accordingly.
+        ## Note that this could lead to authors who should be high up in the hierarchy
+        ## being set to hierarchy 0 because they interact with clusters that
+        ## have no connection between them. We accept this because the case should
+        ## almost never occur in all but the smallest networks.
```
bockthom, line 242: Thanks for adding this comment. While I agree with the first sentence, I don't agree with the second sentence. Could you please rephrase the second sentence? First of all, I'd like to mention that we cannot distinguish between the case where the hierarchy should be low and the case where the hierarchy should be high, as we would need more information about the clusters that are not connected. And second, we should state that we assume that such a case where hierarchy is set to 0 although it should be high occurs only rarely. ("almost never" sounds incorrect, and also the information about "smallest networks" is irrelevant in my opinion.

NEWS.md
```diff
@@ -20,7 +20,7 @@
...
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
-- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, 65d5c9cc86708777ef458b0c2e744ab4b846bdd1, b392d1a125d0f306b4bce8d95032162a328a3ce2, c5d37d40024e32ad5778fa5971a45bc08f7631e0)
+- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, e27acb5ad7b4c4532e787d793301d06f13e8bbba, 217880809a07f970b055bd3b84c05cb6e61ac6af)
```
bockthom, line 23: Did you remove one of the three previously mentioned commits on purpose or inadvertently? So, I agree with the two commits mentioned, and I don't know which one else to add, so I assume that there was one commit too much in the previous version and everything is fine now. I just wanted to double check that this is intended.

bockthom (changes requested):

NEWS.md
Leo-Send: It is indeed intended, previously I had another commit in there that only included tests.

NEWS.md
```diff
@@ -20,6 +20,7 @@
...
 - Add `cumulative` as an argument to `construct.ranges` which enables the creation of cumulative ranges from given revisions (PR #268, a135f6bb6f83ccb03ae27c735c2700fccc1ee0c8, 8ec207f1e306ef6a641fb0205a9982fa89c7e0d9)
 - Add function `get.last.activity.data` to compute developers' last activities in a project, as well as function `add.vertex.attribute.author.last.activity` to add a developer's date of last activity as vertex attribute to a network, as well as helper functions `get.aggregated.activity.data` and `add.vertex.attribute.author.aggregated.activity` to allow for other activity aggregations than first and last activity (PR #275, 9f231612fcd33a283362c79b35a94295ff3d4ef9, 8660ed763ba4b69e909e7fbb01e27e1999522047)
+- Add four new metric that can be used for the classification of authors into core and peripheral: betweenness, closeness, pagerank, and eccentricity (PR #276, e27acb5ad7b4c4532e787d793301d06f13e8bbba, 217880809a07f970b055bd3b84c05cb6e61ac6af)
```
hechtlC, line 23: metrics

util-core-peripheral.R
```diff
@@ -226,11 +236,43 @@ get.author.class.by.type = function(network = NULL,
...
+    } else if (type == "network.pagerank") {
+        pagerank.centrality.vec = igraph::page_rank(network, directed = TRUE)[["vector"]]
+        ## Construct centrality dataframe
+        centrality.dataframe = data.frame(author.name = names(pagerank.centrality.vec),
+                                          centrality = as.vector(pagerank.centrality.vec))
+    } else if (type == "network.eccentricity") {
+        eccentricity.vec = igraph::eccentricity(network)
+        ## since core developers are expected to have a lower eccentricity,
+        ## we need to invert all non-zero values.
+        ## all entries with value zero are isolated vertices, so the expected value is also zero.
```
hechtlC, line 269: If you use punctuation at the end of the sentences in these comments, then you need to start the sentence in upper case.

util-core-peripheral.R
```diff
@@ -669,6 +711,146 @@ get.author.class.network.hierarchy = function(network, result.limit = NULL,
     return(result)
 }
 
+## * Betweenness-based classification --------------------------------------
+
+#' Classify authors into "core" and "peripheral" based on the betweenness-centrality of author vertices in the network
+#' and return the classification result.
+#'
+#' The details of the classification algorithm is explained in the documentation of \code{get.author.class.by.type}.
```
hechtlC, line 719: are explained
Also for the other functions

README.md
```diff
@@ -375,6 +378,55 @@ Analogously, the `NetworkConf` parameter `unify.date.ranges` enables this very f
 
 In some cases, it is not necessary to build a network to get the information you need. Therefore, please remember that we offer the possibility to get the raw data or mappings between, e.g., authors and the files they edited. The data inside an instance of `ProjectData` can be accessed independently. Examples can be found in the file `showcase.R`.
 
+#### Core/Peripheral classification
+
+Core/Peripheral classification descibes the process of dividing the authors of a project into either `core` or `peripheral` developers based on the principle that the core developers contribute most of the work in a given project. The concrete threshold can be configured in `CORE.THRESHOLD` and is set to 80%  per default, a value commonly used in literature. In practice, this is done by assigning scores to developers to approximate their importance in a project and then dividing the authors into `core` or `peripheral` based on these scores such that the desired split is achieved.
```
hechtlC, line 383: describes
also two spaces after 80%

hechtlC (changes requested):
First of all thank you for the enhancement @Leo-Send!
I have just found some small typos and wording issues. Once these are fixed, we can merge this PR.

hechtlC (approved):

PR #278, opened 2025-01-29
Implement splitting of simplified networks

maxloeffler:
### Prerequisites

### Description

Changed:

* Allow splitting of simplified networks (which was previously considered illegal)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,18 @@
 
 # coronet – Changelog
 
+## unversioned
+
+### Added
+
+- Add the possibility to split networks that contain simplified edges (PR #278, 9798d33512dcdf50d3b88a1223fc4913a2a88eeb, 0ed437c14423c1917f1ba470e7e55db4626d380b, 67a6651b94d50cb7c2ab4a74888b0556d607b102, 98ef83158204be2a67b115cb25df5ba375cccf60, 7ec4d83fdeb308a24a350acd808941807b9511f1, 637d62ab70f098f26f241e588a99cdc49d10f56a, 2c70666f128f96a3a573f29a0cbbef14d803d193, 1cbc6fa36859d6db3a7ff4493ef19763e87d2de3, 41788ff029d038969bfc6b5773e919201c5ac595, b042c0dd08e2229514ccd25dee7a119f25b1ab45, 36d23d657f412aa1953c4773076e593273f19d8e, 402c256d9a05e4ffb297d4ea1fc25d0230787bc0, 54af2b19a112070f10d191b98b055482748426a7, 894414a4a970822b9ecd59c0b6c480860707f636, 0fe32a259ef703c2de79135bfa6932a595fdc1c5)
+
+### Changed/Improved
+
+- For consistency reasons: Ensure that the values of edge attributes are always lists even when they represent singular values (PR #278, 6fae1843740ed8e48c89c2ee4e61f995b5d0b8f5, 416c817998540fc0b82d9959574838b571b4d6fb)
+
+### Fixed
+
 ## 5.0
 
 ### Announcement
--- a/tests/test-networks-author.R
+++ b/tests/test-networks-author.R
@@ -21,7 +21,7 @@
 ## Copyright 2018 by Jakob Kronawitter <kronawij@fim.uni-passau.de>
 ## Copyright 2018-2019 by Anselm Fehnker <fehnker@fim.uni-passau.de>
 ## Copyright 2021 by Johannes Hostert <s8johost@stud.uni-saarland.de>
-## Copyright 2023-2024 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
+## Copyright 2023-2025 by Maximilian Löffler <s8maloef@stud.uni-saarland.de>
 ## Copyright 2024 by Leo Sendelbach <s8lesend@stud.uni-saarland.de>
 
 ## All Rights Reserved.
--- a/tests/test-networks-author.R
+++ b/tests/test-networks-author.R
@@ -457,7 +457,8 @@ test_that("Network construction of the undirected simplified author-cochange net
                           list("Base_Feature", "Base_Feature"))),
         weight = 2,
         type = TYPE.EDGES.INTRA,
-        relation = "cochange"
+        relation = I(list(list("cochange", "cochange"), list("cochange", "cochange"),
+                          list("cochange", "cochange"), list("cochange", "cochange")))
     )
 
     ## remove the 'AsIs' class from the edge attributes that have been inserted via `I(...)`
@@ -466,6 +467,7 @@ test_that("Network construction of the undirected simplified author-cochange net
     data[["hash"]] = unclass(data[["hash"]])
     data[["file"]] = unclass(data[["file"]])
     data[["artifact"]] = unclass(data[["artifact"]])
+    data[["relation"]] = unclass(data[["relation"]])
 
     ## build expected network
     network.expected = igraph::graph_from_data_frame(data, directed = FALSE, vertices = authors)
--- a/tests/test-networks-multi-relation.R
+++ b/tests/test-networks-multi-relation.R
@@ -81,8 +81,7 @@ test_that("Network construction of the undirected author network with relation =
                                      as.list(rep(NA, 4)))),
                       weight = 1,
                       type = TYPE.EDGES.INTRA,
-                      relation = c(rep("cochange", 8),
-                                   rep("mail", 4)),
+                      relation = I(c(as.list(rep("cochange", 8)), as.list(rep("mail", 4)))),
                       message.id = I(c(as.list(rep(NA, 8)),
                                        "<4cbaa9ef0802201124v37f1eec8g89a412dfbfc8383a@mail.gmail.com>",
                                        "<6784529b0802032245r5164f984l342f0f0dc94aa420@mail.gmail.com>",
@@ -98,6 +97,7 @@ test_that("Network construction of the undirected author network with relation =
     data[["hash"]] = unclass(data[["hash"]])
     data[["file"]] = unclass(data[["file"]])
     data[["artifact"]] = unclass(data[["artifact"]])
+    data[["relation"]] = unclass(data[["relation"]])
     data[["message.id"]] = unclass(data[["message.id"]])
     data[["thread"]] = unclass(data[["thread"]])
 
@@ -205,7 +205,7 @@ test_that("Construction of the bipartite network for the feature artifact with a
                                    as.list(rep(NA, 16)))),
                   weight = 1,
                   type = TYPE.EDGES.INTER,
-                  relation = c(rep("issue", 24), rep("mail", 16))
+                  relation = I(c(as.list(rep("issue", 24)), as.list(rep("mail", 16))))
     )
 
     ## remove the 'AsIs' class from the edge attributes that have been inserted via `I(...)`
@@ -215,6 +215,7 @@ test_that("Construction of the bipartite network for the feature artifact with a
     network.expected.data[["thread"]]     = unclass(network.expected.data[["thread"]])
     network.expected.data[["issue.id"]]   = unclass(network.expected.data[["issue.id"]])
     network.expected.data[["event.name"]] = unclass(network.expected.data[["event.name"]])
+    network.expected.data[["relation"]]   = unclass(network.expected.data[["relation"]])
 
     ## 3) build expected network
     network.expected = igraph::graph_from_data_frame(network.expected.data, vertices = vertices,
@@ -309,8 +310,8 @@ test_that("Construction of the multi network for the feature artifact with autho
                                       as.list(rep(NA, 21)))),
                        weight = 1,
                        type = c(rep(TYPE.EDGES.INTRA, 13), rep(TYPE.EDGES.INTER, 27)),
-                       relation = c(rep("cochange", 8), rep("mail", 4), rep("cochange", 1), rep("cochange", 6),
-                                    rep("issue", 21)),
+                       relation = I(c(as.list(rep("cochange", 8)), as.list(rep("mail", 4)), as.list(rep("cochange", 1)), as.list(rep("cochange", 6)),
+                                    as.list(rep("issue", 21)))),
                        message.id = I(c(as.list(rep(NA, 8)),
                                         "<4cbaa9ef0802201124v37f1eec8g89a412dfbfc8383a@mail.gmail.com>",
                                         "<6784529b0802032245r5164f984l342f0f0dc94aa420@mail.gmail.com>",
@@ -335,6 +336,7 @@ test_that("Construction of the multi network for the feature artifact with autho
     edges[["hash"]] = unclass(edges[["hash"]])
     edges[["file"]] = unclass(edges[["file"]])
     edges[["artifact"]] = unclass(edges[["artifact"]])
+    edges[["relation"]] = unclass(edges[["relation"]])
     edges[["message.id"]] = unclass(edges[["message.id"]])
     edges[["thread"]] = unclass(edges[["thread"]])
     edges[["author.name"]] = unclass(edges[["author.name"]])
@@ -425,7 +427,7 @@ test_that("Construction of the multi-artifact bipartite network with artifact re
         artifact = I(c("A", "Base_Feature", "A", "Base_Feature", "Base_Feature", "foo", as.list(rep(NA, 24)))),
         weight = c(rep(1, 30)),
         type = c(rep("Bipartite", 30)),
-        relation = c(rep("cochange", 6), rep("issue", 24)),
+        relation = I(c(as.list(rep("cochange", 6)), as.list(rep("issue", 24)))),
         issue.id = I(c(as.list(rep(NA, 6)),
                        "<issue-jira-ZEPPELIN-328>", "<issue-jira-ZEPPELIN-328>", "<issue-jira-ZEPPELIN-328>",
                        "<issue-jira-ZEPPELIN-328>", "<issue-jira-ZEPPELIN-328>", "<issue-jira-ZEPPELIN-328>",
@@ -444,6 +446,7 @@ test_that("Construction of the multi-artifact bipartite network with artifact re
     edges[["hash"]] = unclass(edges[["hash"]])
     edges[["file"]] = unclass(edges[["file"]])
     edges[["artifact"]] = unclass(edges[["artifact"]])
+    edges[["relation"]] = unclass(edges[["relation"]])
     edges[["issue.id"]] = unclass(edges[["issue.id"]])
     edges[["event.name"]] = unclass(edges[["event.name"]])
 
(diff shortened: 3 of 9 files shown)


maxloeffler:
Hey, I implemented splitting of simplified networks / edges as discussed in our meeting(s). However, there is one case that we should discuss, and I figured that this is a good place to do so. I would be interested in all of your opinions so we can find the best solution.

## How should we deal with numeric attributes on simplified edges?

Lets say we have the `diff.size` attribute set before simplifying two edges into one. The `diff.size` of the simplified edge will be the sum of each its partial edges. In our meeting we established, that there is no good way to separate such attributes again if the two partial edges are now supposed to end up in different bins when we split, so we just remove the attribute.

Here is the catch though, how do we remove the attribute exactly? If we split and have a bin that has some edges that did not have to be separated (i.e., because they were not simplified in the beginning or because all partial edges belong to the same bin) then these edges all have a valid value of the `diff.size` attribute. Now lets say we have to add an additional edge that is only partial. This partial cannot have a semantically valid `diff.size`. So now, do we remove the `diff.size` attribute from the entire bin, i.e., from all edges or do we keep it on all edges where it is valid and set it to `NA` where it isn't?

If we remove it entirely for the sake of not keeping a potentially really useless attribute around, do we also remove it from all other bins? Discussion with @hechtlC revealed that in practice it seems unlikely that even a single bin would only have complete (non-partial) edges, so when we go the step of removing it from all edges in a bin, its not so far fetched to say, that we remove it from all bins altogether.

And finally: What do we do with the `weight`? On first glance it is also a numeric attribute so it should be treated the same. However, we discussed ways to separate it in the last meeting. Having 2 edges simplified into one and the merged edge having a `weight` of 2 *most likely* means that both source edges have had `weight` 1. But this is only an assumption and could also lead to bullshit values as with other numeric attributes. So I personally think that we should treat it as other numeric attributes. However, @hechtlC raised concern that it may be critical to have a `weight` attribute on edges at any time.

bockthom:
> ## How should we deal with numeric attributes on simplified edges?

> [...] So now, do we remove the `diff.size` attribute from the entire bin, i.e., from all edges or do we keep it on all edges where it is valid and set it to `NA` where it isn't?

Having partial information is as valuable as having no information at all. Thus, we should remove it from all edges. **Consistency should take precedence over partial information.** (Especially as we don't lose any information here, since the entire network still is present unless the user deliberately deletes the global network from memory.)

> If we remove it entirely for the sake of not keeping a potentially really useless attribute around, do we also remove it from all other bins?

Yes, same reason as above: *Consistency should take precedence over partial information.* If we remove it from one of the networks that result from splitting, then I (as a user) would expect that all splits contain the same attributes, that is, that it is removed from all networks. I totally agree with @hechtlC here.

> And finally: What do we do with the `weight`? On first glance it is also a numeric attribute so it should be treated the same. However, we discussed ways to separate it in the last meeting. Having 2 edges simplified into one and the merged edge having a `weight` of 2 _most likely_ means that both source edges have had `weight` 1. But this is only an assumption and could also lead to bullshit values as with other numeric attributes.

Could you come up with an example where we would get bullshit values? I can't think of such an example right now. But if you can come up with any, we can discuss it.

> So I personally think that we should treat it as other numeric attributes. However, @hechtlC raised concern that it may be critical to have a `weight` attribute on edges at any time.

According to our corresponding section in our [README.md](https://github.com/se-sic/coronet?tab=readme-ov-file#vertex-and-edge-attributes), **`weight` is a mandatory edge attribute**. Therefore, we cannot remove it if we stick to our own rules. Assuming that this attribute has been created according to our rules, it should be possible to determine the correct weight during splitting.

bockthom:
@maxloeffler Would you like us to review this PR already, or do you plan to add additional commits or make changes to the existing commits first?
As soon as you would like us to review this PR, please mark this PR as "Ready for review" :wink: 
In addition, please don't forget to rebase onto `dev`

util-split.R
```diff
@@ -1180,21 +1286,36 @@ split.unify.range.names = function(ranges) {
...
-    dates.breaks[length(dates.breaks)] = max(dates) + 1
+    dates.breaks[length(dates.breaks)] = max(flat.dates) + 1
     ## generate charater strings for bins
     dates.breaks.chr = get.date.string(head(dates.breaks, -1))
 
     ## find bins for given dates
-    dates.bins = findInterval(dates, dates.breaks, all.inside = FALSE)
-    ## convert to character factor and set factor's levels appropriately
-    dates.bins = factor(dates.breaks.chr[dates.bins], levels = dates.breaks.chr)
+    if (is.list(dates)) {
```
bockthom, line 1304: Same as above. I am not sure whether this differentiation here is useful. We should think about the possible use cases of this function and whether it is possible to use this function manually with non-list `dates`.

Anyway, it might be necessary to adjust the documentation of parameter and return value, as they don't consider lists already. This might potentially apply also to other functions. However, before we make the documentation more complex, we should discuss the usefulness of potential documentation changes...

tests/test-networks.R
```diff
@@ -90,24 +90,28 @@ test_that("Simplify basic multi-relational network", {
...
 
     ## simplify network without simplifying multiple relations into single edges
     network.simplified = simplify.network(network, simplify.multiple.relations = FALSE)
     assert.networks.equal(network.simplified, network.expected)
 
+    network.expected = igraph::make_empty_graph(n = 0, directed = FALSE) +
+        igraph::vertices("A", "B", type = TYPE.ARTIFACT, kind = "feature") +
+        igraph::edges("A", "B", type = TYPE.EDGES.INTRA, relation = list(list("mail", "cochange",
+                                                                              "mail", "cochange",
+                                                                              "mail", "cochange")))
```
bockthom, line 110: I am confused by two different `expected.networks` here. So is there any assertion missing regarding the first expected network, or can it just be removed? Please double-check to make sure that we don't overlook anything here.

tests/test-networks.R
```diff
@@ -155,7 +159,8 @@ test_that("Simplify author-network with relation = c('cochange', 'mail') using b
                          list("Base_Feature", "Base_Feature"), as.list(rep(NA, 2)), as.list(rep(NA, 2)))
     data$weight = rep(2, 6)
     data$type = rep(TYPE.EDGES.INTRA, 6)
-    data$relation = c(rep("cochange", 4), rep("mail", 2))
+    data$relation = list(list("cochange", "cochange"), list("cochange", "cochange"), list("cochange", "cochange"),
+                       list("cochange", "cochange"), list("mail", "mail"), list("mail", "mail"))
```
bockthom, line 163: If you've thought I won't catch this: I've caught it! :upside_down_face:

tests/test-networks.R
```diff
@@ -272,14 +280,63 @@ test_that("Simplify multiple basic multi-relational networks", {
...
+                         "mail", "mail",
+                         list("mail", "cochange"), list("mail", "cochange", "mail"),
+                         list("mail", "mail"), list("mail", "mail"))
+
+    ## build expected network
+    network.built = igraph::graph_from_data_frame(data, vertices = c("A", "B", "C", "D", "E", "F"),
+                                                  directed = FALSE)
+
+    ## ---------------------- simplify.multiple.relations == FALSE -------------------------- ##
+
```
bockthom, line 316: If you do that, I would expect to have a second check within this test below in which `simplify.multiple.relations` is `TRUE`...

tests/test-networks.R
bockthom, line 1: As you have changed this file (tests/test-networks.R) significantly within this PR, I would expect you to update your copyright header here.

tests/test-networks-multi-relation.R
bockthom, line 1: As you have changed this file (tests/test-networks-multi-relation.R) significantly within this PR, I would expect you to update your copyright header here.

tests/test-networks-author.R
bockthom, line 1: As you have changed this file (tests/test-networks-author.R) significantly within this PR, I would expect you to update your copyright header here.

tests/test-split-network-time-based.R
```diff
@@ -37,6 +37,81 @@ if (!dir.exists(CF.DATA)) CF.DATA = file.path(".", "tests", "codeface-data")
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Split network -----------------------------------------------------------
 
+## * helper functions ------------------------------------------------------
+
+#' Construct a (sub-)network with the same vertices and new edges
+#'
+#' Note: This method is an adoption of \code{split.network.by.bins} with less functionality.
```
bockthom, line 44: This needs some additional explanation. First, it is unclear what the "missing" or "kept" functionality is. And, second, it is unclear why such a method is necessary or justified to be used in tests. So, I basically see why you have introduced these helper functions. But I would like to be a little bit more concrete instead of just saying "with less functionality". This also holds for second helper function below.

tests/test-split-network-time-based.R
```diff
@@ -37,6 +37,81 @@ if (!dir.exists(CF.DATA)) CF.DATA = file.path(".", "tests", "codeface-data")
...
+
+    return(subnet)
+}
+
+#' Extract a partial edge from a simplified edge with multiple components
+#'
+#' Note: This method is an adoption of \code{split.network.by.bins} with less functionality.
+#'
+#' @param edge The edge to extract the partial edge from
+#' @param partial The indicess of the partial edge to extract
```
bockthom, line 86: Mismatch between parameter documentation and actual parameter name. One is singular, the other plural.

tests/test-split-network-time-based.R
```diff
@@ -481,3 +627,150 @@ patrick::with_parameters_test_that("Split a list of networks time-based with equ
...
+
+    ## construct expected networks
+    edges = igraph::as_data_frame(network, "edges")
+    expected = list(
+        "2025-01-01 12:00:00-2025-01-01 12:01:00" = subnet.with.new.edges(network, rbind(edges[1, ],
+                                                                                         edges[3, ])),
+        "2025-01-01 12:01:00-2025-01-01 12:02:00" = subnet.with.new.edges(network, edges[2, ]),
+        "2025-12-01 12:02:00-2025-01-01 12:03:01" = subnet.with.new.edges(network, edges[4, ])
+    )
+    ## do not the remove the 'diff.size' attribute as all subnet
```
bockthom, line 728: subnet :arrow_right: subnets

tests/test-split-network-time-based.R
```diff
@@ -481,3 +627,150 @@ patrick::with_parameters_test_that("Split a list of networks time-based with equ
...
+    ## construct expected networks
+    edges = igraph::as_data_frame(network, "edges")
+    expected = list(
+        "2025-01-01 12:00:00-2025-01-01 12:01:00" = subnet.with.new.edges(network, rbind(edges[1, ],
+                                                                                         edges[3, ])),
+        "2025-01-01 12:01:00-2025-01-01 12:02:00" = subnet.with.new.edges(network, extract.partial.edge(edges[2, ], 1)),
+        "2025-12-01 12:02:00-2025-01-01 12:03:01" = subnet.with.new.edges(network, rbind(extract.partial.edge(edges[2, ], 2),
+                                                                                         edges[4, ]))
+    )
+    ## remove the 'diff.size' attribute as subnet 2 and 3
```
bockthom, line 692: subnet ➡️ subnets

bockthom (commented):
Thanks for working on this challenging PR @maxloeffler.
While I basically agree with all your implementation, I have a couple of questions here and there (mostly regarding comments, but also regarding improving code readability). See further details in the individual comments below.

In addition, I would like you double-check whether the function `get.data.sources.from.relations` still works in every conceivable situation, after the changes that we have made to the handling of the `relation` attribute.

util-split.R
```diff
@@ -877,19 +882,120 @@ split.dataframe.by.bins = function(df, bins) {
...
+            edges.per.bins[[bin]] = c(edges.per.bins[[bin]], i)
+        }
+    }
+
+    ## Track whether to remove numeric edge attributes after splitting.
+    ## Numeric attributes cannot be correctly separated when they
+    ## appear in multi-partial edges removing their semantic meaning
+    numeric.attrs = edge.attr.names[EDGE.ATTR.HANDLING[edge.attr.names] == "sum" & edge.attr.names != "weight"]
+    ignore.numeric.attrs = length(numeric.attrs) == 0
+    ignore.weight.attr   = !("weight" %in% edge.attr.names)
```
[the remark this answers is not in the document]
maxloeffler: Oh, yes, this entire section is a left-over of the previous implementation. I'll clean that up.

util-split.R
```diff
@@ -940,7 +939,9 @@ split.network.by.bins = function(network, bins, bins.vector, bins.date = NULL, r
 
                 ## numeric attributes cannot be correctly separated when they
                 ## appear in multi-partial edges removing their semantic meaning
-                ignore.numeric.attrs = TRUE
+                if (!all(which.partials)) {
+                    ignore.numeric.attrs = TRUE
+                }
```
bockthom: Does this really make sure that holds for all bins? Does `which.partials` really have information of all bins although we are iterating over all bins via `mclapply`? Please double check.

And: In the commit message of the commit that introduced this change, the e-mail address of the signed-off tag is missing...

tests/test-networks-multi-relation.R
```diff
@@ -403,31 +394,31 @@ test_that("Construction of the multi-artifact bipartite network with artifact re
...
+                                      "2016-07-12 14:59:25", "2016-07-12 16:02:30",
+                                      "2016-07-12 16:06:01", "2016-07-15 19:55:39",
+                                      "2017-05-23 12:32:39", "2016-07-12 15:59:59",
+                                      "2016-07-15 20:07:47", "2016-07-27 20:12:08",
+                                      "2016-07-28 06:27:52", "2013-05-25 03:25:06",
+                                      "2013-05-25 06:06:53", "2013-05-25 06:22:23",
+                                      "2013-06-01 06:50:26", "2016-07-12 16:01:01",
+                                      "2016-07-12 16:02:02", "2013-04-21 23:52:09",
+                                      "2016-07-12 15:59:25", "2016-07-12 16:03:59")),
+        artifact.type = c(rep("Feature", 6), rep("IssueEvent", 24)),
```
bockthom, line 412: Do you have an idea why the timezone was included before and why you did change that? I am just curious...

tests/test-networks.R
bockthom: Never mind, I think you can ignore the comment regarding `expected.networks`...

tests/test-networks-multi-relation.R
maxloeffler: I don‘t know why it was there. It was just a check for my curiosity to see whether it makes a difference (it doesn’t), and then I left it in the commit because why not. I can revert it though 😅

util-split.R
maxloeffler: This check should work as intended, here is a quick rundown of how it does so:

First, before we iterate over all bins, we pre-distribute all edges to the bins. In this step, we assign an edge to a bin if there is at least one partial in the edge that belongs in that bin, but an edge might still appear in multiple bins when it has partials that belong to different bins.

Then we iterate all bin-predistributions and all its edges. Here is this check that you commented. It just checks whether an edge inside a bin-predistribution belongs solely into this bin or if it has partials that belong into other bins as well. If it has partials that belong in other bins, i.e., some values in the which.partials vector are FALSE as in that they do not belong into the current bin, then we would need to separate numeric values again, which is not possible. Otherwise, we are fine.

maxloeffler:
Regarding the last open points for discussion:

1. Yes, there are already explicit tests for `get.date.from.string` that check different date formats and check if parsing to `POSIXct` works as expected. Therefore, it really does not matter if we keep the `UTC` in the test where I removed it from or if we add it back.

2. I checked that the new option of passing `dates` to `split.get.bins.time.based` as a nested list does not lead to unforeseen consequences in calling contexts.

3. I tried to find better ways of checking whether we have numeric attributes present on edges than to check if there are attributes that are simplified with the `sum` strategy:
* We discussed potentially checking in the network configuration to see the expected types of edge attributes there, but turns out that we there specify all edge attributes as being `character`.
* Next I checked if we can check if the actual edge attribute values are numeric using `is.numeric`. To make it short, yes we can, we would still need the case distinction for `weight` (and other hypothetical future attributes that are either mandatory or have custom simplification strategies that allow for separation) though.
* Lastly, I also checked the default simplification strategies provided by `igraph`. There are 11 options that have been constant for at least 15 years and none of them are separable (other than `concat`). We can assume that no new options get added, yet I don't know how smart it is to distinguish numeric attributes solely by the simplification strategy if we want to be as precise as we can. We always add custom functions. Furthermore, a default simplification strategy of `igraph` is not limited to numerics even when it may seem so, e.g., you can apply `mean` to `characters` (the result will be `NA` or `NULL` in a list).
In conclusion: yes we can do better than just check for `"sum"` but it would still  requires explicit edge case handling in addition to still potentially requiring change when new attributes get introduced. Personally, I don't think it is likely that we will add new attributes that require a change in this implementation, and even so, the change would be minor. Therefore, I would just leave the current version.

bockthom:
Everything looks good to me now. Thanks for all the investigations and fruitful discussions @maxloeffler. I'll wait for the NEWS before I'll finally approve this PR.

@hechtlC: As I am done with my pass, it is your turn now :smile:

hechtlC (approved):
Look good to me. Thanks @maxloeffler 
As soon as you have written the news and we have reviewed that, we can merge this.

NEWS.md
```diff
@@ -22,10 +22,11 @@
...
 
 ### Changed/Improved
 
-- **Breaking Change**: Change the default representation of edge attributes from vectors to lists. This change is necessary for the interplay of `coronet` networks with certain `igraph` functionality since igraph version 2.1.0 (PR #274, 1c35d1fa2548deb297dbfa5e2b07fce31962c5b7, eda30b838369ec46376812298a3ea8159eec5789, 0c6b2eba79b37f8ef2af7ffc41d86f1f307581bf, 44c7b72e3234cb332bb2713fb408c124e67255d9, 7303eabef6a78198575fe5bdfc02813fde3d3974, 0c27012641d24e19e5fa037406b480034c93f1aa)
+- **Breaking Change**: Change the default representation of edge attributes from vectors to lists. This change is necessary for the interplay of `coronet` networks with certain `igraph` functionality since igraph version 2.1.0 (PR #274, PR #278, 1c35d1fa2548deb297dbfa5e2b07fce31962c5b7, eda30b838369ec46376812298a3ea8159eec5789, 0c6b2eba79b37f8ef2af7ffc41d86f1f307581bf, 44c7b72e3234cb332bb2713fb408c124e67255d9, 7303eabef6a78198575fe5bdfc02813fde3d3974, 0c27012641d24e19e5fa037406b480034c93f1aa, 6fae1843740ed8e48c89c2ee4e61f995b5d0b8f5, 416c817998540fc0b82d9959574838b571b4d6fb)
```
bockthom, line 29: Please don't touch the entries for the already released version 5.0. Everything you have added will end up in the next release. Therefore, please add a new section `## unversioned` in which you need to add new items for all of your changes of this PR. Thank you!

maxloeffler: Oversight 😅

NEWS.md
```diff
@@ -2,6 +2,18 @@
 
 # coronet – Changelog
 
+## unversioned
+
+### Added
+
+- Add the possibility to split networks that contain simplified edges (PR #278, 9798d33512dcdf50d3b88a1223fc4913a2a88eeb, 0ed437c14423c1917f1ba470e7e55db4626d380b, 67a6651b94d50cb7c2ab4a74888b0556d607b102, 98ef83158204be2a67b115cb25df5ba375cccf60, 7ec4d83fdeb308a24a350acd808941807b9511f1, 637d62ab70f098f26f241e588a99cdc49d10f56a, 2c70666f128f96a3a573f29a0cbbef14d803d193, 1cbc6fa36859d6db3a7ff4493ef19763e87d2de3, 41788ff029d038969bfc6b5773e919201c5ac595, b042c0dd08e2229514ccd25dee7a119f25b1ab45, 36d23d657f412aa1953c4773076e593273f19d8e, 402c256d9a05e4ffb297d4ea1fc25d0230787bc0, 54af2b19a112070f10d191b98b055482748426a7, 894414a4a970822b9ecd59c0b6c480860707f636, 894414a4a970822b9ecd59c0b6c480860707f636, 0fe32a259ef703c2de79135bfa6932a595fdc1c5)
```
bockthom, line 9: Could you please remove the duplicated commit hash 894414...? It is enough to mention it once :wink: 

Everything else looks good to me.

hechtlC (dismissed):
Looks good to me as well!
@bockthom you can merge this PR once the last issue is resolved.

bockthom (approved):

PR #281, opened 2025-03-04
Commitmessage functionalities

Leo-Send:
### Prerequisites

- [ ] I adhere to the coding conventions (described [here](../CONTRIBUTING.md)) in my code.
- [ ] I have written appropriate commit messages, i.e., I have recorded the goal, the need, the needed changes, and the location of my code modifications for each commit. This includes also, e.g., referencing to relevant issues.
- [ ] I have put signed-off tags in *all* commits.
- [ ] I have checked whether I need to adjust the showcase file `showcase.R` with respect to my changes.
- [ ] The pull request is opened against the branch `dev`.

### Description

Adds commit message functionalities to coronet, as discussed in issue X. Includes Stemming, Tokenizing, Lemmatization, keyword search and a token count metric.


install.R
```diff
@@ -79,11 +79,14 @@ if (length(p) > 0) {
...
+        matrix.1.5.4.url = "https://cran.r-project.org/src/contrib/Archive/Matrix/Matrix_1.5-4.tar.gz"
+        install.packages(matrix.1.5.4.url, repos = NULL, dependencies = NA, verbose = TRUE, quiet = TRUE)
+        ## redo installation of textstem, which fails if matrix is outdated or not present
+        install.packages("textstem", dependencies = NA, verbose = TRUE, quiet = TRUE)
         Matrix.version = installed.packages()[rownames(installed.packages()) == "Matrix", "Version"]
-        if (compareVersion(Matrix.version, "1.3.0") == -1) {
+        if (compareVersion(Matrix.version, "1.5.0") == -1) {
             print("WARNING: Re-installation of package Matrix did not end up in the necessary package version.")
         }
     }
```
bockthom: Weird that this worked for R4.1, but not for R4.0. Unfortunately, I did not find any other reason for why quanteda stills fails to be installed on R4.0 although Matrix is present.

I have a couple of ideas, but I don't know whether any of this helps:

(1) Try to install system dependency "libtbb-dev" as suggested by quanteda. To do so, could you please add "sudo apt-get install --assume-yes libtbb-dev" to file ".github/workflows/pull_request.yml" in which all the system dependencies are installed for our test pipeline? Maybe this helps... (Please make sure to be able to undo all these installation changes via rebase if they don't work. I don't want to have such things ending up in the coronet commit history...)

(2) If this does not help, we need to figure out why the package installation fails. To do so, we need to find a way to figure out why this fails and need to get access to the full installation logs. Actually, I don't have an idea whether we can enable verbose logs for our CI - I don't think so. The best option would be to run the package installation on a local VM or container on which R4.0 is installed. Unfortunately, we don't have such a local container available right now...

Leo-Send: Adding libtbb-dev does not solve the problem. I will try to run the installation on R4.0 locally next.

Leo-Send: I set up a conda virtual environment for further testing.
Installing r-base 4.0.5 and running 'install.packages("quanteda")' (after installing matrix 1.5.0) results in the following error message:

> ** R
> Error in parse(outFile) : 
>   /tmp/RtmpNJiizT/R.INSTALL1d8d2689383b6/quanteda/R/char_select.R:53:61: unexpected '>'
> 52:                       valuetype = valuetype, 
> 53:                       case_insensitive = case_insensitive) |>
>    ^
> ERROR: unable to collate and parse R files for package ‘quanteda’

I have no idea what that could imply. The only other warning was the following:
> configure: WARNING: Intel TBB not installed; install TBB devel package for parallel processing or update PKG_CONFIG_PATH environment variable

I do not think this has anything to do with it since it seems like a configuration option and not something that would result in a failing installation.

bockthom: Congratulations @Leo-Send! You have found a bug/inconsistency that very likely violates the dependencies of package `quanteda` :tada: 

The code of `quanteda` (assuming that the lines you quoted from "char_select.R" belong to this package) uses the so-called "native pipe" `|>`, which has been introduced in R 4.1.0. Therefore, this does not work with R 4.0 and ends up in an error there.

We definitively should report this to `quanteda`. There are two options how this can be fixed:  
(1) Either they want to support older R versions and don't want to restrict the R version - then they need to replace the native pipe in their code by code that uses only language constructs that are present in R 4.0.
(2) Or they don't want to replace the native pipe - then they have to update their R version dependency in their `DESCRIPTION` file to be compatible with CRAN requirements.

Option (1) would be the best option for us and won't require any changes on our side (except for the need to wait until they release their fix).
Option (2) would restrict us to discontinue the support for R 4.0 or to install an older version of `quanteda` in this case.

In the end, it is up to the package maintainers which option they choose, so we need to wait for that - there is nothing that we can do in the meantime, as we don't know how they will decide. In the past, I've already reported two of such requirements violations to two different R packages (one of them also was about the native pipe). The one R package reacted to my report in a way that they fixed their dependency requirement. The other R package reacted to my report by removing the native pipe from their code as they wanted to support R 4.0 for at least one more year.

So, the question is: Would you like to report this to the package yourself, or would you like me to do that? If you would like to do that, here a few tips: 
- Don't link it to our project here - just mention that you have an environment that still relies on R 4.0.5 and that you are unable to change for legacy reasons. (This might still lead package maintainers to recommend to use a more recent R version, but one could still give it a try).
- Don't mention any other packages (such as matrix) or unrelated warnings (such as the TBB warning) because this is not necessary for reproducing this problem.
- Explain that the use of the native pipe conflicts with the required R version in their `DESCRIPTION` file of their most recent release.
- Cc @bockthom and @hechtlC in the issue there such that we receive notifications from the issue.

You can also draft a message first and let Christian or me check it before you post it there if you are unsure, or @hechtlC  or I can report it if you don't want to do that.

Leo-Send: Thanks, I am opening an issue with quanteda. I will, however, need to mention Matrix as the manual installation of matrix 1.5.0 is needed to reproduce the bug (because the installation would otherwise fail earlier).

install.R
```diff
@@ -83,11 +83,25 @@ if (length(p) > 0) {
...
+
+    quanteda.version = installed.packages()[rownames(installed.packages()) == "quanteda", "Version"]
+    if (identical(quanteda.version, character(0)) || compareVersion(quanteda.version, "3.3.1") == -1) {
+        print("WARNING: quanteda version 3.3.1 or higher is necessary for using coronet. Re-install package quanteda...")
+        ## installation of this quanteda version requires manual installation of packages 'RcppParallel' and 'stringi'
+        install.packages("RcppParallel", dependencies = NA, verbose = TRUE, quiet = TRUE)
+        install.packages("stringi", dependencies = NA, verbose = TRUE, quiet = TRUE)
+        quanteda.3.3.1.url = "https://cran.r-project.org/src/contrib/Archive/quanteda/quanteda_3.3.1.tar.gz"
+        install.packages(quanteda.3.3.1.url, repos = NULL, dependencies = NA, verbose = TRUE, quiet = TRUE)
```
Leo-Send, line 99: I added a check that installs an older version of quanteda manually, similar to the workaround for matrix. It is ugly, as this version requires additional dependencies (which apparently need to be installed seperately... I am unsure why). Nevertheless, we could use this workaround to continue supporting R4.0 without relying on quanteda  fixing their bug.

bockthom: Thanks for the issue report and for finding a temporary workaround. 
A few comments on this:

- Regarding the other hard-coded dependencies: Did you figure out whether setting the `dependencies` parameter of the quanteda installation to `TRUE`  (instead of `NA`) would help to avoid hard coding the installation of the other dependent packages?
- We are not on a rush. Let's wait how the quanteda maintainers react to your issue. If they fix it on there side, we don't need this fix on our side. If not, we need to think about whether we prefer such lengthy workarounds with hard-coded packages, or whether we prefer ceasing support for older R versions - but there still is enough time to do discuss this in a couple of weeks.
- Just a note for future debugging (nothing to change): It might help to temporarily set the `quiet` parameter to `FALSE` to get a more detailed error log on the CI pipeline if we end up in a situation where we don't know what's going on...
- Please don't forget to remove the commit that added the system library. (And when I say "remove", I don't mean to undo the commit, but really remove it via interactive rebase).

Leo-Send: > Did you figure out whether setting the dependencies parameter of the quanteda installation to TRUE (instead of NA)

I tried that, but it didn't work. The installation fails because it could not get the dependencies.

>  It might help to temporarily set the quiet parameter to FALSE to get a more detailed error log on the CI pipeline

Since the bug was reproducible locally, I did not think it was necessary. I did debug without the quiet parameter though.

> remove it via interactive rebase

On it :)

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+#' @param commit.hashes the hashes of the commits that should be considered
+#'
+#' @return a list of vectors containing the tokens from the commit messages
+get.tokenized.commit.messages = function(proj.data, commit.hashes = NULL) {
+    ## get commit message data of the given hashes
+    ## if no hashes are given consider all commits
+    commit.message.data = proj.data$get.commit.messages()
+    if (!is.null(commit.hashes)) {
+        commit.message.data = commit.message.data[commit.message.data$hash %in% commit.hashes, ]
+    }
```
bockthom, line 874: Why is this check contained for the tokenization, but not for stemming or lemmatization?
Also, why is this the only function that does not make use of preprocessing?

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+    tokens = list()
+    for (i in seq_len(nrow(commit.message.data))) {
+        current = commit.message.data[i, "title"]
+        if (proj.data$get.project.conf.entry("commit.messages") == "message") {
+            current = paste(current, commit.message.data[i, "message"])
+        }
+        ## add tokens to result
+        tokens[[length(tokens)+1]] = tm::Boost_tokenizer(current)
+    }
```
bockthom, line 883: Can we transform this to an lapply?
(If not, don't forget about the missing spaces around +)

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+    return(lemmatized.messages)
+}
+
+#' Get Commits messages that match given strings.
+#'
+#' @param proj.data the \code{ProjectData} containing the commit message data
+#' @param commit.hashes the hashes of the commits that should be considered
+#' @param strings the strings that are searched for
+#' @param match the method which describes how many of the strings need to be in a message in order for
+#'              that message to be returned (default: any)
```
bockthom, line 916: the method which :arrow_right: the method that

and also: explain the possible values for this param and use our standard notation for default values, please.

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+    commit.message.data = proj.data$get.commit.messages()
+    if (!is.null(commit.hashes)) {
+        commit.message.data = commit.message.data[commit.message.data$hash %in% commit.hashes, ]
+    }
+
+    for (i in seq_len(nrow(commit.message.data))) {
+        current = commit.message.data[i, "title"]
+        if (proj.data$get.project.conf.entry("commit.messages") == "message") {
+            current = paste(current, commit.message.data[i, "message"])
+        }
```
bockthom, line 932: Same as above: This needs proper documentation...

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+        ## check if message contains strings
```
bockthom, line 933: Use single quotation marks around strings, to differentiate between comment and variable name

tests/test-data-misc.R
```diff
@@ -0,0 +1,445 @@
...
+##
+## You should have received a copy of the GNU General Public License along
+## with this program; if not, write to the Free Software Foundation, Inc.,
+## 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+##
+## Copyright 2025 by Leo Sendelbach <s8lesend@stud.uni-saarland.de>
+## All Rights Reserved.
+
+
+context("Tests for the file 'util-core-peripheral.R'")
```
bockthom, line 18: Nope, definitively not :wink:

README.md
```diff
@@ -429,6 +430,12 @@ In this section, we provide descriptions of the different algorithms we provide
     * calculates scores based on the eccentricity of vertices in a network
     * eccentricity measures the length of the shortest path to each vertex's furthest reachable vertex
 
+#### Commit message functionalities
```
bockthom, line 433: Commit-message functionalities (also in the list of contents)

README.md
```diff
@@ -429,6 +430,12 @@ In this section, we provide descriptions of the different algorithms we provide
...
+
+In this section, we give an overview of the functionalities we offer regarding commit messages. These consist of basic NLP tasks, such as stemming (`get.stemmed.commit.messages`), tokenization (`get.tokenized.commit.messages`), and
```
bockthom, line 435: regarding commit messages :arrow_right: regarding the textual analysis of commit messages.

Maybe we should also add the "textual analysis" to the title... Or "content analysis" would also make sense?

README.md
```diff
@@ -429,6 +430,12 @@ In this section, we provide descriptions of the different algorithms we provide
     * calculates scores based on the eccentricity of vertices in a network
     * eccentricity measures the length of the shortest path to each vertex's furthest reachable vertex
 
+#### Commit message functionalities
+
+In this section, we give an overview of the functionalities we offer regarding commit messages. These consist of basic NLP tasks, such as stemming (`get.stemmed.commit.messages`), tokenization (`get.tokenized.commit.messages`), and 
+lemmatization (`get.lemmatized.commit.messages`), as well as preprocessing steps (`get.preprocessed.commit.messages`) such as lowercase transformation and removal of punctuation, stopwords and extra whitespaces. Apart from these,
+there is the option of using a set of strings to search for matching commits (`get.commit.messages.by.strings`) as well as getting token counts for commit messages (`get.commit.message.counts`).
```
bockthom, line 437: Please use line breaks only between sentences. One sentence, one line.
And add the oxford comma after "stopwords" please.

And please replace "using a set of strings to search for matching commits" by "searching for a set of strings in commit messages for matching commits"

bockthom (changes requested):
I just performed a high-level review, pointing out stylistic issues, missing comments, potential performance issues, and raising a few questions. Please find my individual comments attached.

I did not have a look at the tests yet.

bockthom:
And codecov is complaining about that you don't cover the cases in which `commit.hashes` is **not** `NULL`. Please take care of that in your tests.

bockthom:
And we might also think about whether we add an example on how to use the new functionality in the showcase? It is not necessary, might be nice to have?

util-data-misc.R
Leo-Send: The same check exists in preprocessing, which the other two methods make use of. Therefore, the check would be redundant there.

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+
+    ## create a corpus with all selected commit messages
+    messages = c()
+    for (i in seq_len(nrow(commit.message.data))) {
+        current = commit.message.data[i, "title"]
+        if (proj.data$get.project.conf.entry("commit.messages") == "message") {
+            current = paste(current, commit.message.data[i, "message"])
+        }
+        messages = c(messages, current)
+    }
```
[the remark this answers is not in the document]
Leo-Send: I changed this to now make use of do.call.

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+        corpus = tm::tm_map(corpus, tm::stripWhitespace)
+    }
+
+    ## trim leading and trailing spaces
+    corpus = tm::tm_map(corpus, trimws)
+
+    ## create output dataframe
+    for (i in seq_len(nrow(commit.message.data))) {
+        preprocessed.messages[i,] = c(commit.message.data[["hash"]][i], corpus$content[i])
+    }
```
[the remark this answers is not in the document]
Leo-Send: I changed this to not use a loop anymore. It now builds the dataframe at the end instead of building it in the beginning and then adding things to it later. This way, I do not have to iterate over the data again.

util-data-misc.R
Leo-Send: I changed it to one use of do.call folowed by an lapply.

util-data-misc.R
```diff
@@ -768,3 +771,198 @@ get.issue.is.pull.request = function(proj.data) {
...
+}
+
+#' Apply lemmatization to commit messages of given commits. Preprocessing will be executed as part of this.
+#'
+#' @param proj.data the \code{ProjectData} containing the commit message data
+#' @param commit.hashes the hashes of the commits that should be considered
+#' @param preprocessing the preprocessing steps to be executed (all enabled by default)
+#'
+#' @return a dataframe containing the hashes and lemmatized messages
+get.lemmatized.commit.messages = function(proj.data, commit.hashes = NULL, preprocessing = c("lowercase", "punctuation", "stopwords", "whitespaces")) {
```
hechtlC, line 895: This line is too long as well

hechtlC (changes requested):
Thanks for the additions @Leo-Send !
I have just found two small issues with line length. Please fix these in addition to the issues @bockthom found.

bockthom (changes requested):
Please find a few comments attached.

util-data-misc.R
```diff
@@ -806,13 +806,10 @@ get.preprocessed.commit.messages = function(proj.data,
...
-        if (proj.data$get.project.conf.entry("commit.messages") == "message") {
-            return (paste(title, message))
-        } else {
-            return (title)
-        }
-    }, commit.message.data)
+    messages = commit.message.data$title
+    if (proj.data$get.project.conf.entry("commit.messages") == "message") {
+        messages = paste(messages, commit.message.data$message)
+    }
```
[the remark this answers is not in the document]
Leo-Send: The reason is that we keep the original dataframe in `get.commit.messages.by.strings` while building a new dataframe in the other methods. I thought it made sense this way because it is a sort of filtering operation. I can change the other methods to work the same way, but I would prefer not to change `get.commit.messages.by.strings` because it would mean that I have to reattach the hashes again which requires some kind of loop, since I can't build the new dataframe the way I did before after throwing away some rows.

util-data-misc.R
```diff
@@ -940,21 +934,20 @@ get.commit.messages.by.strings = function(proj.data, commit.hashes = NULL, strin
...
-        if (match(check)) {
-            messages[nrow(messages) + 1, ] = c(commit.message.data[["hash"]][i], current)
-        }
+    ## prepare the dataframe for searching commit messages by merging the 'title' and 'message' columns if desired
+    if (proj.data$get.project.conf.entry("commit.messages") == "message") {
+        commit.message.data$message = paste(commit.message.data[["title"]], commit.message.data[["message"]])
+    } else {
+        commit.message.data$message = commit.message.data[["title"]]
     }
+    commit.message.data = commit.message.data[c("hash", "message")]
```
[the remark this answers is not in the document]
Leo-Send: I changed it to `commit.message.data[, c("hash", "message")]` - I did not even know it works with row/column names if you don't specify whether you mean row or column, so I'm lucky this even worked to begin with ;)

tests/test-data-misc.R
```diff
@@ -402,14 +402,15 @@ test_that("Commit message keyword search: any match, multiple strings", {
                                       "Add some more stuff ",
                                       "I added important things the things are\nnothing",
                                       "Wish intensifies"))
+    rownames(result) = NULL
```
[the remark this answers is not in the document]
Leo-Send: I delete the rownames here because the method used filters rows from the dataframe, and only in this test the rownames do not match the expected result later, as we end up with commits that were in rows 1, 2, 3, and 5. The other tests do not run into this issue. For consistency, I can also remove rownames from the other tests for this method, as the method in general results in a dataframe where rownames do not match row indices.

I do not use this in any other tests, although it is being used in tests for other files, such as `test-core-peripheral`, or in general anywhere where a method changes the ordering within a dataframe and the original indexing is not relevant.

showcase.R
```diff
@@ -501,3 +501,13 @@ calculate.cohens.kappa(author.classification.list = author.class.overview,
...
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Commit-message content analysis -----------------------------------------
+
+get.stemmed.commit.messages(y.data)
+
+get.tokenized.commit.messages(y.data)
+
+get.lemmatized.commit.messages(y.data,
+                               commit.hashes = c("3a0ed78458b3976243db6829f63eba3eead26774"),
+                               preprocessing = ('stopwords'))
```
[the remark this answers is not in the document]
Leo-Send: I specified the order in the documentation of all methods that use preprocessing.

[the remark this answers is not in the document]
Leo-Send: The order does have an impact: consider the string `stem.and.preprocess`, which might be a name of a function in R. Currently, we remove punctuation before removing stopwords, which results in `stem.and.preprocess` -> `stemandpreprocess` -> ´stemandpreprocess` - meaning that stopwordremoval does not affect the function name at all. The other ordering would be to do stopword removal first, resulting in `stem.and.preprocess` -> stem..preprocess` -> `stempreprocess`. In my opinion, removing punctuation first is better, however you probably do not want to do either if you are interested in function names in the commit messages.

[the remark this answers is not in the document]
bockthom: Regarding function names: If you need to search for function names, you should not do any preprocessing, because function names always might contain punctuation characters or stop words. One could think about whether a parameter `search.for.variable.and.function.names` makes sense - but I guess, this would just set the preprocessing to an empty vector and therefore unnecessary - please correct me if I am wrong.

Regarding the order of preprocessing steps: Could you please come up with non-coding related examples (i.e., real technical sentences that could occur in commit messages) and explain the impacts of different preprocessing orders using these examples? This might be more intuitive than looking at function names only. And then we can discuss in our next meeting which ordering makes the most sense.

util-data-misc.R
```diff
@@ -32,6 +32,7 @@ requireNamespace("logging") # for logging
 requireNamespace("tm") # for NLP functionalities
 requireNamespace("SnowballC") # for text stemming used by NLP package "tm"
 requireNamespace("textstem") # for lemmatization
+requireNamespace("parallel") # for parallelization of commit-message keyword search
```
bockthom, line 35: Please remove the last part of the comment. In the future, there might also be other use cases of parallelization, so "for parallelization" should be enough (maybe you can also look into other files what we have written there...)

util-data-misc.R
```diff
@@ -806,9 +808,11 @@ get.preprocessed.commit.messages = function(proj.data,
     }
 
     ## create a corpus with all selected commit messages
-    messages = commit.message.data$title
+    messages = c()
```
bockthom, line 811: We don't need the variable declaration at all if we declare it in all if/else cases, if I am not mistaken - please check again.

util-data-misc.R
```diff
@@ -869,32 +875,36 @@ get.stemmed.commit.messages = function(proj.data,
...
+        messages = commit.message.data[["title"]]
     }
     tokens = lapply(messages, tm::Boost_tokenizer)
 
     return(tokens)
 }
 
 #' Apply lemmatization to commit messages of given commits. Preprocessing will be executed as part of this.
+#' Preprocessing steps are always performed in the following order:
+#' \code{lowercase} -> \code{punctuation} -> \code{stopwords} -> \code{whitespaces}
```
bockthom, line 903: As you use the code environment here, I would suggest to add quotation marks around the strings, as they are actually strings :wink:

bockthom (changes requested):

util-data-misc.R
```diff
@@ -21,13 +21,18 @@
...
 
 ## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
 ## Libraries ---------------------------------------------------------------
 
 requireNamespace("sqldf") # for SQL-selections on data.frames
 requireNamespace("logging") # for logging
+requireNamespace("tm") # for NLP functionalities
+requireNamespace("SnowballC") # for text stemming used by NLP package "tm"
+requireNamespace("textstem") # for lemmatization
+requireNamespace("parallel") # for for parallel computation
```
hechtlC, line 35: two "for"

util-data-misc.R
```diff
@@ -768,3 +773,225 @@ get.issue.is.pull.request = function(proj.data) {
...
+
+#' Apply preprocessing steps to commit messages of given commits. Preprocessing steps are always performed in
+#' the following order: \code{"lowercase"} -> \code{"stopwords"} -> \code{"punctuation"} -> \code{"whitespaces"}
+#'
+#' \code{"lowercase"} transforms all upper case characters into their lowercase counterparts
+#' \code{"stopwords"} removes all stopwords using a list of stopwords
+#'                    for the english language provided by the package \code{tm}
+#' \code{"punctuaton"} removes all punctuation, as described in the ASCII \code{[:punct:]} class,
+#'                     using the r-base \code{regex} functionality. This includes standard punctuation
+#'                     characters such as ",", ".", ":", etc. but also dashes, parantheses, mathematical
```
hechtlC, line 788: parentheses not parantheses

hechtlC (approved):
Thanks for the additions @Leo-Send !
I am fine with this now except for two small spelling/wording errors. Please fix these.

Also I will now let copilot review the PR as well so let's see if it finds anything. If not I'm fine with merging this after you fixed the two small issues.

hechtlC:
> ## Pull Request Overview
> 
> This pull request adds several NLP-based functionalities for analyzing commit messages, including stemming, tokenization, lemmatization, keyword search, and token count metrics. In addition, the PR updates dependency checks, documentation, and workflow configurations to support these new features.
> 
>     * Added new functions in util-data-misc.R for commit-message preprocessing, stemming, tokenization, and lemmatization.
> 
>     * Updated installation scripts and README/NEWS documentation to reflect new dependencies and usage.
> 
>     * Modified the GitHub workflow to update the list of supported R versions.
> 
> 
> ### Reviewed Changes
> 
> Copilot reviewed 7 out of 7 changed files in this pull request and generated 1 comment.
> Show a summary per file
> File 	Description
> util-data-misc.R 	Introduces NLP functionalities and updates related documentation.
> showcase.R 	Demonstrates usage examples of commit-message analysis functions.
> install.R 	Adds new package dependencies and upgrades version checks for Matrix.
> README.md 	Updates documentation to include commit-message content analysis.
> NEWS.md 	Documents the changes and removal of support for older R versions.
> .github/workflows/pull_request.yml 	Updates R version matrix to reflect new minimum requirements.
> Comments suppressed due to low confidence (3)
> 
> **util-data-misc.R:786**
> 
>     * Spelling mistake: 'punctuaton' should be corrected to 'punctuation'.
> 
> 
> ```
> ##' \code{"punctuaton"} removes all punctuation, as described in the ASCII \code{[:punct:]} class,
> ```
> 
> **util-data-misc.R:788**
> 
>     * Typographical error: 'parantheses' should be updated to 'parentheses'.
> 
> 
> ```
> ##'                     characters such as ",", ".", ":", etc. but also dashes, parantheses, mathematical
> ```
> 
> **README.md:438**
> 
>     * Possible typo in the function name: 'get.llemmatized.commit.messages' should match the implementation 'get.lemmatized.commit.messages'.
> 
> 
> ```
> and lemmatization (`get.llemmatized.commit.messages`), as well as preprocessing steps (`get.preprocessed.commit.messages`)
> ```

Ok so copilot found two spelling errors that I missed. Please also fix those and when @bockthom is also fine with the PR we can merge it afterwards!

README.md
```diff
@@ -435,7 +435,7 @@ In this section, we provide descriptions of the different algorithms we provide
 
 #### Commit-message content analysis
 
-In this section, we give an overview of the functionalities we offer regarding the textual analysis of commit messages. These consist of basic NLP tasks, such as stemming (`get.stemmed.commit.messages`), tokenization (`get.tokenized.commit.messages`), and lemmatization (`get.lmmatized.commit.messages`), as well as preprocessing steps (`get.preprocessed.commit.messages`) such as lowercase transformation and removal of punctuation, stopwords, and extra whitespaces. Apart from these, there is the option of searching for a set of strings in commit messages for matching commits (`get.commit.messages.by.strings`) as well as getting token counts for commit messages (`get.commit.message.counts`).
```
Leo-Send: This is a PSA that copilot still hallucinates things and we can not take everything it writes at face value:
Copilot recognized this spelling mistake ('get.lmmatized.commit.messages'). However, it claimed that I had spelled it 'get.llemmatized.commit.messages' in it's comment - which is not what was written here. In this case, it does not lead to a problem, since there was a spelling mistake there, but we still have to be critical of copilot reviews.

Issue #119, opened 2018-05-02
Improve construction of multi-relation networks


clhunsen:
### Description

The construction of multi-relation networks is quite complex in the sense of run-time overhead: We construct the separate networks (by constructing edge lists and, then, networks), transform these to edge lists again, merge them, and construct a merged network in the end. Instead, we should construct edge lists, merge those, and construct a network only once – and improve the run-time for complex and large networks!

Thus, we need to merge edge lists instead of networks [here](https://github.com/se-passau/codeface-extraction-r/blob/d791df8e2c41314f86c36b3af566141e7713f46c/util-networks.R#L548-L563) and [here](https://github.com/se-passau/codeface-extraction-r/blob/d791df8e2c41314f86c36b3af566141e7713f46c/util-networks.R#L605-L624).

In the end, we may end up having the private network-construction methods (such as `get.author.network.cochange()`) returning a similar value as the function `construct.edge.list.from.key.value.list()`.

### Versions

This affects the upcoming version `v3.2` as it contains the code of PR #115.

clhunsen:
> As a first step, we may want to check whether there is only one relation present in the network configuration, then only merge if there are more than one. This is basically a patch analogous to the function `simplify.network`:
> 
> https://github.com/se-passau/codeface-extraction-r/blob/ef094eb7e94691e2a29a8e0f598e514947ebd6b8/util-networks.R#L1168

This is fixed in commit https://github.com/se-passau/coronet/commit/0666f1fffb718063024351b9ccf3c0885bec4acf#diff-7a643d3b27ce60a62055308f46770341R1128-R1132.

bockthom:
> Hey, I thought about this proposed change and think that in addition to the described benefits of reduced redundancy in network construction, it may also bring a couple of disadvantageous consequences.
> 
>     1. [Minor] The private network creation method `get.artifact.network.callgraph` does not create an edge list from which it constructs a network later.

First of all, I am not sure which kind of argument you are going to make here? Is this about inconsistency (as this is the only function that would behave differently than the others), or is there anything else that could become a problem?
And yes, you are right, we don't build the call graph in coronet, we just read it in "pajek" format. Don't remember how fancy this format is, but this would basically be similar to an edge list, I guess.
Second, I am not sure whether we allow callgraphs to be part of multi-relation networks. Could you please check that? 
In the end, we shouldn't care about callgraphs that much...
> 
[...]

PR #282, opened 2025-04-07
Internally cache network data instead of networks to reduce redundancy in network construction

maxloeffler:
### Prerequisites

### Changelog

This works towards fixing #119.

Make internal network creation functions return network data instead of networks. This should improve performance as it removes the redundancy in the `get.author.network`, `get.artifact.network`, and `get.commit.network` when merging the returns of the internal network creation functions into a single network because merging networks requires decomposition in network data. This decomposition is not necessary anymore. As a consequence, this also removes the necessity to first compose networks before decomposing them directly after.

#### Added

- Add helper function `convert.edge.list.attributes.to.list` that converts edge attributes to lists similar to `convert.edge.attributes.to.list` but that takes an edge list as input instead of a network
- Add helper function `construct.network.data` that takes vertex-, and edge-data and constructs a network data object while correctly initializing empty input data

#### Changed / Improved

- Internal network creation functions now return network data instead of networks


util-networks.R
```diff
@@ -1657,12 +1660,27 @@ construct.network.from.edge.list = function(vertices, edge.list, network.conf, d
 #'
 #' @param vertex.data the list of vertex data frames, may be \code{NULL}
 #' @param edge.data the list of edge data frames, may be \code{NULL}
+#' @param network.data the vertex and edge data that should be merged. Each element
+#'                     should be a list with two elements: vertices' and 'edges', that
+#'                     contain the corresponding data for one network.
```
bockthom, line 1746: (1) Shouldn't we remove the parameter documentation for `vertex.data` and `edge.data` if the parameters don't exist any more? :thinking: 

(2) Please use `\code{...}` for parameter mentions within roxygen2 documentation.

(3) What do you mean by "Each element"? I am confused - how many networks are we going to merge? The previous sentence does not say anything about :thinking:

util-networks.R
```diff
@@ -1688,6 +1706,11 @@ merge.network.data = function(vertex.data, edge.data) {
         edges = create.empty.edge.list()
     }
 
+    ## catch case where no vertices (and no vertex attributes) are given
+    if (ncol(vertices) == 0) {
+        vertices = NULL # igraph::graph_from_data_frame fan handle this
```
bockthom, line 1792: Which fan is involved here? :upside_down_face:

util-networks.R
```diff
@@ -176,6 +176,47 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
             return(data.sources)
         },
 
+        ## * * helper functions --------------------------------------------
+
+        #' Construct a network data object from (possibly empty) vertex and edge data.
+        construct.network.data = function(vertex.data, edge.data, allowed.edge.attributes) {
```
bockthom, line 183: Although this is just a helper function, I don't want to go without parameter and return  documentation :wink:

util-networks.R
```diff
@@ -124,17 +124,18 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+        author.network.mail.data = NULL,
+        author.network.cochange.data = NULL,
+        author.network.issue.data = NULL,
+        artifact.network.cochange.data = NULL,
+        artifact.network.callgraph.data = NULL,
+        artifact.network.mail.data = NULL,
+        artifact.network.issue.data = NULL,
+        artifact.network.commit.interaction.data = NULL,
+        commit.network.commit.interaction.data = NULL,
+        commit.network.cochange.data = NULL,
```
bockthom, line 138: [This comment is completely unrelated to this pull request.]

@hechtlC Don't we an author network for commit interactions? If so, shouldn't we have one to be able to compare author networks based on commit interactions and author cochange networks?
If you think we should have something like that (independent of whether you already have a concrete idea on how such a network should look like), please open a new issue to document that we need such kind of network.

[End of off-topic.]

bockthom, line 138: @maxloeffler Thank you for making variable names consistent! I guess there were some discussion about 8 or 9 years ago on why these variables make use of the plural in their first components. I really appreciate that you have fixed this now! :partying_face:

util-networks.R
```diff
@@ -413,85 +416,102 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
 
             ## remove the artifact vertices stemming from untracked files if existing
-            if ("name" %in% igraph::vertex_attr_names(artifacts.net) &&
-                length(igraph::V(artifacts.net)[name == UNTRACKED.FILE.EMPTY.ARTIFACT]) > 0) {
-
-                artifacts.net = igraph::delete_vertices(artifacts.net, UNTRACKED.FILE.EMPTY.ARTIFACT)
+            if ("name" %in% names(vertices) && any(vertices[["name"]] == UNTRACKED.FILE.EMPTY.ARTIFACT)) {
+                vertices = vertices[vertices[["name"]] != UNTRACKED.FILE.EMPTY.ARTIFACT, , drop = FALSE]
+                edges = edges[!(edges[["from"]] == UNTRACKED.FILE.EMPTY.ARTIFACT |
+                        edges[["to"]] == UNTRACKED.FILE.EMPTY.ARTIFACT), ]
```
bockthom, line 427: Could you please explain why we need these statements now? They make totally sense to me - I just wonder whether we shouldn't have had this already beforehand at any other place yet?

util-networks.R
```diff
@@ -690,7 +719,7 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
                 logging::logwarn("Inconsistent issue data. Unequally many 'add_link' and 'referenced_by' issue-events.")
             }
 
-            vertices = unique(artifacts.net.data.raw["issue.id"])
+            vertices = unique(unlist(artifacts.net.data.raw["issue.id"]))
```
bockthom, line 722: Why do we need the `unlist` here? Has this been a bug in a previous implementation?

util-networks.R
```diff
@@ -264,29 +264,33 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
-                author.net.data[["vertices"]],
-                author.net.data[["edges"]],
-                network.conf = private$network.conf,
-                directed = private$network.conf$get.value("author.directed"),
-                available.edge.attributes = private$proj.data$get.data.columns.for.data.source("commits")
+            ## construct network data
+            network.data = private$construct.network.data(
+                vertex.data = authors,
+                edge.data = author.net.data[["edges"]],
+                allowed.edge.attributes = private$proj.data$get.data.columns.for.data.source("commits")
```
bockthom, line 272: Really, we do have attributes named `available.edge.attributes` here but `allowed.edge.attributes` there? Does the difference really make sense, or did it just happen that we used different names without any reason? :thinking: 

I guess this is nothing we should change, not to break the interfaces. But this is something I am interested in on how this had emerged...

bockthom (commented):
Good job @maxloeffler!

I really appreciate your changes. Hope that it really boosts the performance of multi-network creation. And even if this is not the case, this PR contains numerous changes I am in favor of.

util-networks.R
maxloeffler: I wasn't sure about is as most of the private functions don't have docs. I'll add one ^^

util-networks.R
maxloeffler: The code I added there is just a translation of what was there before. Checking if `name` is actually an attribute of the igraph vertices corresponds to checking if there is a column named `name` in the vertices data frame and checking if there are igraph vertices with their name being `UNTRACKED.FILE.EMPTY.ARTIFACT` corresponds to checking if there is such an entry in the data frame. Lastly, deleting all such vertices from the igraph network corresponds to deleting these vertices from the `vertices` data frame as well as deleting all rows in the `edges` data frame that either source from or terminate in such a vertex ^^

util-networks.R
maxloeffler: `artifacts.net.data.raw["issue.id"]` returns a column of a data frame but not a list. Previously this was no problem as igraph accepted this as input to creating a new network but now we do not directly build a network but just network data and in there the `vertices` should be of the format `data.frame(name = <vertices>)` which is only possible when `<vertices>` is a proper list. This structure being relied on by the `merge.network.data` function.

maxloeffler: `artifacts.net.data.raw["issue.id"]` returns a column of a data frame but not a list. Previously this was no problem as igraph accepted this as input to creating a new network but now we do not directly build a network but just network data and in there the `vertices` should be of the format `data.frame(name = <vertices>)` which is only possible when `<vertices>` is a proper list. This structure being relied on by the `merge.network.data` function.

util-networks.R
maxloeffler: The reason for why we have so many different names for edge attributes in this place and related places is that we need to distinguish. First there is the set of all edge attributes that exist in total over all different types of edges. Then there are also sets of edge attributes by data source (e.g., `commits`). I found it reasonable to call the set of all egde attributes `all.edge.attributes` in `construct.network.data`. Now we can debate about the name for the edge attribute sets by data source, `available` or `allowed` or `by.data.source` or whatever, I am fine with it all.

util-networks.R
```diff
@@ -181,7 +181,14 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
         ## * * helper functions --------------------------------------------
 
         #' Construct a network data object from (possibly empty) vertex and edge data.
-        construct.network.data = function(vertex.data, edge.data, allowed.edge.attributes) {
+        #'
+        #' @param vertex.data the vertex data frame
+        #' @param edge.data the edge data frame or NULL to represent an edgeless network [default: NULL]
+        #' @param allowed.edge.attributes a list of all attributes that are may be used in \code{edge.data}
```
bockthom, line 187: (1) the first occurrence of `NULL`  here in the comment needs a `\code{...}`.

(2) I wonder why `edge.data` defaults to `NULL` now, while `vertex.data` doesn't. You have `NULL` checks for both, so even networks without vertices are allowed. What would happen if we have empty `vertex.data` and non-empty `edge.data` - will this result in an error, or will the edges just be ignored because there are no vertices?

(3) "that are may be used" - maybe "that may be used" - but I am still unsure what "may" means in this case and how this is decided. Maybe we should focus more on how this is used. Is `allowed.edge.attributes` just a vector of attribute names, or is it a list/data.frame that contains also the values for all the attributes? And what would happen if the default value is used - do we still get the required attributes then or no attributes at all? :thinking: Maybe you can come up with a short and precise way of describing this...

util-networks.R
```diff
@@ -629,7 +636,6 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
             ## construct edgeless network data
             network.data = private$construct.network.data(
                 vertex.data = data.frame(name = private$proj.data$get.artifacts("mails")),
-                edge.data = NULL,
                 allowed.edge.attributes = private$proj.data$get.data.columns.for.data.source("mails")
```
bockthom: Maybe we would keep that explicitly, even if it is not necessary?
Because if someone wants to add edges in a future implementation this would be easier to spot, I guess.
But in the end I am fine with both - with or without explicitly passing `NULL` here.

util-networks.R
maxloeffler: (2) When you build a network you are not allowed to have edges that source from or terminate in vertices that are not specified, otherwise you will get an igraph exception. De facto `is.null(vertex.data)` is never true I just kept the check (btw the method is a translation of what `construct.network.from.edge.list` did before). Lastly, there is no call to `construct.network.data` with omitted vertices for the above named reasons but there is one without edges and may be more in the future as constructing edgeless networks can be reasonable. Therefore I introduced a `NULL` default on edges but not on vertices.

(3) "may" is supposed to be read like "can" or "should" here. I'll clarify the documentation a bit.

maxloeffler: > What if we have a range with no developer activity at all. Then we might end up passing an empty data frame for vertex.data (and of course, also no edge.data). But in this scenario, it can happen that we have no vertex data and need to create an empty network. But I guess we wouldnt pass NULL` to it in this case but rather an empty data structure, right?

That is right. Currently, there is no explicit invocation of this function with `vertex.data` as `NULL` but there are invocations with empty data frames. What follows from this though? Should we remove the default value?

> But I am still not sure if we would networks without edges consider as the default. To this seems to be rather and exceptional case than a default case...

It's not the common case, yes, but there is no common denominator between all use cases which could be made the default. I just used `NULL` for simplicity when calling the function without edges. Should we just drop all the default values?

maxloeffler:
Before I can finish the `NEWS.md`, I noticed that I forgot the email in the signed-off tag of the first few commit messages. I will add it in so don't wonder for all the commits 😄.

NEWS.md
```diff
@@ -11,6 +11,7 @@
 ### Changed/Improved
 
 - For consistency reasons: Ensure that the values of edge attributes are always lists even when they represent singular values (PR #278, 6fae1843740ed8e48c89c2ee4e61f995b5d0b8f5, 416c817998540fc0b82d9959574838b571b4d6fb)
+- Reduce the amount of redundantly built networks by caching network data internally (PR #282, 64ac42aa743e7f3a724a66bcd551e5b477e30293, 1eda73265a3553e7a785a180118b1c872aeec091, beed2cc9f75619065afce1992a62ecd8ae942ce3, e2dc9954a526eb7ccb1de87571400cbbb8abb76e, 6c3feb9071aa8aa4915825c90d6c3a758538fc8e, 231bc479f7bc04ce4048c6633bae1f1cb15307ca, a64835618f7868c93f6a0b42e196d56cedc799a7, 7537d800542d1bbdd6d3ebcc97472d58a73cbf77, 82fc4cf592c86b1a5ae161ab662433050ef5ce66, f8093033a697472679be045cabda6c1f0197b168, 9c739c0e837184f49d2fe4afb89b1eabf145d912, d6cccd7386cf0c44bfafbd684b454ab7bd1f21c2)
```
bockthom, line 14: Could you please also add the issue number of the issue in which the problem was first discussed (just #xxx as first entry of the parenthesis).

And could you please add the information for which networks this will lead to an improvement? Something like "this is supposed to especially improve the performance of buliding multi-networks", but we should be a little bit more precise than in my suggestions...

And didn't you add cache for some of the commit/commit-interaction networks? Maybe we should add this to fixes? Although it is not really a fix because it was not a bug, just a missing feature :thinking: What do you think?

bockthom:
> Project coverage is 0.00%.

@maxloeffler Could you please figure out why the upload of the coverage report has failed? After the first failing upload (when Codecov started to report a project coverage of 0.00%) I've re-run the upload and it failed again. Thereafter, I've re-run the Build (latest) in order to generate a new coverage report, but even uploading the new report to codecov failed again – I have no idea why. Codecov only provides us with the following hint:

> Unusable report due to issues such as source code unavailability, path mismatch, empty report, or incorrect data format. Please visit our [troubleshooting document](https://docs.codecov.com/docs/error-reference#unusable-reports)
for assistance.

As I have no idea what to do here, I'll leave that up to you to investigate this problem... (but it has a low priority; the other tasks you are currently working on are more important).

NEWS.md
maxloeffler: > And didn't you add cache for some of the commit/commit-interaction networks? Maybe we should add this to fixes? Although it is not really a fix because it was not a bug, just a missing feature 🤔 What do you think?

If it is a missing feature, we should put it in `Added`, we can do that, but I'm not sure whether this is really a feature. To me, this should have existed already and was just an oversight so its more of a `Changed/Fixed`, especially, regarding the fact that it is just an internal change that should not "directly" impact any user of coronet.

NEWS.md
```diff
@@ -14,6 +14,7 @@
...
 - Internally cache commit-network data similarly to how we cache network data for author-, and artifact-networks (PR #282, 6c3feb9071aa8aa4915825c90d6c3a758538fc8e)
 - Remove redundant entries from the list of allowed edge attributes and instead add `event.info.1` and `event.info.2` (PR #282, ad3f8b3e82040a613e91a1744436a2fbdd74fe8d)
+- Ensure that configured or implicitly-enforced undirectedness in partial networks is always dominant over configured or implicit positive-directedness. Furthermore, ensure consistency in the directedness used for edge generation and as a network attribute, especially in networks that consist of multiple partial networks (PR #282, 221990654777dc866b408c9cfdf36976a0edc4b5, e8c642d09db7b11fff4f0b258c901165c0063338, f521dff2814261e1e27ef5e1a92a6cb4b9efb244, 540ba93fcaf439708cba69f8bf3a2de9b03140d2)
```
bockthom, line 17: what do you mean by "implicit positive-directedness"?

And I also would like to explicitly mention multi networks. Maybe wen can write "... especially in networks that consist of multiple partial networks such as multi-networks" ?

bockthom (changes requested):
Thanks for the fixes regarding directedness. I have few comments on that - merely regarding tests and documentation.

tests/test-networks.R
```diff
@@ -1167,3 +1167,93 @@ test_that("Get the data sources from a network with multiple relations on a sing
...
+        update[paste0(network.type, ".directed")] = configured
+        update[[paste0(network.type, ".relation")]] = relations
+        net.conf$update.values(update)
+
+        ## build mutli-network
+        network.builder = NetworkBuilder$new(project.data = ProjectData$new(project.conf = proj.conf), network.conf = net.conf)
+        switch(network.type,
+            "author"   = { network = network.builder$get.author.network() },
+            "artifact" = { network = network.builder$get.artifact.network() }
+        )
```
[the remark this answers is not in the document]
maxloeffler: Copy-paste mistake!

maxloeffler:
Sorry for the many pushes, im just discovering typos ......

maxloeffler:
Sorry for the many pushes, im just discovering typos ......

bockthom:
> Sorry for the many pushes, im just discovering typos ......

"such as example multi networks" is also a typo?

bockthom:
The coverage upload seems to work now again, thanks for the fix @maxloeffler!

Codecov reports 11 lines in `util-networks.R` that are not covered. Could you please have a look at them?
The first two lines reported can be ignored, codecov complains about two lines of comments not being covered :see_no_evil: The remaining 9 affect either callgraph or commit-interaction networks. Callgraph can also be ignored.
For the commit-interaction networks, @hechtlC and I are not sure whether these lines should be covered by tests or not.
What is your opinion?

NEWS.md
```diff
@@ -17,6 +17,7 @@
 - Internally cache commit-network data similarly to how we cache network data for author-, and artifact-networks (PR #282, 3608214b9bcf1ac5edc0c47182993c4fcc95d8b0)
 - Remove redundant entries from the list of allowed edge attributes and instead add `event.info.1` and `event.info.2` (PR #282, 1b156c17f261d8b70d8d48c6cb94d3ee591559f3)
 - Ensure that configured or implicitly-enforced undirectedness in partial networks is always dominant over configured or implicitly-enforced directedness. Furthermore, ensure consistency in the directedness used for edge generation and as a network attribute, especially in networks that consist of multiple partial networks such as multi-networks (PR #282, 65ead39b7b971e5a0acbaee4e787efcf194aafc4, a776caf72256200e1bfa5578106a9b53547b00e7, 257a1c8a6a9b1c3e2a72960cc4051a87950753ee, 41cff01cf141a377c96048f0645e05fb200138e9)
+- Allow the codeface data attributes `event.info.1` and `event.info.2` on network edges (PR #282, 1b156c17f261d8b70d8d48c6cb94d3ee591559f3)
```
bockthom, line 20: Should be issue data instead of codeface data.

bockthom (approved):

PR #285, opened 2025-06-02
Miscellaneous fixes and improvements in network construction

maxloeffler:
### Prerequisites

### Description

Fix miscellaneous minor problems all around coronet:

- Fix a bug in `construct.edge.list.from.key.value.list` that can lead to a crash when building networks. The bug was triggered when two different vertices have the same assigned date.
- Fix a bug in `construct.edge.list.from.key.value.list` and `get.artifact.network.issue` in which the `author.name` attribute could be added again to edges that already had the `author.name` attribute leading to the creation of a redundant `author.name.1` attribute.
-  Add a parameter to `get.networks` to specify which networks to construct to improve performance in use-cases where not all networks are necessary (i.e., when building a multi-network).
- Adjust the flattening of list attributes values in `add.vertex.attribute`. Previously, a problem could lead to `POSIXct` attributes being converted in `numeric`. In addition, add documentation for an undocumented parameter to `add.vertex.attribute` and `split.and.add.vertex.attribute`.

### Changelog

see above (for now)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,12 +14,19 @@
 
 - For consistency reasons: Ensure that the values of edge attributes are always lists even when they represent singular values (PR #278, 6fae1843740ed8e48c89c2ee4e61f995b5d0b8f5, 416c817998540fc0b82d9959574838b571b4d6fb)
 - Reduce the amount of redundantly built networks by caching network data internally. This should improve the performance of building multi-networks, especially, when parts of the multi-networks have been built before (#119, PR #282, 06a814c945f0b20af842d20247126083523cde55, 4793eab02e8792b0640fad88a90018292b1b2ab9, 8ba907fff0534c6fef39bd289ab163c90b053530, 28d22902e32e93c0d4990576da2ef3de88fdffbd, 3608214b9bcf1ac5edc0c47182993c4fcc95d8b0, b30c7f2b5b0a6d12e8024fafada5490170530ebe, 1fa340d6347090a327b4c32ece705c1f700234e5, 40cd55423be7b6521e2fc35f5aa200ff0594e77c, 8fcc74439c28b1592e964dd753bfc1cd57c062be, ca348f1de8e3b4e5786a6d2726ca14e530446896, 1d233af734f79e677d3388f7c3589ce186cc3a8d, 5dd5fc18940ce9ac9598902f175193167b471966)
-- Internally cache commit-network data similarly to how we cache network data for author-, and artifact-networks (PR #282, 3608214b9bcf1ac5edc0c47182993c4fcc95d8b0)
+- Internally cache commit-network data and bipartite-network data similarly to how we cache network data for author-, and artifact-networks (PR #282, PR #285, 3608214b9bcf1ac5edc0c47182993c4fcc95d8b0, aa7e3fae9f45df73054f7d9c6a96177575106c7d)
 - Remove redundant entries from the list of allowed edge attributes and instead add `event.info.1` and `event.info.2` (PR #282, 1b156c17f261d8b70d8d48c6cb94d3ee591559f3)
 - Ensure that configured or implicitly-enforced undirectedness in partial networks is always dominant over configured or implicitly-enforced directedness. Furthermore, ensure consistency in the directedness used for edge generation and as a network attribute, especially in networks that consist of multiple partial networks such as multi-networks (PR #282, 65ead39b7b971e5a0acbaee4e787efcf194aafc4, a776caf72256200e1bfa5578106a9b53547b00e7, 257a1c8a6a9b1c3e2a72960cc4051a87950753ee, 41cff01cf141a377c96048f0645e05fb200138e9)
 - Allow the issue data attributes `event.info.1` and `event.info.2` on network edges (PR #282, 1b156c17f261d8b70d8d48c6cb94d3ee591559f3)
+- Add a `network.type` parameter to `get.networks` in which the caller can specify the types of networks to be constructed. This improves performance in cases where not all network types are needed, such as when building multi-networks (PR #285, bc2efd643d92da547d4677f9026540c25e730a03, e9a0c1681a0b63bd831fe7657ce25a9d50dc6e83)
+- Rename the `list.attributes` parameter in `add.vertex.attribute` and `split.and.add.vertex.attribute` to `flatten.values` with inverted semantics and introduce documentation for it to improve comprehensibility (PR #285, 7dab04a5251d89c9cb286452528ef8b6775a7347)
 
 ### Fixed
+
+- Fix a bug in `construct.edge.list.from.key.value.list` that could cause a crash when constructing a network where different vertices have identical associated timestamps (PR #285, d694a6847c36d32dfee60f7d5b082e1d3ba57e01)
+- Fix a bug in network construction that could lead to edges having an unwanted `author.name.1` attribute (PR #285, 105fec1acc436378a9282c40fcae0eb0257f00be)
+- Ensure that POSIXct values are correctly handled in `add.vertex.attribute`, i.e., that they are not converted to numeric values (PR #285, 7dab04a5251d89c9cb286452528ef8b6775a7347, 4924ac23737dec6f915edaeb350a5cbaebbeec79)
+- Handle empty edges when constructing commit networks using commit-interaction data (PR #285, d5e1e4801230224e6272cd66873bd43bb7f04a00)
 
 ## 5.0
 
--- a/tests/test-networks.R
+++ b/tests/test-networks.R
@@ -1272,3 +1272,67 @@ test_that("Enforcement of directedness in multi-networks", {
     assert.directedness(expected=FALSE,  author.directed=TRUE,  artifact.directed=TRUE, artifact.relations=c("mail", "cochange"))
 
 })
+
+## / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / / /
+## Constructing multiple networks ------------------------------------------
+
+test_that("Construct multiple different networks at once", {
+
+    ## configurations
+    proj.conf = ProjectConf$new(CF.DATA, CF.SELECTION.PROCESS, CASESTUDY, ARTIFACT)
+    proj.data = ProjectData$new(project.conf = proj.conf)
+    net.conf = NetworkConf$new()
+    net.conf$update.value("commit.relation", "commit.interaction")
+
+    ## construct network builder
+    network.builder = NetworkBuilder$new(project.data = proj.data, network.conf = net.conf)
+
+    # available network types
+    available.network.types = c("author", "artifact", "commit", "bipartite")
+    network.type.map = list(
+        author    = "authors.net",
+        artifact  = "artifacts.net",
+        commit    = "commits.net",
+        bipartite = "bipartite.net"
+    )
+
+    assert.correct.networks = function(network.builder, network.type) {
+
+        if (is.null(network.type)) {
+            # get all available networks
+            networks = network.builder$get.networks()
+            network.type = available.network.types
+
+        } else {
+            # get specified networks
+            networks = network.builder$get.networks(network.type = network.type)
+        }
+
+        expected.network.types = unname(sapply(network.type, function(type) network.type.map[[type]]))
+        actual.network.types = names(networks)
+
+        expect_identical(expected.network.types, actual.network.types,
+            info = paste0("all networks (",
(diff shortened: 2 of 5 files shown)


util-networks.R
```diff
@@ -1352,31 +1352,46 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+        #' @param network.type the type(s) of network(s) to be constructed
+        #'                     [default: c("author", "artifact", "commit", "bipartite", "authors.to.artifacts")]
+        #'
+        #' @return networks in a list
+        get.networks = function(network.type = c("author", "artifact", "commit", "bipartite",
+                                                 "authors.to.artifacts")) {
+
+            logging::loginfo("Constructing networks.")
+
+            network.type = match.arg.or.default(network.type, several.ok = TRUE)
```
bockthom: Please check: Isn't there a mismatch between the default value declared in the documentation and the actual `match.arg.or.default` call?
If I understand the documentation of `match.arg.or.default` correctly, the `default` parameter of `match.arg.or.default` defaults to the first element to the allowed choices if no `default` parameter is given - but you would like to have all choices as default and not only the first option...

util-networks.R
```diff
@@ -1352,31 +1352,46 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
+
+            ## bipartite network
+            if ("bipartite" %in% network.type) {
+                networks[["bipartite.net"]] = self$get.bipartite.network()
+            }
+
+            ## author-artifact relation
+            if ("authors.to.artifacts" %in% network.type) {
+                networks[["authors.to.artifacts"]] = private$get.bipartite.relations()
+            }
```
bockthom, line 1392: What the hell is the difference between "authors.to.artifacts" and "bipartite.net"? Could you please figure that out? I know this was already present before this PR (I assume it is present since the creation of coronet), but it looks really strange to me now. So one difference that I can clearly spot here is that "authors.to.artifacts" is a private function, whereas "bipartite.net" is public. But what is different in "authors.to.artifacts" that makes it necessary to have a separate function and to even cache the data separately? There might be some good reason that I might not remember any more, but at the moment it just looks strange to me... Welcome again in the area of software archaeology :wink:

util-networks.R
maxloeffler: You would be correct if the `several.ok` flag would be set to `FALSE` (as it is by default). Then the default return value would be the first value of the input list. When it is set to `TRUE` the default return value is the entire list.

util-networks.R
maxloeffler: Both seem to have their "use-case". `authors.to.artifacts` is sort of the raw edge list for the bipartite edges. This edge list is used in the construction of `bipartite.net` which is just the pure bipartite network.

> But what is different in "authors.to.artifacts" that makes it necessary to have a separate function and to even cache the data separately?

Actually, I think neither one is cached. We could think about caching one.

In [one of my other open PRs](https://github.com/se-sic/coronet/pull/282) we established caching of network data instead of networks. The reason for this was to reduce unnecessary recomputation and redundant network constructions and decompositions. Similarly, we could add caching `get.bipartite.relations`. This would potentially improve performance everywhere where `get.bipartite.network` is accessed (*1*. when constructing the bipartite network *2*. when constructing a multi-network *3*. when constructing an author network and `authors.only.comitters` is set to `TRUE`).

bockthom: Thanks for investigating this. I still wonder while each of these two is callable via `get.networks`. If I use your analogy, only `get.author.network` is called directly in `get.networks`, but not `get.author.network.cochange` - so why should we call `get.bipartite.relations()` in `get.networks`? Could you please check whether this is used anywhere? And if not, could you think about potential use cases? If it is not used and there are no actual use cases, we might think of removing this from `get.networks`...

And regarding the caching:  @maxloeffler Didn't we cache the bipartite network at all? That sounds strange to me.
@hechtlC What is your opinion here? Should we cache the bipartite network data?

bockthom: @maxloeffler Thanks for investigating this. Could you please check whether it breaks any test if you call `get.bipartite.relations` directly instead of using the `networks[["authors.to.artifacts"]]` within `get.multi.network`?

maxloeffler: Of course I did that already before posting my comment ;)

bockthom: Great. Then the only reason I could imagine why it has been previously stored in `get.networks` might have been the caching. That is, even if we would cache the bipartite network, we would always have to compute the bipartite relations ever and ever again from scratch each time we call `get.multi.network`, right?

bockthom: After a short conversation with @hechtlC, we both agree that we would like to cache bipartite networks and remove "authors.to.artifacts" from `get.networks`.

util-networks.R
```diff
@@ -1246,6 +1248,13 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
         #'
         #' @return the bipartite network
         get.bipartite.network = function() {
+
+            ## do not compute anything more than once
+            if (!is.null(private$bipartite.network)) {
+                logging::logdebug("get.bipartite.network: finished. (already existing)")
+                return(private$bipartite.network)
+            }
+
```
bockthom, line 1257: I am confused. Do we cache the data or the network? I guess we don't cache networks any more - but not sure how this can be applied to bipartite networks...

bockthom (approved):
Looks good to me.
@maxloeffler Could you please rebase your branch onto current dev? (Also holds for this branch as well as for #282) After your rebases, I might ask Copilot for a review.

bockthom:
The commit hashes in the most recent update commit 1c17924 of the NEWS.md for this PR (285) still are outdated. But this needs to be updated again anyway after merging #282... So, let's wait until the previous PR is fixed.

bockthom:
> Codecov seems to do nothing at the moment... while the upload job ends succesful, no report is available on codecov itself... Not sure what's going on there...

@maxloeffler Here are the two error messages displayed by Codecov:

---

1. "**There is an error processing the coverage reports with [12045e5](https://app.codecov.io/gh/se-sic/coronet/commit/12045e55e7173b5ec0445c5ba784f684150b12be). As a result, some of the information may not be accurate.**"  

    When clicking on the commit hash, you get the following error message: "**Unusable report due to issues such as source code unavailability, path mismatch, empty report, or incorrect data format. Please visit our [troubleshooting document](https://docs.codecov.com/docs/error-reference#unusable-reports) for assistance.**"

    And this is the report that is uploaded: https://api.codecov.io/upload/gh/se-sic/coronet/download?path=shelter/github/se-sic::::coronet/12045e55e7173b5ec0445c5ba784f684150b12be/8dffdcbc-1cfe-44f1-b4aa-39ee8cf18858.txt

2. "Missing Base Report: Unable to compare commits because the base commit did not upload a coverage report.
[Learn more here](https://docs.codecov.com/docs/error-reference#missing-base-report)"
---

bockthom:
@maxloeffler Could you please rebase this PR onto current `dev`? Then let's see whether codecov works again or not...

bockthom:
@maxloeffler Codecov seems to work now :partying_face: Drawback: It enforces you to write additional tests in order to get the green check mark because the project coverage has decreased by 0.14%...

And it seems that you have not entirely rebased this PR onto `dev` as GitHub still shows that the branch is out-of-date with the base branch. You might have missed to pull the merge commit of the last PR prior to rebasing onto `dev`. But as a master of rebasing you are used to updating commit hashes already, do it one more time :wink:

maxloeffler:
You were right, I missed the merge commit. I correctly rebased on to dev now and added tests for the `network.type` parameter of `get.networks` to cover the uncovered lines. In the tests, I repeatedly build networks, so I configured the `commit.network`s to be commit interaction networks so that we cover the caching related lines that were left uncovered in #282.

tests/test-networks.R
```diff
@@ -1272,3 +1272,72 @@ test_that("Enforcement of directedness in multi-networks", {
...
+        bipartite = "bipartite.net"
+    )
+
+    assert.correct.networks = function(network.builder, network.type = available.network.types) {
+
+        # build networks
+        networks = ifelse(network.type,
+                          network.builder$get.networks(network.type = network.type),
+                          network.builder$get.networks())
+
```
bockthom, line 1305: How does this `ifelse` work? `network.type` seems to be a character string - I have no idea what R does here... Could you please expain / check how/why this works?

tests/test-networks.R
```diff
@@ -1272,3 +1272,72 @@ test_that("Enforcement of directedness in multi-networks", {
...
+                          paste(expected.network.names, collapse = ", "),
+                          ")"))
+    }
+
+    # get all subset combinations of an input list
+    get.subsets = function(input.list) {
+        n = length(input.list)
+        output.list = list(NULL)
+
+        for (i in 1:n) {
```
bockthom, line 1330: Why not using `seq_along` here?

util-networks.R
```diff
@@ -858,11 +858,13 @@ NetworkBuilder = R6::R6Class("NetworkBuilder",
...
-                edges[["artifact.type"]] = ARTIFACT.COMMIT.INTERACTION
+            if (!is.null(edges)) {
+                if (nrow(edges) > 0) {
+                    edges[["artifact.type"]] = ARTIFACT.COMMIT.INTERACTION
+                }
+                colnames(edges)[1] = "to"
+                colnames(edges)[2] = "from"
             }
-            colnames(edges)[1] = "to"
-            colnames(edges)[2] = "from"
```
bockthom, line 862: How did you find this one?

tests/test-networks.R
maxloeffler: Its a simple ternary instruction, if `network.type`: is provided, then call `get.networks` with it as parameter, if not, don't use a parameter. The reason for this is that if it is not given, then it is `NULL` and explicitly providing `NULL` as an argument to `get.networks` does not work.

util-networks.R
maxloeffler: After implementing this test and verifying that it works, I configured commit networks to be commit interaction networks. After I updated the config regarding this, I stumbled upon this that arises because when when `edges` is `NULL`, `nrow(edges)` is `logical(0)` and that is not a valid boolean. The test failed because of it.

tests/test-networks.R
bockthom: The ternary instruction is fine. But the first parameter usually is a boolean. But here it is not. `network.type` is a character string. When does a character string evaluate to `TRUE` in R? That's what I have never seen/used in R before...

tests/test-networks.R
bockthom: Ok, then please use `seq_along` here and get rid of the `n`.
But don't break Codecov's decision making. However, as you were able to fool Codecov last time, I am confident that you will fool it again this time :stuck_out_tongue_winking_eye:

util-networks.R
bockthom: Thanks to Codecov's strict rules, you have found and fix yet another bug :partying_face: 
Without Codecov being so strict in its rules, this PR would already have been merged without this fix...

NEWS.md
```diff
@@ -14,13 +14,20 @@
...
+- Rename the `list.attributes` parameter in `add.vertex.attribute` and `split.and.add.vertex.attribute` to `flatten.values` with inverted semantics and introduce documentation for it to improve comprehensibility (PR #285, 7dab04a5251d89c9cb286452528ef8b6775a7347)
 
 ### Fixed
 
+- Fix a bug in `construct.edge.list.from.key.value.list` that could cause a crash when constructing a network where different vertices have identical associated timestamps (PR #285, d694a6847c36d32dfee60f7d5b082e1d3ba57e01)
+- Fix a bug in network construction that could lead to edges having an unwanted `author.name.1` attribute (PR #285, 105fec1acc436378a9282c40fcae0eb0257f00be)
+- Ensure that POSIXct values are correctly handled in `add.vertex.attribute`, i.e., that they are not converted to numeric values (PR #285, 7dab04a5251d89c9cb286452528ef8b6775a7347, 4924ac23737dec6f915edaeb350a5cbaebbeec79)
+- Handle empty edges when constructing commit networks using commit interaction data (PR #285, 4ce8ddd0174a773ccb36b6ca8135786f655379ed)
```
bockthom, line 29: commit-interaction data needs a hyphen :wink:

tests/test-networks.R
maxloeffler: Ohhh, now I see what you mean. To be honest, I am not sure whether it did what I expected it to do or worked by accident, so I changed this section to make more sense.

bockthom (approved):

PR #286, opened 2025-07-26
Enhance the codeface data for testing

maxloeffler:
### Prerequisites

### Description

Enhance the codeface data used for testing to include two new corner cases. Additionally, maintain consistency in the commit ids between proximity and feature data.

### Changelog

- Ensure that commit ids are unique between proximity and feature codeface data
- Adjust codeface data s.t. one commit touches multiple files / functions
- Adjust codeface data s.t. there are commits by different authors that are issued at the same time
- Retain correct order in data when adding data from additional data sources, e.g., synchronicity data.


bockthom:
> Commit interaction data is indeed never ordered. Now the question is, should it be? Every other major data source seems to be ordered in some way, commits, issues and mails by date, authors by id etc.

I agree that it would make sense to order the commit-interaction data in some way - but no idea which way it would make sense here. @Leo-Send @hechtlC any ideas on that? Or comments why ordering might not be a good idea?

> authors by id etc.

Thank you very much @maxloeffler for mentioning this! @hechtlC This was what I wanted to mention in our last meeting but did not remember any more: While debugging stuff I noticed that authors are not ordered by name but by id, which made debugging more tedious to me. I wonder why don't sort authors by name?  @maxloeffler Could you please figure out how many tests would break if we would order authors by their name?

 > Regarding `update.commit.message.data` I found out that when the ordering of commit messages does not align with the ordering of commits that the merge call destroys the ordering of commits. Commit message data is not ordered upon reading so I think it is best to enforce an ordering after the merge call here.

Good question. If merging commit messages does destroy the ordering of commits, we should definitely fix that. Is there any reason why not ordering commit-message data initially? I guess we could make it similar to pasta data. Do we order pasta data initially? If so, I'd suggest to also order commit messages initially. Otherwise I'd go with your suggestion.

maxloeffler:
> @maxloeffler Could you please figure out how many tests would break if we would order authors by their name?

Its exactly one test in `test-read.R` that fails.

> I guess we could make it similar to pasta data. Do we order pasta data initially?

No, pasta is initially unsorted. Only after pasta data is merged onto the major data sources, sorting of those major sources is reinforced. So lets do the same for commit message data. As we merge commit message data on either `commits` or `commits.unfiltered` and both of those are sorted by date, I will use the fixed version for reordering that I introduced in this PR.

tests/test-networks.R
```diff
@@ -1289,40 +1289,45 @@ test_that("Construct multiple different networks at once", {
...
                           ")"))
     }
 
     # get all subset combinations of an input list
     get.subsets = function(input.list) {
+        n = length(input.list)
         output.list = list(NULL)
-        for (i in seq_along(input.list)) {
+
+        for (i in 1:n) {
```
bockthom, line 1330: I dislike this change. `1:n` might fail if n is 0. `seq_along` is fail-save.

bockthom:
@maxloeffler Thanks for fixing this, looks good now.

Could you please draft the update for the `NEWS.md`? Thank you!

